PR #4, opened 2013-10-05
Refactor drawing functions, so that all functions now work on a graph.

eskimoblood:
- a graph has now `draw` ,`drawMetrcis` and `drawPoints` function
- every function works with canvas and SVG
- decouple the rendering from `graph`
- update API doc and example


Jolg42:
It looks like an old issue … 
I'm closing it for now, tell me if it needs to be reopened 😉

PR #65, opened 2014-10-08
Added support for post version 3.0 which doesn't come with name info.

sunng87:
I‘m not 100% sure if I made it correctly. I was working on some fonts with postscript table version 3.0, which opentype.js doesn't support. So I added another branch in the version switch. And [the spec says](https://www.microsoft.com/typography/OTSPEC/post.htm) it might not come with a names table.

### DIFF
--- a/src/encoding.js
+++ b/src/encoding.js
@@ -188,15 +188,26 @@ function GlyphNames(post) {
             this.names[i] = exports.standardNames[i + post.glyphNameIndex[i]];
         }
         break;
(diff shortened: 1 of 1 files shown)


Jolg42:
It looks like this PR is now outdated.
Version 3 is implemented, see https://github.com/nodebox/opentype.js/blob/e97ad9c37c21b8bd92733d3cfa6346caa6ecbd36/src/encoding.js#L224

PR #81, opened 2014-11-13
add ascender and descender to Font documentation

joshmarinacci:
Add missing docs for ascender and descender properties of the Font object.


fdb:
Thanks @joshmarinacci — sorry it took so long.

PR #86, opened 2015-01-15
fixed: Add support for quad curves

zswang:
@see https://raw.githubusercontent.com/DmitryBaranovskiy/raphael/master/raphael.js
q2c

PR #89, opened 2015-02-03
Add CFF Type 2 CharString operator 'flex1'

### LINKED ISSUE #85: Some font characters lead to wrong display
I have uploaded two fonts ("Neris Light" and "Leaque Gothic Regular") to your glyph inspector. Please see the attached screenshots.

Your can reproduce the error there:
Font: Neris Light, Incorrect character "t" (unicode 0074)
Font: Leaque Gothic Regular, Incorrect character "D" (unicode 0044)

The font files are not broken generally.

You can download the free fonts here:
http://www.fontsquirrel.com/fonts/neris
http://www.fontsquirrel.com/fonts/league-gothic

Please let me know if you need more information.

glyph-inspector-font-leaque-gothic
glyph-inspector-font-neris

tshinnic:
Issue #85 reported malformed characters displayed from two OTF font files.
Investigation found that the CFF CharString instructions in these fonts
employed the rather unusual operator 'flex1', which was not implemented
by opentype.js.  About 8 out of 300+ characters used the operator.

Upon encountering the 'flex1' operator the code would simply skip over
the operator.  This resulted in the accumulated operand values being
left on the stack, which disrupted execution of following operators, and
left the affected characters obviously distorted.  Clearing of the stack
when an unsupported operator is found was added to fix that.

This still left the characters distorted because one of the drawing
instructions was missing.  (In this case the missing part was vertical
so the resulting curves were not tall enough).

Rereading the Adobe documentation induced dread but finally resulted in
working code to implement this operator, though not completely.  When fully
implemented code is supposed to decide whether the Bézier curves will
actually be so shallow that they should just be replaced with a straight
line.  This fix instead always outputs the two Bézier curves.

Tested against the two fonts referenced in issue #85.  (Thank you @georg-paul)
The glyphs 'D' (GID#18) and 'P' (GID#79) were example problems that were
corrected when using font LeagueGothic-Regular.otf    Glyph 't' (GID#294)
was the example problem in font Neris-Light.otf and it was also corrected
with these fixes.

Note that JSHint pointed out the lack of a semicolon from a prior fix #86
and this was added to this fix.

This PR fixes #85

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -425,6 +425,38 @@ function parseCFFCharstring(code, font, index) {
             case 12: // escape
                 v = code[i];
                 i += 1;
+		switch (v) {
+		// 4.1 Path Construction Operators
+		// |- dx1 dy1 dx2 dy2 dx3 dy3 dx4 dy4 dx5 dy5 d6         |-
+		case   37:  // 12 37  0c 25  flex1
+		    var jpx, jpy, c3x, c3y, c4x, c4y;
+
+		    c1x = x   + stack.shift();    // dx1
+		    c1y = y   + stack.shift();    // dy1
+		    c2x = c1x + stack.shift();    // dx2
+		    c2y = c1y + stack.shift();    // dy2
+		    jpx = c2x + stack.shift();    // dx3
+		    jpy = c2y + stack.shift();    // dy3
+		    c3x = jpx + stack.shift();    // dx4
+		    c3y = jpy + stack.shift();    // dy4
+		    c4x = c3x + stack.shift();    // dx5
+		    c4y = c3y + stack.shift();    // dy5
+
+		    if ( Math.abs(c4x - x) > Math.abs(c4y - y) ) {
+		      x   = c4x + stack.shift();
+		    } else {
+		      y   = c4y + stack.shift();
+		    }
+
+		    p.curveTo(c1x, c1y, c2x, c2y, jpx, jpy);
+		    p.curveTo(c3x, c3y, c4x, c4y, x, y);
+                    //p.lineTo(x, y);
+		    break;
+
+		default:
+		    console.log('saw unsupported operator %d', 1200+v);
+		    stack.length = 0;
+		}
                 break;
             case 14: // endchar
                 if (stack.length > 0 && !haveWidth) {
@@ -796,7 +828,7 @@ function glyphToOps(glyph) {
                 y1: _13 * y + _23 * cmd.y1,
                 x2: _13 * cmd.x + _23 * cmd.x1,
                 y2: _13 * cmd.y + _23 * cmd.y1
-            }
+            };
         }
 
         if (cmd.type === 'M') {


miguelsousa:
@tshinnic Would you mind testing with [Source Han Sans](https://github.com/adobe-fonts/source-han-sans/releases) as well? I think the use of flex hints in these fonts is a lot more frequent. Thanks in advance.

tshinnic:
Later on I should be able to more easily summarize random CFF fonts, but for now I've surveyed just the one file, SourceHanSansSC-Regular.otf , from the source-han-sans-1.001R release 2014-09-11 you mentioned above.

```
   142185   on operator  29  callgsubr   
   347559   on operator  10  callsubr    
    92195   on operator  20  cntrmask    
    65535   on operator  14  endchar     
    36525   on operator  27  hhcurveto   
   211114   on operator  19  hintmask    
   294730   on operator   6  hlineto     
    13659   on operator  22  hmoveto     
    14203   on operator   1  hstem       
    51268   on operator  18  hstemhm     
    71406   on operator  31  hvcurveto   
   189026   on operator  24  rcurveline  
   466703   on operator  11  return      
    36659   on operator  25  rlinecurve  
    42197   on operator   5  rlineto     
   356768   on operator  21  rmoveto     
   323752   on operator   8  rrcurveto   
   252614   on operator  28  shortint    
    64196   on operator  30  vhcurveto   
   217725   on operator   7  vlineto     
    30580   on operator   4  vmoveto     
     5685   on operator   3  vstem       
       71   on operator  23  vstemhm     
    54608   on operator  26  vvcurveto   
```
No mysterious/unusual operators there...

However, opentype.js is suffering mightily from this 16MB font of 65K glyphs.  Half the lowercase ASCII letters fail to show up at all.  Looks like a larger misunderstanding... but this is a complex CFF font for sure!  I'll look some more at it.

tshinnic:
Yuck, opentype.js doesn't handle CID fonts and their FDArray and FDSelect additions.  Thus in this font most/every use of subroutines fail.  So for instance, a lowercase 'a' 
```
  Glyph# 66
        on operator      10  callsubr     stack: (1) [-72]    
        on operator      10  callsubr     stack: (1) [-3]    
        on operator      19  hintmask     stack: (0) []    
        on operator      10  callsubr     stack: (2) [45, -90]    
        on operator      19  hintmask     stack: (1) [45]    
        on operator      10  callsubr     stack: (2) [-19, -102]    
        on operator      19  hintmask     stack: (1) [-19]    
        on operator      10  callsubr     stack: (2) [45, -98]    
        on operator      14  endchar      stack: (1) [45]    
```
which has nothing _but_ subroutine calls, draws nothing at all.  

I'll try to poke at this some more tomorrow...

tshinnic:
Oh, but then of course this latter problem has little to do with the situation addressed by this PR, and is a separate matter that needs a separate issue.  (Unless there already is one for CID font support?)

So this PR can proceed in the mean time.

fdb:
Hi @tshinnic thanks for the PR! I'll look at it today and try to merge.

fdb:
Looks good!

Thanks so much for the great write-up, that must be the best documented PR I've ever received!

Best,

F

miguelsousa:
@tshinnic I agree that CID font support is a totally different issue. I've opened #90.

fdb:
@tshinnic can I add your name in the AUTHORS file?

tshinnic:
Yes, thank you.  I should have a couple more PR's later...

I've been being ambitious in a different direction, but having started with opentype.js I have notes about confusions and likely bugs.

For instance, [parseOperand()](https://github.com/nodebox/opentype.js/blob/master/src/tables/cff.js#L90).  That line
```javascript
        return b1 << 8 | b2;
```
won't work for negative numbers.  Something like 
```javascript
        // Force sign extension with extra shifts
        value = ( (b1 << 24) | (b2 << 16) ) >> 16;
```
is needed to force sign extension for negative 16-bit numbers.  But I haven't done the work to *prove* that fails on Dict operands inside opentype.js, except experimenting with the Node.js CLI shows:
```
> v = -123
-123
> (v & 0xFFFF).toString(16)
'ff85'
> b1 = 0xff
255
> b2 = 0x85
133
> (b1 << 8) | b2
65413
> ((b1 << 24) | (b2 << 16)) >> 16
-123
```
The implementation is correct in the other place [parseCFFCharstring()](https://github.com/nodebox/opentype.js/blob/master/src/tables/cff.js#L602).  

All of which of course has me scratching my head as to why this hasn't been noticed before.  Just millions of number operands from CharStrings, but only a few and repetitive operands in Dict's?

fdb:
The bytes are unsigned, so `parseOperand` can never receive negative numbers.

Check [parseByte](https://github.com/nodebox/opentype.js/blob/5897dc76a41d515a086a8f26e3fa4a48fd340954/src/parse.js#L101), which refers to getUint8.

However, there might be some issues lurking there, especially with bit-twiddling. If you have a more robust (but not much slower) approach, please let me know.

PR #91, opened 2015-02-05
Add support for the remaining CFF Type 2 CharString flex operators (flex, hflex and hflex1)

miguelsousa:
I've expanded on @tshinnic's groundwork and added support for the remaining flex operators. Relates to issue #85.

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -422,12 +422,63 @@ function parseCFFCharstring(code, font, index) {
                 break;
             case 11: // return
                 return;
-            case 12: // escape
+            case 12: // flex operators
                 v = code[i];
                 i += 1;
                 switch (v) {
+                case 35: // flex
+                    // |- dx1 dy1 dx2 dy2 dx3 dy3 dx4 dy4 dx5 dy5 dx6 dy6 fd flex (12 35) |-
+                    c1x = x   + stack.shift();    // dx1
+                    c1y = y   + stack.shift();    // dy1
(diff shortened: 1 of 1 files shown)


miguelsousa:
Here are before and after screenshots using the font TrajanPro-Regular.otf
before
after

fdb:
Awesome work, thanks! Can I list you under AUTHORS?

miguelsousa:
Sure. Thank you!

fdb:
No problem, thanks again!

tshinnic:
Thanks for pointing out a font that demonstrated the failures when the operators were unhandled.  Finding a font that exercises a particular feature is rather a pain.

And I just love the very explicit 
```Javascript
            // y = y;       // dy6
```
Comments that mean the code reader doesn't have to think! Golden.  :smile:

miguelsousa:
You're welcome. I work at Adobe in the Type Development team, so I'm quite familiar with the font formats that were invented here. Due to its design features, I knew that Trajan would be one of the font families quite affected by the lack of support for the flex operators.
I think that there are one or two more operators from the 12 series that should probably be supported, since they occur seldomly; the `div` comes to mind, which is used for representing non-integer coordinates. We have a substantial font repository in-house that I can test with. I'll open up an issue or submit a patch if I find anything.

Glad my first contribution went so well. Looking forward to do more of it. `opentype.js` is awesome, and I can easily see it becoming the foundation layer for any browser-based font tool.

Issue #90, opened 2015-02-05
Add support for CID-keyed fonts


miguelsousa:
As @tshinnic discovered, [opentype.js doesn't yet support CID fonts](https://github.com/nodebox/opentype.js/pull/89#issuecomment-72994757). This kind of OpenType fonts is the norm for CJK (Chinese/Japanese/Korean) fonts. [Source Han Sans](https://github.com/adobe-fonts/source-han-sans/releases) is the perfect font family to test with.
/cc @kenlunde

fdb:
It seems performance is going to be an issue as well, simply because of the large character sets of CJK fonts. I'll think about lazy loading the font.

kenlunde:
Keep in mind that Source Han Sans represents the extreme end of the "number of glyphs" spectrum for CJK fonts because the number of glyphs is at the architectural limit of 65,535 glyphs. Typical CJK fonts have less than 30K glyphs, which many at or slightly below 10K glyphs.

tshinnic:
I'm making my own code neater for the CID font case, and then I'll try to jam the equivalent into opentype.js.  The high-level TopDict+PrivateDict in usually coded as there can be only one, but at one point there could be / could have been multiple fonts in a CFF font and thus multiple TopDicts.  That almost exactly duplicates the needed handling of FDArray.  I'm trying to make just one routine to handle the "array of TopDicts" situation and apply to both situations.  I hate duplicating code...

axkibe:
This can be closed IMO as recently fixed.

PR #92, opened 2015-02-08
Add support for CID-keyed font in OTF files

tshinnic:
This particular patch is incomplete and needs more work.  It is submitted
only for evaluation and soliciting comments, working towards resolution
of issue #90 

The _major_ incomplete item is glyph naming for CID-keyed fonts;
I've forgotten where I read about what convention is used!  Suggestions?

opentype.js was not looking for the signal that a CFF table represented
a CID-keyed font, and thus was not capturing the information from the
accompanying FDArray and FDSelect subtables.

In CID-keyed fonts some per-glyph information is found only in the FDArray
subtables, not as usual in the single font-wide Private DICT.  And the
directory of which FDArray[] to use for each glyph is found only in the
one FDSelect lookup list.

The FDArray data layout almost exactly parallels that for the high-level
Top DICT INDEX.  Adding support for FDArray was best done by pulling
Top DICT-related code from parseCFFTable() and creating a new routine to
process an INDEX into a list of 'TopDicts' and their linked Private DICTs.

Various notes:
- Added a missing check (in parseCFFTable()) against a multi-font
  'FontSet', which is not supported by opentype.js;
      (is this feature used by modern font/code?)
- as before, only the first Top DICT from Top DICT INDEX is used;
- I preserved the copying 'defaultWidthX' and 'nominalWidthX' into
  'font' from the TopDict; I think this should be deleted as now
  there could be a different value for each glyph;
- had to move code for calcCFFSubroutineBias() upwards to satisfy JSHint;
- now that multiple tables might be created using the default values
  in TOP_DICT_META and PRIVATE_DICT_META, I wonder about the copying
  of the default values that are arrays.  As I read it, multiple
  tables would _share_ the same array values copied from the defaults.
  Perhaps it would be better in interpretDict() to 'clone' the arrays
  with something like "  [].concat(value)  " ?
- the conditional code at beginning of parseCFFCharstring() ought to
  be shortened to just selecting which TopDict/FDArray[] to then copy
  values from;
- the multiple " width + nominalWidthX " should just be done once at
  end of parsing the CharString;

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -30,6 +30,22 @@ function equals(a, b) {
     }
 }
 
+
+// Subroutines are encoded using the negative half of the number space.
+// See type 2 chapter 4.7 "Subroutine operators".
+function calcCFFSubroutineBias(subrs) {
+    var bias;
+    if (subrs.length < 1240) {
+        bias = 107;
+    } else if (subrs.length < 33900) {
+        bias = 1131;
+    } else {
+        bias = 32768;
+    }
+    return bias;
+}
+
+
 // Parse a `CFF` INDEX array.
 // An index array consists of a list of offsets, then a list of objects at those offsets.
 function parseCFFIndex(data, start, conversionFn) {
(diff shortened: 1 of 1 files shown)


miguelsousa:
> The major incomplete item is glyph naming for CID-keyed fonts;
I've forgotten where I read about what convention is used! Suggestions?

There's no standard way, but one of the conventions is to use the string `cid` followed by the glyph index (e.g. cid534).

> Added a missing check (in parseCFFTable()) against a multi-font
'FontSet', which is not supported by opentype.js;
(is this feature used by modern font/code?)

Here's what @kenlunde had to say about it:

> I am not aware of any implementation that supports more than one FontSet, both in terms of fonts and font consumers. Multiple FontSets represented a way to break the 64K-glyph barrier. In a CFF-only content, meaning without considering the other OpenType tables, this may have worked, but all of the 'sfnt' tables that reference glyphs, such as 'cmap', 'GSUB', and 'GPOS', would additionally require a way to specify or index into a particular FontSet. In other words, the concept of multiple FontSets is effectively dead.

fdb:
Wow, great work! I'll try to look at this ASAP.

fdb:
I would prefer not to have to support multiple FontSets, which only makes everything more difficult for users of opentype.js.

fdb:
This PR has been superseded by #284.

PR #97, opened 2015-02-13
Ignore fs module when used with webpack/browserify

jardakotesovec:


fdb:
Could you elaborate on what problem this solves? It doesn't seem to impact file size, for example.

jardakotesovec:
Without this entry you have to have fs module installed, which is not needed if its used only in browser. So webpack/browserify just [ignores](https://gist.github.com/defunctzombie/4339901) this dependency when packaged for browser, because of this `browser` entry. 

In other words it does not require you to `npm install fs` if you don't use it. Does it answer the question?

Other projects that work in browser and node.js use it for this purpose, here is example: https://github.com/kangax/fabric.js/blob/master/package.json

fdb:
Thanks for the extra explanation! Merged.

PR #98, opened 2015-02-20
Adding support for Path2D based drawing

talltyler:
This branch implements drawing based on Path2D objects.
https://developer.mozilla.org/en-US/docs/Web/API/Path2D.Path2D

This is a new feature in latest Chrome and Firefox, there is a polyfill for browsers that don't support it yet.
https://github.com/google/canvas-5-polyfill

Because drawing should happen all at once drawing should be faster.

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -149,6 +149,24 @@ Font.prototype.getPath = function (text, x, y, fontSize, options) {
     return fullPath;
 };
 
+// Create a Path2D object that represents the given text.
+//
+// text - The text to create.
+// x - Horizontal position of the beginning of the text. (default: 0)
+// y - Vertical position of the *baseline* of the text. (default: 0)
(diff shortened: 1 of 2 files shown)


fdb:
Thanks @talltyler!

However, I don't see a good fall-back in case Path2D is not supported. I would suggest using the regular getPath if Path2D is undefined.

Jolg42:
screenshot 2016-09-10 at 13 02 41
It looks like Path2D is still not supported by every major browser (see https://developer.mozilla.org/en-US/docs/Web/API/Path2D) so as @fdb said we will need a fallback before merging this PR 😉

mudcube:
A shim is now available: https://github.com/google/canvas-5-polyfill

It looks like modern browsers all now supports Path2D :-)

PR #108, opened 2015-04-30
It should be possible to set the ascender and descender explicitely

louisremi:
This is an up-to-date version of #88. I suck at rebasing and fixing conflicts so here it goes. Below is the original description of the PR.

My understanding of ascender and descender values is that they are user defined in font editors, not calculated at export time.
When using opentype.js to create fonts from arbitrary shapes, not being able to define those values results in unpredictable and weird text display.

Thank you in advance : )


fdb:
Thanks @louisremi – only weird part is the files in `dist`, which are unnecessary (I will create those when doing a release). I'll update the commit.

fdb:
This is fixed in the latest release. Thanks!

Issue #122, opened 2015-06-04
font.toBuffer() crashes due to the presence of an undefined element in the strings list


felipesanches:
The bug was fixed in the in-tree copy of opentype.js used in Metapolator:
https://github.com/felipesanches/metapolator/commit/003d95c330ef9828056c7754567adf55d04ed1bf

This is a report in the hope the issue gets fixed upstream as well.

Jolg42:
Quoted from #123 

> In the absence of a test case, closing this PR and revisiting once someone runs into this again will probably be better than keeping this PR open or, far worse, merging it in without a test case to tie it to. Someone will eventually hit this again.

davelab6:
@Jolg42 did you review the fix made downstream?

Jolg42:
@davelab6 This fix https://github.com/felipesanches/opentype.js/commit/e17a2d0f32fd8b50775eba47c0613c4fcbc3f127 ?
In the issue #123 we were thinking that without a test case we will wait before merging it.

PR #123, opened 2015-06-04
Address issue #122 (font.toBuffer() crashes [...]) by checking whethe…

felipesanches:
…r the string value is not undefined before adding it to the strings list.

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -919,7 +919,8 @@ function makeStringIndex(strings) {
     ]);
     t.strings = [];
     for (var i = 0; i < strings.length; i += 1) {
-        t.strings.push({name: 'string_' + i, type: 'STRING', value: strings[i]});
+        if (strings[i])
+            t.strings.push({name: 'string_' + i, type: 'STRING', value: strings[i]});
     }
 
     return t;


felipesanches:
This pull request is related to issue #122

src/tables/cff.js
```diff
@@ -919,7 +919,8 @@ function makeStringIndex(strings) {
     ]);
     t.strings = [];
     for (var i = 0; i < strings.length; i += 1) {
-        t.strings.push({name: 'string_' + i, type: 'STRING', value: strings[i]});
+        if (strings[i])
+            t.strings.push({name: 'string_' + i, type: 'STRING', value: strings[i]});
     }
```
Pomax: This fixes the symptom, but not necessarily the problem: why is there an empty string? Because if there's one thing CFF doesn't allow, it's empty strings in the string index, so it feels more like something went wrong with the input getting sent into `makeStringIndex` (e.g. somehow the charset or dict parses are generating illegal strings that aren't getting caught early enough)

fdb: I agree. The string index is a collection of all strings used in the font. If we used an undefined string somewhere, we should catch it *before* we try to create the index. 

@felipesanches do you have any example code that triggers this issue?

felipesanches:
@fdb, we faced this crash when using opentype.js in Metapolator

https://github.com/metapolator/metapolator

Pomax:
Which font did this happen for? (maybe some debugging can be done against it)

Jolg42:
Hi !

Apparently nothing was merged (https://github.com/nodebox/opentype.js/blob/master/src/tables/cff.js#L921)

It's an old issue now… 
@felipesanches @fdb @Pomax Is it something we can merge?

Pomax:
I would hope not, because this PR fixed the symptom rather than the real problem. A new PR that ensures string parsing throws an error on bad input, instead of yielding empty/null/undefined strings is needed to properly address this issue (if it still is an issue. No font was mentioned for which this goes wrong, so there's no test case that reproduces this problem reliably atm, which is crucial to fixing a problem like this)

fdb:
Paging @felipesanches — do you have a test font that causes undefined / empty strings?

felipesanches:
@fdb It was a long time ago. I can't remember which specific font I was using for testing, sorry :-(

Pomax:
In the absence of a test case, closing this PR and revisiting once someone runs into this again will probably be better than keeping this PR open or, far worse, merging it in without a test case to tie it to. Someone will eventually hit this again.

felipesanches:
I agree. Thanks.

Issue #129, opened 2015-06-18
Passive font loading?


Pomax:
I tried to use OpenType.js to load a set of nine Japanese fonts (think 13500ish glyphs per font) totaling less than 50 MB on disk, and was rather surprised to see memory grow to 1625MB before Node.js crashed with a memory allocation failure error =)

To explain how disastrous the full load really is: normally you want to memory map a big font, and then only do glyph outline fetching when those glyphs are necessary for text shaping. If you don't a 6MB font like Ume Minchou turns into 320MB of fully unpacked object. That means you can't even load three, or even two fonts, on the most cost effective common VPS hosts (digital ocean, vultr, etc)

Is there a way to make OpenType.js passively load a font so that it doesn't build the full font object until it actually needs the parts that are revealed necessary due to function calls? Even loading in everything except the glyph data (only loading that in when it's actually necessary for text shaping calls) should save well over a gig of memory, making loading many fonts a lot more tractable.

Pomax:
I had a look at where this would require changes, by turning the font.glyphs array into a GlyphSet object with deferred parsing, but you're using the total glyph list in *a lot* of places to further load data. Probably still worth doing, since the switch from array to deferred loading won't change how the font behaves and will allow progressive improvement of the code once landed, but it looks like a lot of code was written in the assumption that fonts fit entirely in memory, which hasn't been true for a long time =(

yurenju:
This issue looks related to #126, looks like the same issue.

PR #130, opened 2015-06-18
Fixed #126: Added lazyParsing option to opentype.load()

### LINKED ISSUE #126: UI rendering is blocked if loads a large font
I try to load a Chinese font [DroidSansFallbackFull.ttf](https://github.com/android/platform_frameworks_base/blob/master/data/fonts/DroidSansFallbackFull.ttf) to render CJK font but it will block UI rendering and I can't interact with my website for few seconds.

after profiling startup process, I found the root cause is parsing font to glyphs in synchronized way.

at the first I thought I can use web worker to solve it but I realize that is not a easy way since `Font` is not a serializable object.

my colleague @em suggested me that we can use lazy parsing when `charToGlyph()` is executed and don't have to parse all glyphs in `opentype.load()` -> `opentype.parse()`

There is  a [proof of concept patch](https://github.com/yurenju/opentype.js/commit/de08ec9cd00271f6fea095e513b8adedd78ff3ab) to lazy parsing glyph in `charToGlyph()`.

that should be a possible solution for large font loading issue and wanna to know other suggestion to fix it.

thank you!

yurenju:


Pomax:
there's a few cross-platform improvements in my PR for package.json and the server.js files that are probably worth porting back into this PR, https://github.com/nodebox/opentype.js/pull/131/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R29 + https://github.com/nodebox/opentype.js/pull/131/files#diff-8d44631bfd02d64ee75348ab92f466c0R5

Pomax:
Running the test server with `npm start` does not show the first 100 glyphs, despite the fact that even with lazy parsing, those glyphs should be showing up.

screenshot 39

Pomax:
Using this version of OpenType.js with node, before any glyphs are extracted, still shows the same memory footprint, so somewhere this is still not working correctly. Loading `kiloji_p.ttf`, a freely available Japanese handwriting font, 5.8Mb on disk, the footprint via Node is:

master:
screenshot 41

this PR: 
screenshot 40

So that seems to hint at the fact that all glyphs are still getting loaded somewhere.

yurenju:
@Pomax, Thanks your investigation, I will take a look. and Do you think we are implementing the same feature by different ways? if so it will be better to close my pr and duplicate issue #126 to #129. then we can focus on finding a best way to solve this issue.

Pomax:
I think we are, although the strategies we took are slightly different. You're adding lazy loading at the table level, whereas I added it at the glyph level. It might be worth looking at how much work it would be to put your changes on top of how I made path loading a deferred action, to see if that further improves upon the usability - I didn't touch any of the actual table loading, so it might be that your changes even further improve the benefits we can get from #131

yurenju:
okay let's close this pull request and implement it based on your pull request, your patch is awesome!

PR #131, opened 2015-06-18
Switch from "load all glyphs immediately" to "load  glyphs only once needed"

Pomax:
Fixes https://github.com/nodebox/opentype.js/issues/129

This changes glyph loading to a system where glyph paths are not parsed until they are needed by something. It does this by introducing a `Glyphset` class that acts as the `font.glyphs` container, so rather than `font.glyphs[id]` everything now uses `font.glyphs.get(id)`, and the `Glyph` object now uses a special getter/setter for the `path` property, so that glyphs *don't* parse their associated outline data blocks until `glyph.path` is actually requested. The Glyphset object, when glyphs are pushed into it, instead binds a lazy loader to `glyph.path` until that happens.

The difference in performance is staggering.

This patch also cleans up some of the platform-dependencies that were found in `package.json` and `server.js`, so things should work on all platforms now, and the code passes `npm test` without errors.


bin/server.js
```diff
@@ -5,17 +5,6 @@
 var fs = require('fs');
 var http = require('http');
 var path = require('path');
-var spawn = require('child_process').spawn;
```
Pomax: taken out and made an npm script instead, see https://github.com/nodebox/opentype.js/pull/131/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R32 below

package.json
```diff
@@ -29,17 +29,22 @@
...
+    "dist": "rimraf build && rimraf dist && mkdirp build && mkdirp dist && npm run test && npm run browserify && npm run uglify"
+  },
+  "dependencies": {
   },
-  "dependencies": {},
   "devDependencies": {
     "browserify": "^9.0.3",
     "jscs": "^1.12.0",
     "jshint": "^2.6.3",
+    "mkdirp": "^0.5.1",
```
Pomax: cross-platform "mkdir -p" runner.

package.json
```diff
@@ -29,17 +29,22 @@
...
+    "parallelshell": "^1.1.1",
+    "rimraf": "^2.4.0",
```
Pomax: cross-platform "rm -rf" runner

package.json
```diff
@@ -29,17 +29,22 @@
...
+  },
+  "dependencies": {
   },
-  "dependencies": {},
   "devDependencies": {
     "browserify": "^9.0.3",
     "jscs": "^1.12.0",
     "jshint": "^2.6.3",
+    "mkdirp": "^0.5.1",
+    "parallelshell": "^1.1.1",
```
Pomax: cross-platform parallel task runner, so we can watchify and run server.js at the same time

yurenju:
Hi @Pomax,

looks we are solving the same issue but your pull request have more ambition than my patch on #130.

Pomax:
I'll give your PR a go - if you've already implemented "not parsing glyphs until they're actually necessary", then I have no reason to keep working on the same thing =D

Pomax:
*update* I tried your PR but the memory footprint is still the same, so I may still have a look at trying to turn glyph.path into a lazyloaded property and having glyphs not resolve their compound outline until their `path` is requested.

Pomax:
@yurenju so I spent some more time on this, and got everything working with deferred glyph path loading. The difference speaks for itself. I'm testing this on my font server, which loads nine fonts (TSIB is not loaded):

screenshot 45

Using the current master, we see the following impossible load:

screenshot 46

It doesn't even finish loading six fonts before it hits over a gig and a half of RAM and Node just goes "ehhhhh, no I can't do this".

With this patch applied, we see the following:

screenshot 47

Which is awesome. I'd love to bring that down a bit more, but at the very least Opentype.js can now actually load lots of huge fonts without running into memory issues until it's tried loading about 30 of them, which is already a lot better than 5.

fdb:
Okay, couldn't resist taking a look at this while on holiday. This is amazing. I will try to merge this when I get back on Monday.

Pomax:
I was still a little unhappy with the memory consumption, since it's so close to the 512MB limit of a $5/mo Digital Ocean droplet, so I applied a few more deferral tricks and was able to bring down the memory footprint by another ~5% by making the Glyph constructor surface as small as possible. 

screenshot 23

There is one optimization left that I *know* can be made, but something, somewhere, is frustrating the effort to make it work (not binding `this.font` in the glyph class, since it's only used for its `unitsPerEM` value. Instead it should be entirely possible to, when the Glyphset's deferred loader for `glyph.path` kicks in, also assign `glyph.unitsPerEM` and then use that later on. The problem is that somehow "later" seems to happen "earlier".... future thing to solve =)

Pomax:
And done. I made `unitsPerEm` an aliassed property on `this.path` after it gets built so that there can never be an synchronicity error: as both the path commands and the units per em come from the save object, either things simply work, or they would be way more broken than just missing the quad size.

As such, the memory footprint is now about as low as I think it's going to get without "not even building glyph objects" (i.e. deferring construction of the `glyf` and `cff` tables entirely, and even when we do need them, not parsing them to object, but merely following the pointers to the relevant stretches of data and then selectively parsing those instead).

screenshot 24

Pomax:
Some more thinking: `hmtx` isn't really *used* for anything, so we could entirely late-load that, and save a rather significant amount of data, too (I see the 418MB memory footprint reduced to 392MB if I don't load hmtx at all. I'll try to late-loadify that table, too).

Pomax:
Some more thinking: hmtx isn't really used for anything, so we could entirely late-load that, and save a rather significant amount of data, too (I see the 418MB memory footprint reduced to 392MB if I don't load hmtx at all. I'll try to late-loadify that table, too). I've filed https://github.com/nodebox/opentype.js/issues/135 as a followup issue to look at

davelab6:
Dude this is amazing!! So happy to see this :)

fdb:
Thanks again @Pomax! Awesome work.

yurenju:
@Pomax Awesome! :+1:

Pomax:
cheers - worth pushing a new minor version to npm for? =)

fdb:
Hey @Pomax , cleaning up some stuff, will push a release tomorrow.

fdb:
I just pushed the new 0.4.9 release, which includes your changes. I've also added you to the AUTHORS file. Thanks again!

Pomax:
haha, nice. Looking forward to seeing how much more I can help improve things (after this workweek, during which I expect to have zero free time)

terrancesnyder:
Amazing work @Pomax

PR #134, opened 2015-06-21
Use SPDX license identifier

brawer:
Before this change, npm install would give the following warning:

    $ npm install
    npm WARN package.json opentype.js@0.4.8 No license field.

According to https://docs.npmjs.com/files/package.json#license
"licenses" has been deprecated by "license".

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,12 +14,7 @@
     "ttf",
     "type"
   ],
-  "licenses": [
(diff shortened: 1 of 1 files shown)


fdb:
Thanks for the PR and the link to the docs. Didn't know about SPDX yet.

PR #137, opened 2015-06-23
Unit tests for ‘loca’ table

brawer:


oriSomething:
hey maybe adding `--compilers js:babel/register` to mocha execution and the package babel so unit tests can be written in ES6/7

brawer:
Done; now using Babel.

oriSomething:
sorry, i'm adding more tasks, but it this might needs a think. usage of `.babelrc` with:
```
{
  "optional": [
    "es7.asyncFunctions",
    "es7.classProperties"
  ]
}
```

both aren't stable yet. but async functions can make async tests really easy. the usage of this options, is debatable,  might need discussions. 

@fdb what do you think?

fdb:
Thanks for the tests.

I'm not a big fan of including babel, though. It think it more difficult for new people to contribute. For example, we have to explain that everything in `src` is written in one language (ES5), and everything in `test` is written in another language (ES2015). 

Since it doesn't seem your current test code uses any of the ES2015 features, I suggest removing babel for now.

fdb:
@oriSomething as for async, I've just added `loadSync` so we can avoid async loading in test code.

fdb:
I've now added a basic regression test, also using mocha (but without babel).

If you want I can merge this PR myself, or if you prefer you can update it by pulling from master.

brawer:
Updated; have another look?

brawer:
@fdb, what do you think, OK to merge?

fdb:
Just merged. Thanks for the changes!

PR #138, opened 2015-06-23
Fix incorrect use of meta description

moyogo:
&lt;meta description="..."&gt; should be &lt;meta name="description" content="..."&gt;


fdb:
Thanks!

PR #141, opened 2015-06-27
Support ‘fvar’ tables for variable fonts

brawer:
This table is part of Apple’s Advanced Typography, not OpenType; so
this might be considered out of scope for the opentype.js project.
On the other hand, the FreeType project has implemented AAT variation
axes, so the utility goes beyond MacOS. If opentype.js was fine with
accepting changes for AAT tables, I would implement the ‘avar’ and
‘gvar’ tables in subsequent changes.

### DIFF
--- a/font-inspector.html
+++ b/font-inspector.html
@@ -49,6 +49,8 @@
         <dl id="post-table"></dl>
         <h3 class="collapsed">Character To Glyph Index Mapping Table <a href="https://www.microsoft.com/typography/OTSPEC/cmap.htm" target="_blank">cmap</a></h3>
         <dl id="cmap-table"></dl>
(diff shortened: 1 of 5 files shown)


davelab6:
This is totally awesome and will be a massive massive boot for metapolator :) cc @graphicore @felipesanches

fdb:
@brawer awesome! Do you have some fonts to test this with? Built-in fonts on OS X that use it?

moyogo:
@fdb /Library/Fonts/Skia.ttf has an fvar table.

brawer:
@fdb Just sent you an e-mail with a couple test fonts. Sadly these are unfree, so I can’t post them in public. However, fonttools now has a [sample code snippet](https://github.com/behdad/fonttools/blob/master/Snippets/interpolate.py
) for building your own multiple-master fonts (but it’s work in progress, I wouldn’t fully trust it yet).

fdb:
Thanks @brawer I got them. Note that it will take until Monday-Tuesday before I have time to look at it.

PR #145, opened 2015-06-28
Fix buggy encodings of basic types

brawer:
sizeOf.NUMBER16, sizeOf.NUMBER32: Fix off-by-one errors
encode.UTF16: Handle all Unicode characters, even when not in ISO 8859-1
sizeOf.CHAR: Implement, as it was hidden by a typo
sizeOf.OBJECT, sizeOf.TABLE: Implement missing functions

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -36,7 +36,7 @@ encode.CHAR = function(v) {
     return [v.charCodeAt(0)];
 };
 
-sizeOf.BYTE = constant(1);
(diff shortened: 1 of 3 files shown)


fdb:
OMG some of these are really embarrassing. Thanks for taking the time to clean up the mess and add unit tests for them!

fdb:
As you mention in the tests cases, we will need some deeper tests for CFF encoding corner cases.

PR #147, opened 2015-07-10
Move UTF16 decoding to types

brawer:
The previous implementation was probably incorrect for codepoints
above U+7FFF, since it was parsing signed 16-bit numbers. The new
version uses unsigned 16-bit integers, and has a unit test.


fdb:
Thanks! :+1:

PR #149, opened 2015-07-15
Support multilingual names

brawer:
This change affects the public API to OpenType.js: Instead of
`familyName`, the English font family name is now stored as
`names.fontFamily.en`, the Spanish name as `names.fontFamily.es`,
and so on. Likewise for all other names stored in a font.

OpenType names are rather complex to interpret. The API hides this
complexity; language codes are exposed as IETF BCP 47 language tags,
which most web programmers already know from HTML and XML.

Resolves https://github.com/nodebox/opentype.js/issues/144.

Screenshot:

multilingualnames

### DIFF
--- a/examples/creating-fonts.html
+++ b/examples/creating-fonts.html
@@ -134,7 +134,7 @@
     var buffer = font.toBuffer();
     var font2 = opentype.parse(buffer);
 
-    document.getElementById('fontFamilyName').innerHTML = font2.familyName;
+    document.getElementById('fontFamilyName').innerHTML = font2.names.fontFamily.en;
 
     for (var i = 0; i < font2.glyphs.length; i++) {
         var glyph = font2.glyphs.get(i);
--- a/examples/reading-writing.html
+++ b/examples/reading-writing.html
@@ -67,7 +67,7 @@
         inFont = font;
         var buffer = inFont.toBuffer();
         outFont = opentype.parse(buffer);
-        document.getElementById('fontFamilyName').innerHTML = outFont.familyName;
+        document.getElementById('fontFamilyName').innerHTML = outFont.names.fontFamily.en;
 
         for (var i = 0; i < outFont.glyphs.length; i++) {
             var glyph = outFont.glyphs.get(i);
--- a/font-inspector.html
+++ b/font-inspector.html
@@ -71,6 +71,15 @@ var fontSize = 32;
 var textToRender = 'Grumpy wizards make toxic brew for the evil Queen and Jack.';
 var previewPath = null;
 
+function escapeHtml(unsafe) {
+    return unsafe
+         .replace(/&/g, '&amp;')
+         .replace(/</g, '&lt;')
+         .replace(/>/g, '&gt;')
+         .replace(/\u0022/g, '&quot;')
+         .replace(/\u0027/g, '&#039;');
+}
+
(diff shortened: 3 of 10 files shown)


fdb:
Looks good! However, as I mentioned in email, we could also consider reversing the order to `font.names.en.fontFamily` for example. There's probably a good reason why this approach is better, though.

brawer:
No strong preference. Thinking about it, there might be a reason for slightly preferring `names.fontFamily.{en,fr,ru}`, but it is super marginal.

`names.{en,fr,ru}.fontFamily` would be perfect if the user language was a key for a simple lookup. However, with current language tags, selecting a translation can be more involved: to pick a translation, one needs to know all available languages.

For example, say the user has a preference for language `fr-BE` and translations are available in `en`, `ar`, `de-AT-1901`, and `fr-CA`. Clearly, the best result would be the `fr-CA` translation; how to find it? The usual approach goes like this: Walk over all available translations. For each language, compute the distance to the user language according to some matching function. Finally, pick the translation whose language had the smallest distance.

Real life is more complicated because users can prefer multiple languages in prioritized order, eg. via the HTTP Accept-Language header; see [RFC 4647](https://tools.ietf.org/html/rfc4647). Also, practical matching functions may need some sophistication: Occitan speakers typically understand Catalan but not Swahili, so the distance `oc` → `ca` should be smaller than `oc` → `sw`; Bokmål is very close to Nynorsk; etc. The [ICU LocaleMatcher](http://icu-project.org/apiref/icu4j/com/ibm/icu/util/LocaleMatcher.html) has a few heuristics like this, based on data tables maintained by [Unicode CLDR](http://cldr.unicode.org/).

Now, this language matching business is clearly out of scope for OpenType.js. However, what these algorithms have in common is that they need to iterate over all available translations; a simple look-up is not enough. In that situation, `names.familyName.{en,fr,ru}` might be the marginally better data structure, as one can dereference `names.familyName` and pass this object to a general language matching library which does not know anything about fonts. With `names.{en,fr,ru}.familyName`, the language selection happens on a font-specific data structure.

On the other hand, converting things is extremely trivial, and I’ve no idea if anyone will ever build a localized application on top of opentype.js. So, it totally does not matter in practice.

brawer:
Oh, another thing. In the current version, a font variation axis would eventually look like this:

        axis = {
            tag: 'wdth',
            minValue: 50, defaultValue: 100, maxValue: 200,
            name: {en: 'Width', de: 'Breite'}
        }

If we go with `names.en.fontFamily`, the names of variation axes (“Width”), variation instances (“Ultrabold Condensed”) and features (“Small caps”) would probably have to be identified via numeric IDs, and applications would need to do a look-up themselves?

        axis = {
            tag: 'wdth',
            minValue: 50, defaultValue: 100, maxValue: 200,
            nameID: 259
        }
        names.en[259] = 'Width'
        names.de[259] = 'Breite'

Seems a little clumsy, but certainly not the end of the world. However, assigning a correct ID for a new name is not totally trivial; application programmers would need to know both the Microsoft and the Apple specs for assigning a valid ID. (Each system considers a different numeric range as invalid).

fdb:
Thanks for thinking this through! Let's go with your current approach of `font.names.fontFamily.en`.

brawer:
OK, merging.

PR #154, opened 2015-07-29
Add description for SVG support to README.md

shrhdk:
opentype.js already supports SVG path.
I add description for SVG support to `README.md` because it works well.


fdb:
Thanks!

Issue #135, opened 2015-06-21
Don't load `htmx` table until it's actually needed


Pomax:
The hmtx table loads all glyph `advanceWidth` and `leftSideBearing` values, which is fine for small fonts but for large fonts this puts *a lot* of data  in memory. Instead of loading all data during `hmtx.parseHmtxTable`, called from `opentype.js:parseBuffer`, it would be much nicer to only load in a glyph's values the first time those values are actually needed. The glyph prototype could be assigned default `advanceWidth` and `leftSideBearing` values with getter/setters, such that when they are requested, they *then* consult HMTX and bind the values onto the glyph instance, which then shadows the prototype values:

```
Object.define( Glyph.prototype, 'advanceWidth', {
  configurable: true,
  get: function() { this.setMetricsFromHMTX(); return this.advanceWidth; },
  set: function(v) { this.advanceWidth = v; }
});
```

Before any values are bound, the prototype will be hit, but once the instance has these values bound, it won't hit the prototype anymore. A demonstrator of the concept: http://jsbin.com/zopigokeho/edit?js

fdb:
I like this.

fpirsch:
Could be nice, just be aware that putting a getter on a heavily used property on every glyph could affect performance.

Pomax:
depends on how poorly you implement it. An array backed object that grows itself as it's consulted is perfectly fine. The initial hit will be as slow as the data traversal has to be, after that it's the same speed as accessing any other LUT.

(I already did this for glyphs in https://github.com/nodebox/opentype.js/pull/131, the idea was to simply port the same approach to hmtx. Unfortunately, I've not been able to get to this for a long time)

fpirsch:
I was just talking about simple property access vs a getter function call.

Pomax:
The nice thing about a getter is that things are still JS, so with some clever code you can actually overwrite a getter with the static binding once you have all the data in. But, agreed! Benchmarks are always good.

fpirsch:
What we are saying here is that there are two use cases for a font:
* general font information, which should parse only parts of the file
* glyph rendering, which requires parsing the full font file
This could also be implemented with options passed to the opentype.parse() method, or a new method like e.g. opentype.parseHeaders().

fdb:
[...]

PR #160, opened 2015-09-26
Fixes #140

### LINKED ISSUE #140: glyph-inspector: CFF/OTF glyphs are loaded after they need to be displayed
In the glyph-inspector example page, the pagination doesn’t load the glyphs outlines before they are displayed for CFF/OTF fonts. They are only visible after reloading the pagination page. This happens since #131.

fpirsch:
Fixes #140
The hmtx table was parsed only for TTF fonts. For OTF/CFF fonts, the glyphs advanceWidth was calculated only after lazy-loading of the path, resulting in NaN values.

This may be impacted by #135.


fdb:
Thanks! I understand that this will have an impact if we implement hmtx lazy loading (#135), but this is actually correct, whereas before we had NaN values.

PR #162, opened 2015-10-26
Added: font.getPaths()

dominiklessel:
font.getPaths() creates an array of Path objects that represent the glyps of a given text. This comes in quite handy if you would like to manipulate single chars in a text string later on.


fdb:
Thanks, looks very useful!

dominiklessel:
Thanks for merging this so quickly! Any plans on getting out a new version with this addition?

fdb:
I just pushed 0.5.1. which includes your fix. Thanks!

Issue #171, opened 2016-02-17
Implementation of leftSideBearing / LSB


Jolg42:
Hi,

The leftSideBearing is currently forced to 0 (see -> https://github.com/nodebox/opentype.js/blob/master/src/glyph.js#L167)

We need to use it to handle margins, so I wondered why it was implemented this way, are there any specific reason to set it to 0?

Happy to contribute on that :)

fdb:
We normally do set it later though, when we parse the  [hmtx](https://www.microsoft.com/typography/otspec/hmtx.htm) table: https://github.com/nodebox/opentype.js/blob/master/src/tables/hmtx.js#L24

Do you find that this is not working? Maybe we parse them in the wrong order?

Jolg42:
Yes you're right in fact there is just one thing missing it's the implementation in `glyph.getMetrics()`

Example : If `glyph.leftSideBearing = '11'`, `glyph.getMetrics()` is always returning 0

So I was thinking that we can change this (https://github.com/nodebox/opentype.js/blob/master/src/glyph.js#L162)
```
    var metrics = {
        xMin: Math.min.apply(null, xCoords),
        yMin: Math.min.apply(null, yCoords),
        xMax: Math.max.apply(null, xCoords),
        yMax: Math.max.apply(null, yCoords),
        leftSideBearing: 0
    };
```
to
```
    var metrics = {
        xMin: Math.min.apply(null, xCoords),
        yMin: Math.min.apply(null, yCoords),
        xMax: Math.max.apply(null, xCoords),
        yMax: Math.max.apply(null, yCoords),
        leftSideBearing: this.leftSideBearing
    };
```

Here is my pull request by the way #172

PR #172, opened 2016-02-18
glyph.getMetrics() : return the actual value of glyph's leftSideBearing instead of 0

Jolg42:
Here is my pull request following the issue #171


fdb:
Thanks! Weird that we didn't include LSB in the metrics.

Issue #112, opened 2015-05-12
Add export support for kerning


fdb:
Support both the `kern` table and `GPOS`.

felipesanches:
Issue #152 may be the way to go!

Balearica:
I implemented a bare-bones interface for writing fonts with GPOS pair kerning for my project [Scribe OCR](https://github.com/Balearica/scribeocr).  I plan on turning this into a pull request for opentype.js when I have time, however in the meantime if anybody has a need for writing fonts with kerning, let me know and I can provide code snippets. 

One additional note--while it would be easy from a technical perspective to add support for the `kern` table as well (as specified by the original post), I do not see value in doing so as opentype.js does not support writing TrueType outlines.  According to the official spec, for CFF outlines (what opentype.js creates) kerning information should be specified using the `GPOS` table.  While some programs ignore this detail and use the `kern` table for both CFF and TrueType outlines, the [OpenType Sanitizer](https://github.com/khaledhosny/ots)  (used for both Chrome and Firefox) removes the `kern` table for fonts with CFF outlines.

Balearica:
For anybody else looking for this feature--I now have an open PR that that supports kerning at #743.  Feel free to use the fork linked in that PR.

Issue #95, opened 2015-02-11
Font.setKerningValue(l,r,v) method


mattlag:
Would be nice to have a method for setting kern pair values: Font.setKerningValue(leftglyph, rightglyph, value).  It looks like this is possible by manually editing the Font.kerningPairs object... but i'm not sure if there is some other internal stuff (GposKerningValue?) that also needs to be updated.

davelab6:
The Kern table is deprecated, GPOS kerning should be used instead :)

mattlag:
I'm looking through the opentype.Font object, there doesn't seem to be a GposKerning object (only a getGposKerningValue method).  There *is* a kerningPairs object that appears to hold kern values.  But i'm not sure how internally that maps to exported font features.

I'm going to try directly accessing it.. but like I said, it would be better if there was some sort of official method, as opposed to me just poking around :-)

fdb:
Kerning values can be stored in two ways, either through the (deprecated) `kern` table or the (complicated) `GPOS` table. Currently we only have read support for both.

There is no way to set them so they will be saved, since we simply have no write support for `GPOS`. It should be added, but I haven't found the time yet.

@mattlag On a related note, last year I've been able to implement font writing in opentype.js through the Prototypo Kickstarter project. I'm open to discuss sponsored development of specific features.

mattlag:
Thanks for the kerning info, @fdb.  At the moment, Glyphr Studio is a spare time project, and doesn't have any sponsoring capability :-)

davelab6:
@fdb do you prefer cash to sponsor it, or do you prefer someone else to develop it and make a pull request?

fdb:
@mattlag no problem at all. opentype.js is currently a spare time project as well :-)

@davelab6 both suggestions work for me! Since I have a job in academia, I can do sponsored development during the summer holiday, which is between July and August.

davelab6:
@felipesanches are you interested?

felipesanches:
Yes, I'm interested in working on this. I'll take a look at this today and see if I can do it or if I have any technical question.

davelab6:
@felipesanches to be clear, this means implementing write support for the GPOS table. 

https://github.com/behdad/fonttools/ has support for this in python, so basically the feature request is to port the fontTools code to opentype.js

felipesanches:
great!

fdb:
Awesome :+1:

mattlag:
:smile: :tada:

[...]

PR #175, opened 2016-02-28
Improve TABLE writing to support nested subtables

fpirsch:
Complex tables such as GPOS and GSUB heavily use nested subtables.
Supporting this feature will make it easier to create fonts with kerning and ligatures.


fdb:
This looks awesome. Thanks!

behdad:
Note that GSUB/GPOS don't use deeply-nested subtables per se, whereas they refer to other structs by offset.  It is part of the design of the table that the same object can be encoded once and referred to from multiple locations.

fpirsch:
It appears there is a misunderstanding here.
You're right, "reference by offset" is what I meant by subtable. Like coverageTable, scriptTable, langSysTable and all "tables" defined in the spec.
But `type:TABLE` is sometimes used here merely to group fields together, without deporting the data to the end and referencing by offset. This IMHO is not correct since this is not what the spec calls a table.
Thanks to the unit tests, I spotted and fixed this in `test/name.js` and `tables/fvar.js`.

However @Jolg42 just notified me that this PR introduces a regression as `cff.js` (at least) also uses `type:TABLE` incorrectly.
Seems we need more unit testing :-)

FranzPoize:
I confirm that after the latest PR the font generated cannot be loaded inside a browser in Prototypo.

Here is the output of ot-sanitise :
```
ERROR:  6 F: invalid table tag
Failed to sanitise file!
```

I'll try to look into this or provide a unit test.

FranzPoize:
I'm currently trying to fix and convert the cff table to a table of the right types.

fpirsch:
Oh OK, I was going to do it.

FranzPoize:
@fpirsch there seems to be some regression on the construction of the cff table record also which makes extensive use of the ``TABLE`` type.

The table type doesn't seem to have been design to handle subtabling nicely and I find myself not really up to refactoring all the tables that are used as arrays. I think one simple way is to add a ``SUBTABLE`` type that can be used for offseted subtables and to keep the current ``TABLE`` type as is.

This is what I came up with https://github.com/FranzPoize/opentype.js/commit/011f3039817187e76b88403f547b5d6b6618eb47

fpirsch:
I'll try tro find some time to finish this tonight. Technically it could look like what you've done @FranzPoize, but I'd like to keep TABLE for what the spec describes as tables.

fpirsch:
This should be fixed by #177 (`node generate-font-node` is working again).

@behdad The "once instance - multiple references" feature is not implemented here. Hopefully it will come later.

mattlag:
Hi @fpirsch - Does this, by chance, enable kerning?  Writing GSUB/GPOS tables? There are a couple open issues around this #95 and #112 

Sorry for if this issue isn't quite what I think it is.  But, if it does happen to enable kerning, it would be great to have an abstracted method that does just kerning, or at the very least a table writing example.

Thanks!

fpirsch:
@mattlag This does not enable writing GPOS/GSUB tables yet, but it's a step in this direction. And yes, GPOS/GSUP export is the goal here. I just need some time to work on it.

mattlag:
@fpirsch Okay, thanks so much!  I didn't fully understand this work item, so I just wanted to double check :-)

PR #177, opened 2016-03-03
Fix regression introduced by #175

fpirsch:
This introduces a new class `Record` and the associated type `RECORD`.
It is used to group fields together and provides a `sizeOf` method.

Differences with Table:
* the data is inlined alongside other fields, whereas Table data is stored after all other fields and referenced via an offset.
* this structure is not meant to be recursive

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -2,7 +2,6 @@
 
 'use strict';
 
-var check = require('./check');
 var encode = require('./types').encode;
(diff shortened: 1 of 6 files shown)


FranzPoize:
@fdb Did you have time to take a look a this and could it be merged ? Thanks.

fdb:
Thanks @fpirsch!

PR #180, opened 2016-03-30
Add meta table (read & write)

Jolg42:
Here is the support for the 'meta' table with tests for read & write, based on fonttools implementation:
- [fonttools meta test](https://github.com/behdad/fonttools/blob/d8004de04c193ff293942953787087a5646b96ab/Lib/fontTools/ttLib/tables/_m_e_t_a_test.py)
- [fonttools meta table](https://github.com/behdad/fonttools/blob/a5968458015b519bc415f3ca7d882a428f45c347/Lib/fontTools/ttLib/tables/_m_e_t_a.py)
- [Apple specification for meta table](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6meta.html)


Jolg42:
Just one or two questions for you @behdad since I did the same implementation, about your comments [here](https://github.com/behdad/fonttools/blob/d8004de04c193ff293942953787087a5646b96ab/Lib/fontTools/ttLib/tables/_m_e_t_a_test.py#L11), copied below.
```
# From a real font on MacOS X, but substituted 'bild' tag by 'TEST',
# and shortened the payload. Note that from the 'meta' spec, one would
# expect that header.dataOffset is 0x0000001C (pointing to the beginning
# of the data section) and that dataMap[0].dataOffset should be 0 (relative
# to the beginning of the data section). However, in the fonts that Apple
# ships on  MacOS X 10.10.4, dataMap[0].dataOffset is actually relative
# to the beginning of the 'meta' table, i.e. 0x0000001C again. While the
# following test data is invalid according to the 'meta' specification,
# it is reflecting the 'meta' table structure in all Apple-supplied fonts.
```
- I didn't found an Apple font with the meta table, so do you remember which font is using it?
- About your comments, so it looks like Apple did not followed the spec, but is it safe to implement it like you did? Especially for example if I'm creating a new custom & private 'Foob' tag with my custom data inside?

brawer:
To find fonts with a meta table, try running `grep meta /Library/Fonts/*.ttf`. On MacOS 10.11.4, it returns 33 fonts.

Jolg42:
Thanks @brawer! :+1: 
I added the meta table in the Font-inspector too :)

Jolg42:
@brawer I just figured out that it was your code in fact for the meta table with the comments!
So last question is it safe to implement it like you did? Even for adding new private tags?

brawer:
Thanks for bringing this up!

Regarding the `meta` specification being buggy, I’ve asked the folks at Apple to look into it, and they’ve kindly fixed the wording. Today, the revised spec got pushed to their website. Since the (existing) fonttools implementation matches the (revised) spec, the comments about how and why fonttools is deviating from the published spec are now obsolete. So I’ve removed them in https://github.com/behdad/fonttools/commit/13e862da723521d1d39adb4126a56a943685dd2b.

Regarding private tags in `meta` tables, it would be really surprising if readers didn’t silently skip any tags they don’t recognize. For other aspects of AAT, Apple maintains a [registry](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM09/AppendixF.html), which helps to ensure custom tags don’t conflict. That page has an e-mail address for registering custom font features; perhaps you might want to try contacting them about your custom `meta` subtable so that no other application will use the same tag for something different.

PR #181, opened 2016-03-30
Rename manufacturerURL to vendorURL to follow OpenType specification

Jolg42:
Hi!

I renamed `manufacturerURL` to `vendorURL` to follow the specification:

`11 URL Vendor; URL of font vendor (with protocol, e.g., http://, ftp://). If a unique serial number is embedded in the URL, it can be used to register the font.`

See https://www.microsoft.com/typography/otspec/name.htm

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -32,7 +32,7 @@ function Font(options) {
             designer: {en: options.designer || ' '},
             designerURL: {en: options.designerURL || ' '},
             manufacturer: {en: options.manufacturer || ' '},
-            manufacturerURL: {en: options.manufacturerURL || ' '},
+            vendorURL: {en: options.vendorURL || ' '},
             license: {en: options.license || ' '},
             licenseURL: {en: options.licenseURL || ' '},
             version: {en: options.version || 'Version 0.1'},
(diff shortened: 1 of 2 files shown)


Jolg42:
By the way, I didn't included the dist files because I had this:
screenshot 2016-03-30 at 14 40 32

So it looks like the dist file are not up-to-date?

fdb:
That's weird. Judging from the docs you're correct, but apparently I made a change to "fix validation issues on OS X" in this commit: f1f2ae0dbc1329fc0717bc626613193ca37cb6d9

I'll have to dig a bit to figure out the reason.

Also, don't worry about the dist files. You shouldn't include them in a PR (although it's no trouble if you do). I re-generate them when I create a release.

Jolg42:
I don't think it's changing anything in the font file in fact, because the name is only for us to manipulate the data, in the font it's registered as a position / code. 
See: https://www.microsoft.com/typography/OTSPEC/name.htm
screenshot 2016-03-31 at 15 10 46

fdb:
I realize that — that's why commit f1f2ae0dbc1329fc0717bc626613193ca37cb6d9 is so weird. 

Sidenote:  I currently have a deadline for completing my PhD Thesis (tomorrow), so this'll have to wait a little :-)

Jolg42:
@fdb Renaming is easy, but it will break for people using the old name… So we will need to wait for a major update I think.

PR #182, opened 2016-03-31
Feat/os2values

FranzPoize:
I just build opentype after our last commit. I don't really know where the change in dist/opentype.js come from (at least it's not from us :smile: )

We welcome any comment.

This PR only add os2 values for width weight et flag selection. We mimicked the behavior that happens when someone parse a font (i.e the values goes into tables.os2) and we provided an exhaustive list of values accepted by the standard.


fdb:
Something like this is absolutely necessary. 

It seems like you would need to directly set OS2 table values to have it affect the weight. Could you write example usage code of how that would look?

Jolg42:
Customizing OS/2 values is a good idea ;)
@FranzPoize is it ready for merge?

louisremi:
(Feel free to close this PR now that #199 has been merged)

PR #184, opened 2016-04-30
Wrapped parseBuffer in a try/catch and added test

rBurgett:
The lib currently crashes when it encounters certain fonts rather than handling the error. I caught parseBuffer errors and passed them into the callback. I also added a test to make sure the error is being handled. I noticed there was another PR meant to fix the same issue, but it has not been touched in over a month and did not include a test, so I thought I would make my own. Thanks!

### DIFF
--- a/src/opentype.js
+++ b/src/opentype.js
@@ -289,8 +289,12 @@ function load(url, callback) {
         if (err) {
             return callback(err);
         }
-
-        var font = parseBuffer(arrayBuffer);
+        var font;
+        try {
+            font = parseBuffer(arrayBuffer);
+        } catch (e) {
+            return callback(e, null);
+        }
         return callback(null, font);
     });
 }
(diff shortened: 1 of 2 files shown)


fdb:
This patch is really useful — however could you remove the 13MB file download? It seems excessive that our users would have to download a huge file just to have a test pass.

moyogo:
Besides, 华文黑体.ttf cannot be redristributed without a proper license.

rBurgett:
All right. I removed the file.

fdb:
I've manually squash-merged your commits (so we don't include the font in the commit history): d02e3b67145a16005bb7cc6d377dc86b02a01fd8

Thanks!

fdb:
I've also added the simplest font that causes an error: just the four bytes `OTTO` :-)

PR #192, opened 2016-06-11
Show cmap.glyphIndexMap object & text wrap output

Mottie:
The [font inspector demo](http://opentype.js.org/font-inspector.html) `cmap` currently displays `[object Object]` as the value of `glyphIndexMap`.

This PR replaces that with the actual object & adds `word-break: break-all;` to the css to make all the text wrap. Also, in the `fvar` is usually undefined, so an "Undefined" message is now shown.

Also, in case you're interested, I created [a userscript](https://github.com/Mottie/GitHub-userscripts/wiki/GitHub-font-preview) that uses opentype.js to add a font preview on git font file pages. I blatantly copied the font & glyph inspector code - I hope you guys don't mind.

### DIFF
--- a/font-inspector.html
+++ b/font-inspector.html
@@ -161,6 +161,8 @@ function displayFontData() {
                 html += value.map(function(item) {
                     return JSON.stringify(item);
                 }).join('<br>');
+            } else if (typeof value === 'object') {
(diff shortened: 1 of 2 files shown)


fdb:
Thanks! Looks good.

PR #199, opened 2016-06-30
Add os2 table

MartinBolot:
replaces #182: add os2 table, rebased from opentype.js/master

PS: I also just added a small fix to the detection of errors in advanceWidth: `NaN` should throw as well. Let me know if you need a different patch for that.


fdb:
Looks fantastic, thanks!

fdb:
Just created release 0.6.4 that includes these changes.

fdb:
@MartinBolot could you show some example usage code for specifying font weights that I could include in the docs?

MartinBolot:
@FranzPoize is the one knowledgeable on that subject ;-)

Jolg42:
Nice 👍 

Do you know if the "macStyle" from head table is something that can be interesting to implement for Mac compatibility?
ie: Is macStyle something useful, or maybe most of font parsers are only looking at OS/2 values?

"macStyle" is set to 0 here:
https://github.com/nodebox/opentype.js/blob/ffc412e964ddec28f2dd91f30a59e4728a77aa63/src/tables/head.js#L57
https://www.microsoft.com/typography/otspec/head.htm

Issue #194, opened 2016-06-14
Ligatures


casperno:
Hi,

I've spent some time with opentype, using it to load fonts used in txt.js. It was surprisingly easy, but I miss ligatures.

From what I can see in the code, opentypejs uses GPOS tables for kerning values. Isn't the GPOS also the table for ligatures? Anyone working on this now? If not, any pointer for me if I find the time to try and implement it?

Thanks to all developers for an amazing feat - if it wasn't for the obvious proof that it's working, I wouldn't have believed that parsing of binary font data was feasible in javascript

fpirsch:
Hi casperno,
I wrote the GPOS code, and I am currently working on the GSUB which stores ligatures and other interesting features. I will submit some pull requests before the end of the month to support GSUB reading and (partial) writing.

casperno:
Fantastic! Looking forward to it.

fdb:
Hey @fpirsch that's great news! Let me know if I can help.

fpirsch:
#200 Adds full GSUB parsing. Writing simple cases coming soon.

Harbs:
This is amazing! How would you go about using the tables for specific features (like rlig, liga, onum, salt, etc.)? Is that implemented yet? If not, what are your thoughts on implementation?

fdb:
@fpirsch amazing work!

fpirsch:
@Harbs The GSUB data is exposed through the `font.tables.gsub` object. I'm now working on a `Substitution` object, made available as `font.substitution`, carrying all the methods needed to list, search and edit this data (for the most common features).

Harbs:
I'm not sure what other APIs look like. The FTE model is one that makes sense to me.
http://help.adobe.com/en_US/ActionScript/3.0_ProgrammingAS3/WS56773E36-AC97-4047-B9C1-D9DFA9A51F3F.html
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/text/engine/TextElement.html
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/text/engine/ElementFormat.html
Some similar API seems ideal. I like that opentype.js is finer grained than FTE.

Harbs:
@fdb To be clear, I was not suggesting a full FTE-type implementation. I'd love for someone to do my work for me, but I expect to have to do that myself... ;-)

[...]

PR #201, opened 2016-07-07
GSUB table edition

fpirsch:
As discussed in #194, this is the beginning of an API to edit the data in in the `font.tables.gsub` object.
`font.substitution.getFeature()` returns a list of entries in lookup 4 (ligature) tables.
`font.substitution.add()` can add entries in lookup 4 tables.
More lookup types can be added.

### DIFF
--- a/test/substitution.js
+++ b/test/substitution.js
@@ -1,0 +1,92 @@
+/* jshint mocha: true */
+
+'use strict';
+
+var assert = require('assert');
+var opentype = require('../src/opentype.js');
+var Substitution = require('../src/substitution.js');
+
+describe('substitution.js', function() {
+
+    var font;
+    var substitution;
+    var notdefGlyph = new opentype.Glyph({
+        name: '.notdef',
+        unicode: 0,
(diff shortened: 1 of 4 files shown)


test/substitution.js
```diff
@@ -0,0 +1,92 @@
+/* jshint mocha: true */
+
+'use strict';
+
+var assert = require('assert');
+var opentype = require('../src/opentype.js');
+var Substitution = require('../src/substitution.js');
```
fdb: I don't think you need the ".js" here.

fdb:
This is awesome!! I could probably use some help writing the docs for this, though. 

Come to think of it, the README starts to become rather unwieldy. Maybe it's time for a new documentation format (so we can have a chapter on substitution?)

test/substitution.js
fpirsch: Yes, absolutely.

fpirsch:
Yes, substitution on its own could use a whole doc...

PR #204, opened 2016-07-14
(limited) GSUB writing

fpirsch:
Adds simple GSUB writing capabilities.
- Common Layout Tables writing is exposed in `table.js` to be reused with other tables (mostly)
- Only lookup type 1 format 1 (single substitution) and lookup type 4 (ligature)
- Only Coverage format 1 (individual glyphs)
- Only DFLT script and defaultLangSys

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -2,6 +2,7 @@
 
 'use strict';
 
+var check = require('./check');
 var encode = require('./types').encode;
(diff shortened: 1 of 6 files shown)


fdb:
Awesome!

fdb:
@fpirsch I also gave you write access to the OT repo, since you obviously know what you're doing :-)

PR #205, opened 2016-07-14
No glyph names fix

Vildan:
If a font file doesn't have font.glyphNames.names then function
addGlyphNames(font) throws an error "TypeError: Cannot read property '0'
of undefined"


fdb:
Thanks!

PR #207, opened 2016-07-18
Format 12 support

Vildan:
Passed npm test

### DIFF
--- a/src/tables/cmap.js
+++ b/src/tables/cmap.js
@@ -7,37 +7,34 @@ var check = require('../check');
 var parse = require('../parse');
 var table = require('../table');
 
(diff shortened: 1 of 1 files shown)


Vildan:
@fdb , please let me know if I need to change anything else. Tested it on [Noto Sans Egyptian Hieroglyphs](https://www.google.com/get/noto/#sans-egyp), which is in range U+13000 - U+1342E. It works correctly. Also tested on thousands regular BMP Unicode layer fonts. Works correctly too. Please let me know if I need to change something in syntactic or in logic.

Jolg42:
I was curious and tested it, works good for me too :) Thank you @Vildan 👍

fdb:
Looks good, thanks! :+1:

PR #211, opened 2016-07-31
Added JSDocs for a lot of classes

Harbs:
I added JSDocs and type information for a lot of classes. It's not complete, but it's a very good start.

You can view some generated JSDocs [on my site here.](http://in-tools.com/stuff/opentype.js/)

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -205,7 +205,7 @@ Font.prototype.getKerningValue = function(leftGlyph, rightGlyph) {
  * @type Object
  * @property {boolean} [kerning] - whether to include kerning values
  */
- 
+
 /**
  * Helper function that invokes the given callback for each glyph in the given text.
  * The callback gets `(glyph, x, y, fontSize, options)`.* @param  {string} text
@@ -243,7 +243,7 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback) 
  * @param  {number} [x=0] - Horizontal position of the beginning of the text.
  * @param  {number} [y=0] - Vertical position of the *baseline* of the text.
  * @param  {number} [fontSize=72] - Font size in pixels. We scale the glyph units by `1 / unitsPerEm * fontSize`.
- * @param  {GlyphRenderOptions=} options 
+ * @param  {GlyphRenderOptions=} options
  * @return {opentype.Path}
  */
 Font.prototype.getPath = function(text, x, y, fontSize, options) {
--- a/src/font.js
+++ b/src/font.js
@@ -205,7 +205,6 @@ Font.prototype.getKerningValue = function(leftGlyph, rightGlyph) {
  * @type Object
  * @property {boolean} [kerning] - whether to include kerning values
  */
-
  
 /**
  * Helper function that invokes the given callback for each glyph in the given text.
@@ -257,17 +256,8 @@ Font.prototype.getPath = function(text, x, y, fontSize, options) {
     return fullPath;
 };
 
-// Create an array of Path objects that represent the glyps of a given text.
-//
-// text - The text to create.
-// x - Horizontal position of the beginning of the text. (default: 0)
-// y - Vertical position of the *baseline* of the text. (default: 0)
-// fontSize - Font size in pixels. We scale the glyph units by `1 / unitsPerEm * fontSize`. (default: 72)
-// Options is an optional object that contains:
-// - kerning - Whether to take kerning information into account. (default: true)
-//
-// Returns an array of Path objects.
 /**
+ * Create an array of Path objects that represent the glyps of a given text.
  * @param  {string} text - The text to create.
  * @param  {number} [x=0] - Horizontal position of the beginning of the text.
  * @param  {number} [y=0] - Vertical position of the *baseline* of the text.
--- a/src/font.js
+++ b/src/font.js
@@ -9,9 +9,42 @@ var glyphset = require('./glyphset');
 var Substitution = require('./substitution');
 var util = require('./util');
 
-// A Font represents a loaded OpenType font file.
-// It contains a set of glyphs and methods to draw text on a drawing context,
-// or to get a path representing the text.
+/**
+ * @typedef FontOptions
+ * @type Object
+ * @property {Boolean} empty - whether to create a new empty font
+ * @property {string} familyName
+ * @property {string} styleName
+ * @property {string=} fullName
+ * @property {string=} postScriptName
+ * @property {string=} designer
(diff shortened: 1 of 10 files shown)


oriSomething:
wow 👍

Harbs:
On second thought, I'm not sure the JSDocs that I did are totally the right way to go. I needed to do some non-standard tags to make it work and exporting Google externs does not work very well. You will also notice that trying to view the source from the JSDocs does not work, and I believe it's casued by the same problem.

(I'm also not sure I exported all type info correctly.)

If there's any chance of it being merged, I'll work on a bit of a structure refactor to work with JSDocs a bit better.

Thoughts?

src/font.js
```diff
@@ -9,9 +9,42 @@ var glyphset = require('./glyphset');
...
+ * @property {string=} license
+ * @property {string=} licenseURL
+ * @property {string=} version
+ * @property {string=} description
+ * @property {string=} copyright
+ * @property {string=} trademark
+ * @property {Number} unitsPerEm
+ * @property {Number} ascender
+ * @property {Number} descender
+ * @property {Boolean=} createdTimestamp
```
Jolg42: createdTimestamp is a number 😉

src/types.js
```diff
@@ -394,13 +549,18 @@ encode.MACSTRING = function(str, encoding) {
                 return undefined;
             }
         }
-
-        result.push(c);
+        result[i] = c;
```
Jolg42: Do you know @fpirsch if ` result.push(c);` is faster than direct assignment `result[i] = c;`?
It looks like it's faster… and easier to read too :) http://stackoverflow.com/a/15650115/1345244

Harbs: Here's a test that I did in Firefox and Chrome with results:
function timeJoin(){
	var start = new Date();
	var i=-1;
	var len  = 10000000;
	var arr = [];
	while(++i<len)
	{
		arr[i] = "a";
	}
	var str = arr.join("");
	var end = new Date();
	console.log(end.getTime()-start.getTime());
};
function timePush(){
	var start = new Date();
	var i=-1;
	var len  = 10000000;
	var arr = [];
	while(++i<len)
	{
		arr.push("a");
	}
	var str = arr.join("");
	var end = new Date();
	console.log(end.getTime()-start.getTime());
};

Chrome:
timeJoin()
822
timePush()
979

Firefox:
timeJoin()
184
timePush()
303

Unless there's something wrong with my test, push seems to be slower -- at least in Firefox. (Safari seems to be slightly slower with push as well.) Older browsers had a much bigger gap. I'm not sure about IE.

Jolg42:
Hi @Harbs 
It looks good 👍  I think we can merge that soon, more documentation is always nice!

I just think that the changes from '.push()' to direct assignment is less readable & looks not as fast from a performance view, @fpirsch will know that better ;)

I don't know if it's possible to take the file "opentype.js/src/opentype.js" as the entry of JSDoc, because it will be better to have the real source when clicking on the generated doc (not the dist/opentype.js), & it will be definitely faster too.

src/types.js
Jolg42: @Harbs After searching for the best one in term of performance it looks like it's more complicated than that…

> Usually there is no reason not to use push - the method is there for exactly that operation and makes some code easier to read. While a few people think one version is faster than the other, both are equally optimized in browsers. See Why is array.push sometimes faster than array[n] = value? and Using the push method or .length when adding to array? - results vary so wide that it's actually irrelevant. Use what is better to understand.

from http://stackoverflow.com/a/21034712/1345244

> Actually push is just as fast as property assignment. Good measurements are rare, but those that are done properly show varying results across different browser engine versions - changing rapidly and unexpected. See Appending to array, Why is array.push sometimes faster than array[n] = value? and Is there a reason JavaScript developers don't use Array.push()? - the conclusion is that you should use what is most readable / appropriate for your use case, not what you think could be faster.
from http://stackoverflow.com/a/18477047/1345244

Also, I know that AirBnB did a good job with this public style guide https://github.com/airbnb/javascript/blob/master/es5/README.md#arrays and use `.push()`.
In the end maybe the performance is not really easy to compare here and readability is more important?

fpirsch:
The fastest way is to pre-allocate the array with `new Array(length)`
What is slow is changing the size of arrays.
I did not have a close look but it seems to me that in this method the array length is known from the start, so it could be preallocated.

Jolg42:
@fpirsch 100% for pre-allocating the arrays 👍  but for direct vs push it looks more complicated no?

fpirsch:
@Jolg42 
If the array is preallocated, you can't use `push` anyway ;-)

@Harbs 
> On second thought, I'm not sure the JSDocs that I did are totally the right way to go.

JSDoc is basically a port of JavaDoc, but JS is not Java. JS is not about classes and inheritance, so in most practical cases JSDoc is awkward. Anyway JSDoc is better than no doc at all :-)

This is why I only use basic JSDoc syntax, it's a simple and familiar way to document things, but not intended to compile a real JSDoc (which would, btw, be mostly useless).

Harbs:
Cool.

Feel free to revert my changes vis a vis push(), but push() is almost always slower in most browsers.

About JSDoc: I would like to restructure the modules a bit to make JSDoc and exporting Google externs cleaner. externs are pretty important for using the library in strongly typed environments. If that's okay with you, I'll work on that.

It will not be major restructuring, but there's certain structures in the code that does not work so well with JSDoc and extern tools.

Harbs:
BTW, re. array performance, it's likely that using TypedArrays for a lot of the calculations would be more performant and less memory intensive.

Jolg42:
So in the end we will keep your array changes and I will check one day for non pre-allocated arrays ;)
And yes TypedArrays can be great at performance and already used in the lib for some parts 👍

fpirsch:
My experience with typed arrays has always been very disappointing with regards to speed in V8 (Chrome, Node). But I did not check recently if things have changed.

Harbs:
Here's an interesting SO discussion on the topic: http://stackoverflow.com/questions/24853686/javascript-typedarray-performance Both answers are enlightening.

BTW, I regenerated the JSDocs using the original src files, and the results are much better.
http://in-tools.com/stuff/opentype.js/

I'm also still looking into how to structure the code so externs can be generated correctly.

fdb:
@Harbs this looks very good, thanks! 

I would like to see one improvement: keep all the direct assignment / vs push stuff for a separate pull request (I prefer if possible to pre-allocate the array and use direct assignment as @fpirsch mentioned).

PR #212, opened 2016-08-01
Support glyf table component offset by matched points

moyogo:
See https://www.microsoft.com/typography/otspec/glyf.htm
> Argument1 and argument2 can be either x and y offsets to be added to the glyph or two point numbers. In the latter case, the first point number indicates the point that is to be matched to the new glyph. The second number indicates the new glyph's “matched” point. Once a glyph is added, its point numbers begin directly after the last glyphs (endpoint of first glyph + 1).

### DIFF
--- a/glyph-inspector.html
+++ b/glyph-inspector.html
@@ -151,7 +151,11 @@ function displayGlyphData(glyphIndex) {
     } else if (glyph.isComposite) {
         html += '<br>This composite glyph is a combination of :<ul><li>' +
             glyph.components.map(function(component) {
-                return 'glyph '+component.glyphIndex+' at dx='+component.dx+', dy='+component.dy;
+                if (component.matchedPoints === undefined) {
+                    return 'glyph '+component.glyphIndex+' at dx='+component.dx+', dy='+component.dy;
+                } else {
+                    return 'glyph '+component.glyphIndex+' at matchedPoints=['+component.matchedPoints+']';
+                }
             }).join('</li><li>') + '</li></ul>';
     } else if (glyph.path) {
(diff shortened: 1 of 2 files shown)


Jolg42:
Hi @moyogo 

It looks good! Do you have a specific font to test this case?

moyogo:
@Jolg42 Some versions of Molengo still have it like I originally made it with matched points.
For example: https://github.com/Jolg42/FontSquirrel-Fonts/tree/master/Fonts/Molengo/web%20fonts/molengo_regular_macroman

Jolg42:
I just checked and it was working for the first glyph (77)
screenshot 2016-08-05 at 19 07 49

But the second (110) have an error here https://github.com/moyogo/opentype.js/blob/be592efebf0b522494725b8e23d655538e7bbabf/src/tables/glyf.js#L180
screenshot 2016-08-05 at 19 10 19

moyogo:
@Jolg42 Hmm, interesting. Sorry for being naïve about the version by FontSquirrel.

I placed the true original there https://github.com/moyogo/molengo-fonts/releases/tag/v0.10.
It doesn’t trigger the bug.

The FontSquirrel version is messed up. It was autohinted with ttfautohint and the second matchedPoint index is incorrect. In the original in Agrave, the gravecomb component has matchedPoints=[11, 5] but in the FontSquirrel ttfautohint version matchedPoints=[12, 6]. The first matchedPoint (indexed on the composite glyph) is correct since ttfautohint adds a component with one point (bumping +1 all previous point indeces), but the second matchedPoint (indexed on the component glyph) should stay the same since no index changed in that simple glyph.

I added a commit to throw on the error. Is that OK?

Jolg42:
👍

PR #218, opened 2016-08-25
Take optional filename argument in download

dfcreative:
To redefine path where font is generated, for example, `font/fontName.otf` to place font to a directory


fdb:
Thanks!

PR #219, opened 2016-08-25
Add support for letter spacing / tracking

lachmanski:
HTML Letter-spacing is added or subtracted between every characters in a string of text. Letter Spacing is usually declared in ems. Graphics Software uses the term Tracking, and [multiplies the ems by 1000](http://stackoverflow.com/a/36612356/6465140) to avoid working in decimals.

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -238,8 +238,7 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback) 
 
         if (options.letterSpacing) {
             x += options.letterSpacing * fontSize;
-        }
-        else if (options.tracking) {
+        } else if (options.tracking) {
             x += (options.tracking / 1000) * fontSize;
         }
     }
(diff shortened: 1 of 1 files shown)


fdb:
Looks good, thanks!

Sorry for the late reply — I was on holiday.

src/font.js
```diff
@@ -235,6 +235,12 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback)
             var kerningValue = this.getKerningValue(glyph, glyphs[i + 1]);
             x += kerningValue * fontScale;
         }
+
+        if (options.letterSpacing) {
+            x += options.letterSpacing * fontSize;
```
andreisocaciu: Why multiply `letterSpacing` by `fontSize`?
If I try to get a path of a text with letterSpacing of 3px and a font size of 100px, this code will give me a spacing if 300, which is clearly not what I want. Am I missing something?
@lachmanski @fdb

PR #235, opened 2016-10-12
Encode variation adjustment deltas

brawer:
This is a port of Python code which I’d written a while ago for
fonttools. It encodes a structure that is part of the `gvar` and
`cvar` tables.  Actually I had thought I’d sent this out for review
back when I’d written the code, but it looks like it has been silently
sitting in my git client.

Sadly, I won’t have time to contribute a full `gvar` implementation to
OpenType.js, but perhaps you’ll find this small piece still useful.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -583,6 +583,122 @@ sizeOf.MACSTRING = function(str, encoding) {
     }
 };
 
+// Helper for encode.VARDELTAS
+function isByteEncodable(value) {
+    return value >= -128 && value <= 127;
+}
+
+// Helper for encode.VARDELTAS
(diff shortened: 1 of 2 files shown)


fdb:
That's super-useful! Thanks @brawer!

fpirsch:
Fails because the tests use `Array.prototype.fill` (ES6 feature) on node v0.10.47.
The node version could be updated in `.travis.yml`
Or the tests could be rewritten using a small fill function. This is no big deal.

brawer:
After switching to node version 4, the tests are now passing on Travis.

PR #237, opened 2016-10-23
Fixes font loading error callback not fired on Chrome if response is 502

create3000:


fdb:
This is very useful, however the checks currently fail because of formatting issues.

Could you look at correcting the formatting? You can run `npm test` to see where the algorithm falls over.

Jolg42:
Merged manually in https://github.com/nodebox/opentype.js/pull/237

PR #246, opened 2017-01-04
Convert the path to a SVG <path> element as a html element

afternoon2:
I added a function, that returns an svg path as an html element, so you can add your newly rendered path straight into specific DOM element, without struggling with string parsing.

[Codepen Demo here.](http://codepen.io/jakub_antolak/pen/zNYMeK)

### DIFF
--- a/src/path.js
+++ b/src/path.js
@@ -246,4 +246,13 @@ Path.prototype.toSVG = function(decimalPlaces) {
     return svg;
 };
 
+Path.prototype.toHTML = function(decimalPlaces) {
+  var temporaryPath = this.toPathData(decimalPlaces);
+  var newPath = document.createElementNS('http://www.w3.org/2000/svg', 'path');
+
+  newPath.setAttribute('d', temporaryPath);
+
+  return newPath;
+};
+
 exports.Path = Path;
--- a/src/path.js
+++ b/src/path.js
@@ -246,13 +246,13 @@ Path.prototype.toSVG = function(decimalPlaces) {
     return svg;
 };
 
-Path.prototype.toHTML = function(decimalPlaces) {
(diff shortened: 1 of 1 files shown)


fpirsch:
Hi @afternoon2,
this is a good idea.
The automated tests fail because of the indentation (please run `npm test` before committing).
The name `toHTML` suggests the function returns HTML code. Since it returns an SVG element, maybe it should have another more explicit name ?
Note also that `opentype.js` is intended to work also in non-browser environments (e.g. Node.js).

fdb:
I agree regarding the naming. Perhaps `Path.toElement` or `Path.toDOMElement` are more appropriate names.

afternoon2:
Yes, the name might be misleading. `Path.toDOMElement` is much better. My intention was to make a method that would be helpful while coding svg animations (live rendering svg from font), so it's not useful outside the browser.

I have some problems with `npm test.` I mean -after I corrected indentation, in git-bash I'm still seeing some linebreak issues (49):

`validateLineBreaks: Invalid line break at <filepath>`

And here's a dummy question: how to fix it without opening each broken file?

fpirsch:
Yes, this is a pain if you are not using Linux.
To avoid this, I locally add the following line to the `.jscs` file before running `npm test`:
```
"validateLineBreaks": "CRLF",
```
Then I get only 1 error for `./bin/server.js`
Line break errors don't matter since they will be automatically and silently corrected when you commit on github (don't commit `.jscs`). You just have to make sure you don't have other types of errors.

Maybe jscs offers the possibility to disable linebreak checks but I don't know how.

afternoon2:
That works, thank you!

fdb:
Thanks!

Issue #245, opened 2016-12-30
TypeError: subtables[i] is not a function at table.getKerningValue


caub:
```
TypeError: subtables[i] is not a function
    at table.getKerningValue (VM218 opentype.js:5536)
    at Font.getKerningValue (VM218 opentype.js:914)
    at fetch.then.then.then.font (VM491:60)
```

https://jsfiddle.net/crl/2o3sfhrg/1/

This error happens for Ek Mukta, Cardo, Cousine, Bungee Inline, Jomhuria,  Aref Ruqaa, NTR, Gurajada, Buenard fonts too

fpirsch:
Hi,
thanks for the report. It happens with Raleway too. I will fix this soon.

Issue #236, opened 2016-10-14
Add ligature rendering in canvas


Jolg42:
Ligatures are supported but not rendered in canvas right now.

We will need to
- [ ] Add a ligature boolean for [GlyphRenderOptions](https://github.com/nodebox/opentype.js/blob/master/src/font.js#L206)
- [ ] Check all types of ligatures for a given string (maybe only a subset or the active ones for the first version?)
- [ ] Render the ligature in canvas
- [ ] Other modifications?

fpirsch:
I'm on it ;-)

Issue #230, opened 2016-10-07
Remove planned ligature support from readme.


mrapogee:
According to #96, ligature support exists, correct? In the readme, it says ligatures are planned.

fpirsch:
There is some ligature support indeed. Ligature data is read from and written to font files. The thing is, this data is not yet used for rendering.
Still, the readme could be updated to reflect this :-)

mrapogee:
Ah, I see. :+:

Do you know what it take to support ligature rendering? I don't see an open issue for this.

Jolg42:
Hi @mrapogee 

I just opnened the issue https://github.com/nodebox/opentype.js/issues/236 for ligatures rendering 😃 

Feel free to contribute!

PR #250, opened 2017-01-10
Add ligature rendering

fpirsch:
Add GSUB "liga" and "rlig" rendering support.
Solves #236, #230 and also #245.

Glyph rendering options are added to enable each GSUB feature individually.
Only "liga" and "rlig" are supported, and enabled by default as the spec suggests.
A "ligatures" checkbox is added to `index.html`.
Plus a small layout improvement to the glyph-inspector.

This is open for comments, especially on the options API :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ Create a Path that represents the given text.
 
 Options is an optional object containing:
 * `kerning`: if true takes kerning information into account (default: true)
+* `features`: an object with [OpenType feature tags](https://www.microsoft.com/typography/otspec/featuretags.htm) as keys, and a boolean value to enable each feature.
+Currently only ligature features "liga" and "rlig" are supported (default: true).
 
 _Note: there is also `Font.getPaths` with the same arguments which returns a list of Paths._
 
@@ -252,7 +254,7 @@ Convert the path to a SVG &lt;path&gt; element, as a string.
 
 Planned
 =======
-* Support for ligatures and contextual alternates.
+* Support for contextual alternates.
 
 Thanks
 ======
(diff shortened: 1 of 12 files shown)


Jolg42:
Good job! Looks great and the options.features easy. I need to try this soon 👍 

Performance-wise what did you found? Is it fast enough even with a lot of ligatures?

fdb:
`options.features` looks perfect! Great job!

fpirsch:
I haven't tested for performance, but it's probably very poor :-(
2 problems here:
 - there are so many types of substitutions, and so many ways to declare them in lookup tables, including multiple lookup tables, overlapping each other with duplicates... I'm a bit in the dark here, without a global comprehension including complex scripts and features.
- you can't have the same data structures and algorithms for rendering speed (read-only) and ease of edition (list, add, remove...). Or if you can, it's really not obvious.

I chose to sacrifice perf in favor of a more general approach including font editing. The test fonts I used contained very few liga substitutions (about 10 max).

fpirsch:
BTW I just noticed PR #228 already proposed a fix for #245.

Jolg42:
@fpirsch I see!... No problem for the performance if it's good enough for a few substitutions but if it's really slow (reading) for big fonts we probably should note it somewhere and put an example to deactivate it. (Or deactivating it by default is an option too).

fpirsch:
@fdb would you be OK for a merge & new release (including gh-pages for the online demo) ?

fdb:
Sorry for the late reply. I'll make a release now.

The GitHub pages now come from master, so no the gh-pages branch is no longer needed.

PR #251, opened 2017-01-12
Upgrade Travis configuration to node.js version 6

brawer:
Version 0.10 has reached its end of life and is not maintained anymore.
https://github.com/nodejs/LTS#lts-schedule

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '0.10'
+- '6'
 deploy:
   provider: npm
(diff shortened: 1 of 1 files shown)


Jolg42:
Hi!

We're stuck on v4 here (not our fault 😓)... 
So do you think it's better to add the v4 or just replace v6 by 4?

fdb:
For me, the lower the better. So using version 4 is fine for me as well.

fpirsch:
The Travis job should be retriggered to check #235 again.

PR #259, opened 2017-03-05
Development server should ask for no caching

axkibe:


axkibe:
Cache-control: no-cache breaks the sourcemaps, but maxage=0 is fine for chrome.

There is still an auto-update issue with watchify, but IMO thats some watchify bug.

fdb:
👍  Thanks!

PR #262, opened 2017-03-24
fontSize shouldn't be passed as string

axkibe:
In my opinion fontSize should not be passed as string encoded value to opentype.js

In the current implementation it doesn't matter, since as far I can see, it's only used in arithmetic operations where javascript will do an implicit conversation. However, it will be done each time and is considered relatively expensive.

It creates hugh issues in my upcoming fonthinting implementation, as there are fonthinting programms that compare to fontsize and branch depending on it and thus fail if integer !== string.

I leave it up to your decission if you want to typecheck in opentype.js and throw an error if typeof(fontSize) !== 'number' or it's just considered iligitimate usage where a user gets unexpected results if not using the API correctly. I'll do a check at least in the fonthinting part, as this can be a very subtle and nasty to find issue, as it isn't obvious, it just looks suboptimal, when the wrong font hinting branches get executed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,7 +115,7 @@ function ligaturesChanged(e) {
 }
 
 function fontSizeChanged() {
-    fontSize = fontSizeSlider.value;
+    fontSize = parseInt(fontSizeSlider.value, 10);
     document.getElementById('fontSize').innerHTML = '' + fontSize;
     renderText();
 }
--- a/src/font.js
+++ b/src/font.js
@@ -309,8 +309,9 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback) 
  */
 Font.prototype.getPath = function(text, x, y, fontSize, options) {
     var fullPath = new path.Path();
+	var self = this;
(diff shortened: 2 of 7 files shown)


Jolg42:
Hi @axkibe 
I agree that fontSize should not be a string, I think we can do some sort of parseInt() or parseFloat() (As I think that the CSS recommendation allows floats) and if the value is NaN throw an error with a specific message.
But this needs to be done for all methods in glyphs.js and font.js to be consistent.
Maybe somebody has a better idea?

axkibe:
This is a design philosophical decission, personally I'm not a hugh fan of silently fixing things, but rather throwing an error, so the caller is made aware how to use the API correctly...

However, thats personal preference, some people like the "let me fix this for you" approch better..

Jolg42:
Yes, less code is better code...! Maybe a simple parseFloat instead of parseInt would do the job + a small comment too to clarify that.

axkibe:
Awww github appened the fonthinting pull request to this originally trivial one. I'll make a new one

PR #263, opened 2017-03-27
Truetype Font Hinting

axkibe:
Implementing a font hinting interpreter for truetype:

Demo:

The actual canvas is the top one, the lower one has each pixel enlarged by a factor of 7.
The font used in the demo is DejaVuSans.ttf
First line is default opentype.js
Second line is opentype.js with fonthinting enabled
Third line is canvas.fillText() with @font-face.

hintingdemo

As you see there is hardly a difference between second and third line, except glyph advancement not being hinted. I didn't do that yet, as that would require some changes to the opentype.js.

I tried to keep all changes to opentype.js outside of src/hintingtt.js as minimal as possible. 

PS: I'll need to append a fix commits until the code styling checks stop complaining.

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,7 +115,7 @@ function ligaturesChanged(e) {
 }
 
 function fontSizeChanged() {
-    fontSize = fontSizeSlider.value;
+    fontSize = parseInt(fontSizeSlider.value, 10);
     document.getElementById('fontSize').innerHTML = '' + fontSize;
     renderText();
 }
--- a/src/glyph.js
+++ b/src/glyph.js
@@ -126,7 +126,8 @@ Glyph.prototype.getPath = function(x, y, fontSize, options, font) {
     x = x !== undefined ? x : 0;
     y = y !== undefined ? y : 0;
     fontSize = fontSize !== undefined ? fontSize : 72;
-    var commands, hPoints;
+    var commands;
+	var hPoints;
     if (!options) options = { };
     var xScale = options.xScale;
     var yScale = options.yScale;
--- a/src/glyph.js
+++ b/src/glyph.js
@@ -5,6 +5,7 @@
 var check = require('./check');
 var draw = require('./draw');
 var path = require('./path');
+var glyf = require('./tables/glyf');
 
 function getPathDefinition(glyph, path) {
     var _path = path || { commands: [] };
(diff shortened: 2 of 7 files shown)


axkibe:
Finally a green check :)

src/glyph.js
```diff
@@ -118,19 +119,41 @@ Glyph.prototype.getBoundingBox = function() {
...
+        hPoints = font.hinting.exec(this, fontSize);
+    }
+
+    if (hPoints) {
+        // in case the hinting engine failed revert to
+        // plain rending
+        commands = glyf.getPath(hPoints).commands;
+        x = Math.round(x);
+        y = Math.round(y);
+        // TODO in case of hinting xyScaling is not yet supported
```
Jolg42: I don't understand this TODO, could you explain it a little bit more?

src/hintingtt.js
```diff
@@ -0,0 +1,3043 @@
+/* A TrueType font hinting interpreter.
+*
+* (c) 2017 Axel Kittenberger
+*
+* This file is covered by the MIT (expat) License,
+* see the LICENSE file in the root of this project.
```
Jolg42: I don't if it's relevant to add the mention to the MIT license as the LICENSE file is here for that

index.html
```diff
@@ -115,7 +115,7 @@
 }
 
 function fontSizeChanged() {
-    fontSize = fontSizeSlider.value;
+    fontSize = parseInt(fontSizeSlider.value, 10);
```
Jolg42: Just one question, is it doable to replace `parseInt` by `parseFloat`?

Jolg42:
Your screenshot comparing the hinting looks great! :+1: 

I see that hinting is not activated by default, `Font.prototype.getPath` needs `options.hinting` to be true, I think it's the best to not impact memory consumption & performance for people that don't need it!

By the way, you don't have to push commits to check code styling, it's a lot easier to do `npm run test` (you can look at scripts, in [package.json](https://github.com/nodebox/opentype.js/blob/master/package.json))

->You worked a lot! Thank you for that contribution! :smiley:

src/glyph.js
axkibe: The options object can have an xScale or yScale setting to stretch the paths, if hinting is enabled it is currently ignored.

src/hintingtt.js
axkibe: I agree it's not important. You can cut it if you want to. It's oft generally recommended a good idea for every file tough. src/opentype.js has for example.

index.html
axkibe: Sure! I'll append it.

fdb:
Thanks, this looks awesome!

axkibe:
Exactly, thats why I made it off by hinting default. It's only relevant between aprox. 8ppem and 60ppem, below it's a blur anyway and above normal subpixels are beautiful enough. Oh and I noted some font programs turn themselves off.

PS: This all applies only to truetype, postscript fonts have a completely different hinting engine.

axkibe:
PPS: DejaVuSans has one of the best font hinting programs I encountered, other fonts the difference may not be that staggering.

Jolg42:
As it is now merged, next step could be editing the README & add a short description of current hinting with an example :)

fdb:
Good point! @axkibe could you provide an example for the README?

axkibe:
OK.

Jolg42:
@fdb The git history is a little verbose now, maybe you can try to squash next time @fdb :wink:

fdb:
Oh right the code styling commits. Too late now, thanks for reminding me :blush:

axkibe:
Im sorry.

fdb:
@axkibe no problem, don't worry about it 😄

index.html
Jolg42: Sorry for that, I checked and as the slider is configured to only output integers parseFloat is not needed...!

src/glyph.js
Jolg42: Sure!

Jolg42:
@fdb @axkibe I added the hinting checkbox for http://opentype.js.org/ to try it ;)

axkibe:
Cool, unfortunially the default FireSansMedium.woff is a postscript font, so it works when loading a ttf font. But then it does :)

axkibe:
I noticed DEBUG is turned on the live website.
This seriously dampens perfomance.

Jolg42:
@axkibe  Fixed :) I forgot to rebuild the dist files....

Jolg42:
@axkibe so maybe we could disable the hinting checkbox for non ttf fonts

PR #282, opened 2017-04-20
Add GSUB and kern output in font-inspector

Jolg42:


fdb:
Very useful, thanks!

Also, I'm currently reworking the font and glyph inspector to integrate them into one thing, with drag-and-drop, so it would be easy to inspect any font just by dropping it on the opentype.js homepage.

Jolg42:
Yes! A drag and drop (or click to select the old way) would be great!

PR #283, opened 2017-04-20
fixing rounding issues (as tested on DejaVuSans-Regular)

axkibe:
I saw some not so nice rendering differences in DejaVuSans-Regular and debugged them.


src/hintingtt.js
```diff
@@ -79,14 +79,14 @@ function roundToGrid(v) {
 * Rounding to double grid.
 */
 function roundToDoubleGrid(v) {
-    return Math.sign(v) * Math.round(Math.abs(v / 2)) * 2;
+    return Math.sign(v) * Math.round(Math.abs(v * 2)) / 2;
```
fpirsch: There is no need to separate sign and absolute value with round().

axkibe: "In TrueType, rounding is symmetric about zero" - https://developer.apple.com/fonts/TrueType-Reference-Manual/RM05/Chap5.html

For example -7.5 is supposed to round to -8, not -7 as Math.round(-7.5) would yield

PR #287, opened 2017-04-27
Fix "Object x already has key" error in Safari

neiltron:
## Description

Changed `entriesToObject` to only throw an error if the pre-existing key isn't NaN.

## Motivation and Context

Safari throws an error when calling `opentype.load` on a font file that has already been loaded. This might only happen with certain types of fonts.

Related issue: https://github.com/nodebox/opentype.js/issues/281

## How Has This Been Tested?

I ran this against an existing project and checked that the opentype.js dev site still worked after making the change. Also ran `npm run test`.

## Types of changes

## Checklist:


fdb:
This is now in 0.7.3, which should be pushed soon to npm.

PR #288, opened 2017-05-02
remove DEBUG code from minified distribution

axkibe:
## Description
remove DEBUG code from the minified distribution

## Motivation and Context
make the minified distribution smaller

## How Has This Been Tested?
Both versions work as expected with my project.
The minified version doesn't include the DEBUG upon inspection.

PS: I had to move DEBUG to exports.DEBUG otherwise an exception is raised in the non-minified version of using an undefined global variable. It cannot be defined with 'var DEBUG' tough, since then uglify will not remove it as it detects a shadowed variable. I could remove 'use strict' to simply avoid the exception, however e.g. V8 is said to be aprox. 20% faster in 'use strict' mode, so that would be a bummer, just to remove it for that.


Jolg42:
@axkibe Good idea!
By the way, are you using `envify` to set DEBUG to true? Or maybe something else?

axkibe:
No envify, it's just an additional option to the already existing uglify to remove those code parts.

To set DEBUG to be actually true one has to change the code anyway. Usually I do this not globally but only around the problematic part.

Jolg42:
Ok I see!
Could you add a small example here on how to set DEBUG to true here?

axkibe:
I just add a codeline to a place where I see it fit, often for example in hintingtt, line 647 in execGlyph():
```exports.DEBUG = true;```
since the FPGM and PREP programm are already executed there, which are often not interesting when looking at particular glyph hinting anomality.

Some fonts with large hinting programs (I'm looking at Roboto for example with its multiple function calling taking over thousands steps for a glyph) I added a condition to it when I had already an idea where it gets problematic, like in exec() ````if (state.ip >= 57) exports.DEBUG= true;```

PR #289, opened 2017-05-08
Fix unclosed paths for TrueType glyphs.

jasondavies:
TrueType glyph outlines were not being closed properly; only a single closePath command was being issued for each glyph.  This fix ensures that a closePath command is issued for each contour.


fdb:
Looks good, thanks!

Issue #260, opened 2017-03-10
Tree-shaking ready


kononenko-a-m:
Hello, I try to make a library based on opentype, but also I want to reduce size build size, for example, if I gonna use it only for WOFF, then no code for TTF, SVG parsing needed, and it can be removed from target build (which is very desirable) using tree-shaking approach. Is there any chance that you would deliver this? Or maybe I just can prepare a PR which would introducing support for tree-shaking?

Jolg42:
Hi @kononenko-a-m 
So feel free to prepare a PR, just keep in mind that OpenType.js needs to run in the browser and Node.js ;) (Node.js as a package and in the browser with a browserify build).

axkibe:
I suppose this is in a way a duplicate of #247 
since all "treeshaking tools" I saw require an ES6 style export/import scheme.

fdb:
I have absolute no issue with this. Moving to the ES6 style export / import scheme would work for me. @kononenko-a-m if you want to work on this, feel free! I would prefer two pull requests:

- one to move to ES6 modules
- one to support tree shaking

Jolg42:
I guess it's already done as we now use Rollup.js

Issue #247, opened 2017-01-06
Uplift to ES6 classes and rollup bundling?


Pomax:
The codebase right now is actually extremely well organized and would be relatively easily uplifted to using ES6 classes, which would make every class a potential module that can easily be used by other open source projcets, or customized for opentype experiments =)

fpirsch:
Aren't ES6 classes the new "bad part" of JS ?

Pomax:
Subjectively: to some. Objectively: no.

As with everything in JS, you need to know how it works before you use it, because ES6 classes are *not* the same as "class based object oriented programming" in things like C# or Java. That's not "good" or "bad", that's just "welcome to JavaScript" (yet again - there are so many things in JS that you're going to do wrong if you think it does things like other languages you're familiar with).

In this case: the class syntax is an incredibly powerful way to implement *prototype inheritance* without  the pages and pages of code to do that well. It offers single word syntax where before we needed 10 lines of codes, and abstracts away all the "safety code" we need for raw prototype inheritance to ensure the right prototype's properties are accessed. So as long as your write your codebase fully aware of that, there is literally nothing bad about ES6 classes: they make the code infinitely more legible, and in this codebase specifically, infinitely easier to write things like plugins that introduce additional table parsing by making it trivial to write your own `X implements SFNTTable` or `X implements CommonTable`.

fpirsch:
Isn't it precisely the point: keywords like `class` and `implements` make things look like java or C++ classes, and everybody wants to use them because it feels like good old java OOP, but nobody really understands what it is exactly.
Btw, :+1: for plugins.

oriSomething:
@fpirsch If you already use `prototype`, `class` is a great way to organize this code in simple way.
You can claim that you can't create truly multiple inheritance, but on the same time many will claim that inheritance is bad and multiple inheritance is multiple bad.
Another thing, many IDEs do really well autocomplete only from using the `class`.

fdb:
Personally, I like ES6 classes. As @oriSomething mentioned, we're using `prototype`-based inheritance already anyway, so it's just renaming some things and cleanup. 

What I *don't* like is Babel, but yeah, necessary evils and all that....

My favorite feature of ES6 is to use `let` and `const` everywhere instead of `var`. So, once we're using Babel I would like to go through the code and do var replacement everywhere.

Pomax:
[...]

PR #290, opened 2017-05-12
Use ES6 modules.

fdb:
Convert the old require-style modules to native ES6 modules.

This is a (partial) fix for #260 and preparation for #247.


fdb:
Thanks for checking it out! I'm testing a bit deeper for interop within node.js, once that works I'll merge.

fdb:
Node.js doesn't support ES6 imports yet, so I've updated `package.json` to refer to the precompiled `dist/opentype.js` (same as the one we use in the browser).

This works, but apparently adding a ES2015-compatible `module` property in `package.json` is also a good idea ([read why here](https://github.com/rollup/rollup/wiki/pkg.module)). However, I don't have anything to test this with, so I'm keeping it as-is for now.

Jolg42:
I see! I like the idea of adding an ES2015-compatible `module` to support older Node.js versions.

So if we do this we'll need to watch & build the dist file when debugging for Node.js for example? We should add a note and a simple `npm run` command for that I think.

fdb:
I've added a `npm run watch` command to watch for changes. I've also switched to Rollup which made working with source maps easier, and also adds support for tree shaking, which should result in smaller builds.

Jolg42:
Tree-shaking is cool :+1: 
Could you add https://github.com/mjeanroy/rollup-plugin-license or I could add it later, to keep the header in dist files?

axkibe:
Nice, proper tree shaking will require API changes in a future step, however.

Since now, much behavior is decided by calling options. like load() can take any extension and thus has all font formats linked to it, the tree shaker has to include all. loadTTF() loadPS() or so would only be able to do those formats and thus only included that code trees. One could still provide an overall load() that selectes the other load functions.

Similar thing with fonthinting. It is activated by options.hinting = true, as command line option. So even somebody who never wants to do it, has the code in its build even when shaken. Here an option that actually requires importing an object would be better, since only then the trees would be linked.

fdb:
@axkibe I agree -- we need some way to check for a font signature (e.g. `wOFF`), then dynamically require that code in a separate module.

fdb:
@Jolg42 I've added the license header.

axkibe:
I'd say rather you will have to have the user make it. And provide a catch-all loading function only for those that don't care if their bundle is going to include it all, even if not needed.

fdb:
@axkibe If people want to use this in their own projects we can make it so that they only import the correct modules (e.g. no tiny-inflate if they just use desktop fonts). That will have to wait for another PR, though.

PR #291, opened 2017-05-16
Remove erroneous use of `exports`.

jasondavies:


fdb:
Hi @jasondavies thanks for catching these!

Issue #168, opened 2015-12-12
Unicode values outside Basic Multilingual Plane cannot be processed correctly.


lixiangliu:
For example:

`font.stringToGlyphs('\u{2f894}')` will return an array with two “.nodef” in it. 
<img width="419" alt="2015-12-12 21 26 42" src="https://cloud.githubusercontent.com/assets/10280733/11762111/0fcb0a76-a117-11e5-98a3-87f744a44504.png">

If `font.glyphs.glyphs[...]` is being used to access a certain glyph directly, then it will return an object with an undefined unicode value. 
<img width="527" alt="2015-12-12 21 26 59" src="https://cloud.githubusercontent.com/assets/10280733/11762112/1e584752-a117-11e5-9bfb-8dd1495fbcda.png">

lixiangliu:
It seems this is the limitation of cmap format, since format 4 only supports 16-bit unicode value.

fdb:
Would it help if we parsed a different cmap format?

lixiangliu:
In that case, cmap format 12 is probably a good choice then, it’s widely adopted and the superset of format 4.

Jolg42:
@lixiangliu cmap 12 read support was just added with PR #207 😉

fdb:
Closing this issue because of the cmap 12 support. @lixiangliu  let me know if there's something else that blocks you from working wit these fonts.

amart1n:
The same issue still exists with values outside the BMP. Looks like it may be because of the limitations of `charCodeAt()` https://mathiasbynens.be/notes/javascript-unicode#code-point-from-string

Jolg42:
@amart1n You're right `charCodeAt()` "will always return a value that is less than 65536." ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt))

[codePointAt()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) is the new method to use but because of the compatibility we would need to add Mathias's polyfill: https://github.com/mathiasbynens/String.prototype.codePointAt and use it instead of `charCodeAt()` to have a full Unicode support.

fpirsch:
Closed through #338

Issue #193, opened 2016-06-13
Add support for openType SVG color fonts


Mottie:
It looks like [Microsoft is adopting the openType SVG color fonts](https://lists.w3.org/Archives/Public/public-svgopentype/2016Mar/0014.html) format, so adding support for the SVG `colr` and `cpal` objects would be necessary.

I was trying to see if I could figure it out myself, but I am just now learning how to process `ArrayBuffer` objects and `DataViews`. That combined with figuring out an unfamiliar code base would take me a long time. I won't stop trying to learn, and I'll still work on adding this feature in my free time, unless someone beats me to it.

Useful links:
- [W3C Spec](https://www.w3.org/2013/10/SVG_in_OpenType/).
- [`SVG`](https://www.microsoft.com/typography/otspec/svg.htm) table details (since it doesn't appear to be included in this library).
- SVG [`COLR`](https://www.microsoft.com/typography/otspec/colr.htm) (color) table details.
- SVG [`CPAL`](https://www.microsoft.com/typography/otspec/cpal.htm) (color palette) table details.
- Color Gecko Emoji [ttf](http://people.mozilla.org/~jkew/opentype-svg/GeckoEmoji.ttf) or [woff](http://people.mozilla.org/~jkew/opentype-svg/GeckoEmoji.woff) for testing ([demo](http://people.mozilla.org/~jkew/opentype-svg/GeckoEmoji.html)).
- [Awesome examples](https://hacks.mozilla.org/2014/10/svg-colors-in-opentype-fonts/).

mqudsi:
I believe this is required for opentype.js to support emoji fonts.

Jolg42:
@mqudsi Yes but actually it depends on the platform as Google's emojis are in `CBDT` `CBLC` tables, Microsoft's `COLR` `CPAL`, Apple's `sbix`

talltyler:
Supporting a single standard is a whole lot better than not supporting any of them. 
As a first step I would suggest having the forEachGlyph function to return information about what character it can't render allowing things like emojis to be handled as exceptions in user code rather than returning glyph 0.

Jolg42:
@talltyler the problem you're describing needs to be fixed, check the issue https://github.com/nodebox/opentype.js/issues/168 for more info

talltyler:
[...]

PR #297, opened 2017-06-16
Addressing issues #168 related to not supporting multi byte characters

talltyler:
JavaScript strings handle two byte characters as two items which breaks some multi lingual and emoji support.

## Motivation and Context
This addresses https://github.com/nodebox/opentype.js/issues/168

## How Has This Been Tested?
npm test passes. And this has been tested in IE 11 and Chrome.

## Types of changes
Mostly this just uses a for...of loop if the browser enables it. Because this is a new feature checks are made to make sure this method is available, and the build tools have been updated to support these changes.

## Checklist:


schuylr:
I tried your branch with a non-BMR character (this one: 🙂 ) on a font I made with a flat vector path (no colour like #193 mentions), and I still get a blank square. I'm using `getPath` and `toPathData` on the response.

talltyler:
This branch doesn't solve any of the rendering issues these types of glyphs. It only allows multibyte characters to be passed through the system and handled as single units rather than two unrelated code points.

schuylr:
I understand that, but I made a custom font with the same sort of glyphs that you would find in the ASCII/BMR range and it still didn't work. The result was still a blank square.

talltyler:
Can you post this font and the character that you are passing in?

schuylr:
[Cursive.ttf.zip](https://github.com/nodebox/opentype.js/files/1459006/Cursive.ttf.zip)

The character I tried is above. "slightly smiling face" or `U+1F642`

fpirsch:
Closed through #338

PR #303, opened 2017-09-04
Node.js LTS  support

pit-rpg:
Node.js LTS  support

## Description
replace ES7 module syntax with commonjs syntax

## Motivation and Context
Node.js LTS  support

## How Has This Been Tested?
manually on Node.js

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -32,4 +32,10 @@ function checkArgument(expression, message) {
     }
 }
 
-export { isBrowser, isNode, nodeBufferToArrayBuffer, arrayBufferToNodeBuffer, checkArgument };
+module.exposrts = {
+    isBrowser: isBrowser,
+    isNode: isNode,
+    nodeBufferToArrayBuffer: nodeBufferToArrayBuffer,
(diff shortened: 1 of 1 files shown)


src/util.js
```diff
@@ -32,4 +32,10 @@ function checkArgument(expression, message) {
     }
 }
 
-export { isBrowser, isNode, nodeBufferToArrayBuffer, arrayBufferToNodeBuffer, checkArgument };
+module.exposrts = {
```
fdb: I think you have a typo in "exports".

fdb:
Although we use the new import syntax, OpenType.js *should* work in Node.js LTS because we include a [main file](https://github.com/nodebox/opentype.js/blob/1b823458ac97187b4b42d773fb9166ec4c0924d2/package.json#L22) in `package.json`. This should load the compiled version for use in Node.js.

Can you point out if you have a specific use case where this breaks down?

Jolg42:
Unactive PR, closing for now, feel free to add more details and we'll open it again.

PR #312, opened 2017-10-21
Fixing 2 issues when hinting "mutlu.ttf"

axkibe:
Before fix:

mutluhint1

With fix:
mutluhint2

In Detail:
"superround" has a special case when going over the zero boundary, which I had wrong, since 'v' is always already positive in that case.

Instruction "MIAP" now respects the control value cutin, it was a TODO before, and it really ruined the hinting for this font not having been done. 

Kind regards, Axel


Jolg42:
@axkibe Thanks for the PR!

PR #319, opened 2018-02-10
Enable uglify's mangle and compress optimizations for a ~30% smaller …

lojjic:
## Description

The existing uglify minification did not enable its 'mangle' or 'compress' optimizations. Doing so results in a ~30% smaller opentype.min.js, mostly by shortening local variable names, and is theoretically a safe transformation.

Size before: 202780 bytes
Size after: 143972 bytes

I also removed the extra older `uglifyjs` dependency from package.json, as it did not appear to be used.

## How Has This Been Tested?

The existing mocha test for the minified file still passes. I also tested the index.html example with the newly minified file and verified it works as before.

## Types of changes

- [x] Bug fix (improvement, really) (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "watch": "rollup -c -w",
     "test": "mocha --require reify --compilers js:buble/register --recursive && jshint . && jscs .",
     "build": "rollup -c",
-    "minify": "uglifyjs ./dist/opentype.js > ./dist/opentype.min.js",
+    "minify": "uglifyjs  --compress --mangle -- ./dist/opentype.js > ./dist/opentype.min.js",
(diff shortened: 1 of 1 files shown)


fpirsch:
:+1:
This could also be the occasion to upgrade to the latest uglify-js 3.3.10

lojjic:
@fpirsch Good thought ... though technically it's already "upgraded" to 3.3.10 because it's using a caret range of `^3.0.14` which accepts all but major version bumps. I could go ahead and bump it up in package.json if you like, but it shouldn't make a functional difference for clean npm installs.

Jolg42:
I added the sourcemap and dist files in latest commit 😉

PR #323, opened 2018-03-01
Rewrite GPOS parsing

### LINKED ISSUE #292: ligature seems to not work correctly
Using Zilla-Regular (otf version, found here: https://github.com/mozilla/zilla-slab/tree/master/OTF_release%20files) on a page with opentypejs shows the font "doing the right thing" with the double slash in `Moz://a` as a text element, but when obtained through opentype.js, shows incorrect ligature of the `//` sequence (there should be a ligature, but there isn't).

Screenshot of a a page with a `<p>Moz://a</p>` and a canvas with the same string rendered using OpenType.js:

image

Code used:
```
      opentype.load(ZILLA_FONT_REGULAR, function(err, font) {
        if (err) {
          return console.log(err);
        } else {
          var ctx = document.getElementById('canvas').getContext('2d');
          var path = font.getPath('Moz://a', 0, 150, 72, {
            features: true,
            kern: true
          });
          path.draw(ctx);
        }
      });
```

(ZILLA_FONT_REGULAR is a data-uri for the font linked to above)

fpirsch:
## Description
Full rewrite of the old GPOS table parser.
 * exposes raw data in font.tables.gpos
 * code is simpler and consistent with the GSUB parser, with less duplication
 * better kerning

Fixes #292 and maybe other reported issues.

## Known limitations (what this doesn't fix)
 * only lookups 1 (single glyph positioning) and 2 (pair positioning) are parsed
 * kerning rendering still limited to xAdvance
 * extended tables (lookup 9, uses e.g. in Times New Roman) still not supported
 * no writing

## How Has This Been Tested?
New unit tests added. Tried the home page with at least a dozen various fonts.

## Types of changes

## Checklist:

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -4,6 +4,7 @@ import Path from './path';
 import sfnt from './tables/sfnt';
 import { DefaultEncoding } from './encoding';
 import glyphset from './glyphset';
+import Position from './position';
 import Substitution from './substitution';
 import { isBrowser, checkArgument, arrayBufferToNodeBuffer } from './util';
 import HintingTrueType from './hintingtt';
@@ -87,6 +88,7 @@ function Font(options) {
     this.supported = true; // Deprecated: parseBuffer will throw an error if font is not supported.
     this.glyphs = new glyphset.GlyphSet(this, options.glyphs || []);
     this.encoding = new DefaultEncoding(this);
+    this.position = new Position(this);
     this.substitution = new Substitution(this);
     this.tables = this.tables || {};
 
@@ -234,9 +236,7 @@ Font.prototype.glyphIndexToName = function(gid) {
 Font.prototype.getKerningValue = function(leftGlyph, rightGlyph) {
     leftGlyph = leftGlyph.index || leftGlyph;
     rightGlyph = rightGlyph.index || rightGlyph;
-    const gposKerning = this.getGposKerningValue;
-    return gposKerning ? gposKerning(leftGlyph, rightGlyph) :
-        (this.kerningPairs[leftGlyph + ',' + rightGlyph] || 0);
+    return this.kerningPairs[leftGlyph + ',' + rightGlyph] || 0;
 };
 
 /**
@@ -275,6 +275,11 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback) 
     options = options || this.defaultRenderOptions;
     const fontScale = 1 / this.unitsPerEm * fontSize;
     const glyphs = this.stringToGlyphs(text, options);
+    var kerningLookups;
+    if (options.kerning) {
+        const script = options.script || this.position.getDefaultScriptName();
+        kerningLookups = this.position.getKerningTables(script, options.language);
+    }
     for (let i = 0; i < glyphs.length; i += 1) {
         const glyph = glyphs[i];
         callback.call(this, glyph, x, y, fontSize, options);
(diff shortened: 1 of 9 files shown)


brawer:
Shameless plug: If you need test cases, have a look at this [test suite](https://github.com/unicode-org/text-rendering-tests).

Jolg42:
@fpirsch Also did you see @ghenania's comment https://github.com/nodebox/opentype.js/issues/292#issuecomment-369638965 about `So maybe you should parse subtables backward in Position.prototype.getKerningValue`
What do you think?

fdb:
@fpirsch sorry I don't have the bandwidth right now to go through this. Style-wise, it looks great; but I don't have the time right now to go deeper in the implementation stuff.

fpirsch:
@Jolg42 No, tables were parsed backwards and that was the problem. Now the function exits as soon as the first kerning pair is found, and this fixes #292 .

Jolg42:
@fpirsch There were a few JSCS errors, I just fixed that.
Next time just try to run `npm run dist` and check if all tests + JSCS are green ;-)

ghenania:
@fpirsch My mistake! I parsed your code too quickly.
`Position.prototype.getKerningValue` works right as it is for a pair of glyphs. But just a remark about: 
`Now the function exits as soon as the first kerning pair is found`

The true reason why it fixes #292 is that the kerning pair of the subtable with the lowest index always wins (even it is not really documented anywhere). We should keep in mind this point.

Pomax:
on a spec note, the traversal behaviour is documented in the OT spec in the chapter on the Common Table Layout, specifically in the section on [features and lookups](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#features-and-lookups):

"To implement features, a client applies the lookups *in the order the lookup definitions occur in the LookupList*. As a result, within the GSUB or GPOS table, lookups from several different features may be interleaved during text processing. A lookup is finished when the client locates a target glyph or glyph context and performs a substitution (if specified) or a positioning (if specified)."

(italics added here for empahsis)

ghenania:
@Pomax Thanks for the quote. I saw it too. Here the spec defines how we should parse multiple lookups. But it says nothing about on how we handle multiple subtables inside the same lookup when they cover the same feature for the same glyphs (e.g GPOS kern with one subtables in format 1 and another in formtat 2).

brawer:
> which font has multiple GPOS subtables for the same lookup for the same glyph?

Any font whose feature file uses an [ignore pos statement](https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#6.h.vi). For a test font, see [test case GPOS-2](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#GPOS-2).

Jolg42:
@Pomax Do you have a link for the `opentype mailing list`?

ghenania:
Thanks @brawer! Very interesting resources you shared here!

@Pomax good point, joining the opentype mailing list should help. Ho can we? 
To answer your question:
>which font has multiple GPOS subtables for the same lookup for the same glyph?

=> any font that needs to introduce few single kerning pairs exceptions for glyphs involved in kerning classes. 
The sample [font](https://rawgit.com/unicode-org/text-rendering-tests/master/fonts/TestGPOSOne.ttf) proposed by @brawer is a good example.

Pomax:
@Jolg42 you bet, https://blog.typekit.com/2008/01/29/opentype_15_draft/ has the various addresses at the top of the post, and all the people you'd hope read it, read it (all the ms, apple, adobe, etc. folks as well as most of the de facto font engine maintainers), and is a great place to ask for clarification or ask about modifying text in the spec itself.

Pomax:
Two other good places to ask (but less likely to be seen by the people who actually write the spec) is https://typedrawers.com and https://typophile.com (the latter used to be _the_ place for anything typography related but it kind of went offline for a few years and everyone kind of moved to typedrawers. It's back up, but I'm not sure how many people still pay attention to it)

Jolg42:
Thanks @Pomax I subscribed to the mailing list and waiting for manual approval!

fpirsch:
@Pomax 

> which font has multiple GPOS subtables for the same lookup for the same glyph?

For example Zilla slab, for the pair '//' (346-346), and that was the reason of the initial bug you filed ;-)
https://github.com/nodebox/opentype.js/issues/292#issuecomment-369316856

PR #331, opened 2018-03-13
Support UTF-16 surrogate pairs

antonytse:
## Description
Changed the code of getting the code point of a character in `stringToGlyphs` and `charToGlyph` for supporting surrogate pairs. JavaScript strings are encoded in UCS-2 and using the concept of surrogate pairs in UTF-16 for `U+10000` to `U+10FFFF`.

## Motivation and Context

It addresses #168 and #327, and the issue mentioned by @schuylr in #297.

## How Has This Been Tested?

1. `npm run test`
2. `npm run dist` and `npm run start` to check the usages of `opentype.js` via `http://localhost:8080`
3. Run the rendering test from https://github.com/unicode-org/text-rendering-tests. After the fix, it passed `CFF-1` and `CFF-2` test cases. Results in `GVAR-4`, `GVAR-5` and `GVAR-6`

 are still incorrect, but the icon is shown instead of 2 dots. See the screenshots below.

## Screenshots (if appropriate):
CFF-1 and CFF-2
GVAR-4, GVAR-5 and GVAR-6

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I did `npm run test` and all tests passed green (including code styling checks).
- [ ] I have added tests to cover my changes.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the **README** accordingly.
- [x] I have read the **CONTRIBUTING** document.

Issue #327, opened 2018-03-05
Cannot render codepoints above U+FFFF in CFF


brawer:
When rendering CFF fonts, OpenType.js currently fails to handle codepoints about U+FFFF.  See test cases [CFF-1](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#CFF-1) and [CFF-2](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#CFF-2). To reproduce, run the [test suite](https://github.com/unicode-org/text-rendering-tests).

antonytse:
I have just tried to fix this in #331, which should pass the test cases CFF-1 and CFF-2.

fpirsch:
Closed through #338

PR #338, opened 2018-03-16
Support non-Basic-Multilingual-Plane (BMP) characters

antonytse:
## Description
A new pull request after changing the code per the comments in #331.

Replaced `charCodeAt()` to `codePointAt()`  in `encoding.js` and modified `font.js` to handle surrogate pairs, for supporting non-Basic-Multilingual-Plane characters (i.e. `U+10000` to `U+10FFFF`).

## Motivation and Context

It addresses #168 and #327, and the issue mentioned by @schuylr in #297.

## How Has This Been Tested?

1. `npm run test`
2. `npm run dist` and `npm run start` to check the usages of `opentype.js` via `http://localhost:8080`
3. Run the rendering test from https://github.com/unicode-org/text-rendering-tests. After the fix, it passed `CFF-1` and `CFF-2` test cases. Results in `GVAR-4`, `GVAR-5` and `GVAR-6`
4. Tested in Internet Explorer 11 for the polyfill of `String.prototype.codePointAt`

 are still incorrect, but the icon is shown instead of 2 dots. See the screenshots below.

CFF-1 and CFF-2
GVAR-4, GVAR-5 and GVAR-6

## Types of changes

## Checklist:

### DIFF
--- a/bin/test-render
+++ b/bin/test-render
@@ -61,9 +61,9 @@ function renderSVG() {
 
     var glyphSet = new Set();
     let x = 0;
-    for (let i = 0; i < textToRender.length; i++) {
-        const c = textToRender[i];
-        const glyph = font.charToGlyph(c);
+    const glyphs = font.stringToGlyphs(textToRender);
+    for (let i = 0; i < glyphs.length; i++) {
+        const glyph = glyphs[i];
         const symbolId = testcase + '.' + glyph.name;
         if (!glyphSet.has(glyph)) {
             glyphSet.add(glyph);
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "fs": false
   },
   "dependencies": {
+    "string.prototype.codepointat": "^0.2.0",
     "tiny-inflate": "^1.0.2"
   }
 }
--- a/src/encoding.js
+++ b/src/encoding.js
@@ -133,7 +133,7 @@ function DefaultEncoding(font) {
 }
 
 DefaultEncoding.prototype.charToGlyphIndex = function(c) {
-    const code = c.charCodeAt(0);
+    const code = c.codePointAt(0);
     const glyphs = this.font.glyphs;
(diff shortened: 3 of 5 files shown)


fpirsch:
In `stringToGlyphs`, there is a trick to iterate simply over characters without checking surrogate pairs :
```
const chars = [...s];
const indexes = [];
for (let i = 0; i < chars.length; i += 1) {
        const c = s[i];
        indexes.push(this.charToGlyphIndex(c));
}
```

antonytse:
Thank you @fpirsch. I have tried to change the code as above but it seems to be broken after `rollup`, as `rollup-plugin-buble` converted the code as follow:
```
var this$1 = this;

var chars = [].concat( s );
var indexes = [];
for (var i = 0; i < chars.length; i += 1) {
    var c = chars[i];
    indexes.push(this$1.charToGlyphIndex(c));
}
```

`s` becomes the only element in the `chars` array and therefore `chars.length` always equal to `1`...

fpirsch:
Thanks @antonytse for trying :-)
I should have thought about bublé. If new syntax doesn't transpile, we can use the new method with good old syntax `Array.from` and "ponyfill" it (just discovered the word).

```
// Array.from implementation for strings thanks to Mathias Bynens ( https://github.com/mathiasbynens/Array.from )
var arrayFromString = Array.from || (s => s.match(/[\uD800-\uDBFF][\uDC00-\uDFFF]?|[^\uD800-\uDFFF]|./g) || []);

arrayFromString('a💩b');       // ['a','💩','b']
```

antonytse:
Thanks. I have added a commit for this. 😃

fpirsch:
The `arrayFromString` will be created over and over again if it is declared inside the `stringToGlyphs`method. It should be placed in the top-level scope, after the imports.
Sorry @antonytse, it should be perfect after that :-)

PR #339, opened 2018-03-20
Prevent auto-generated postScriptName from containing whitespace

mqudsi:
Whitespace characters are illegal in postScriptName and will result in
the font file being flagged as corrupt/invalid (at least on Windows 10).
Since all other fields are copied as-is from the options initialization
parameter without additional validation, I'm only stripping whitespace
in the case that opentype.js generates the postScriptName field
automagically from two other fields that _don't_ have the same
no-whitespace restriction.

### DIFF
--- a/src/font.js
+++ b/src/font.js
@@ -65,7 +65,8 @@ function Font(options) {
             fontFamily: {en: options.familyName || ' '},
             fontSubfamily: {en: options.styleName || ' '},
             fullName: {en: options.fullName || options.familyName + ' ' + options.styleName},
-            postScriptName: {en: options.postScriptName || options.familyName + options.styleName},
+            // postScriptName may not contain any whitespace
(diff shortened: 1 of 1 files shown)


fpirsch:
Good ! But it's not only a question of whitespaces :
> 
"the name string must be no longer than 63 characters and restricted to the printable ASCII subset, codes 33 to 126, except for the 10 characters '[', ']', '(', ')', '{', '}', '<', '>', '/', '%'"
(https://docs.microsoft.com/en-us/typography/opentype/spec/name)
So something like `.replace(/[^!-~[\](){}<>/%]/g, '')` would be more accurate.

mqudsi:
Oh, I was planning on opening another issue for the general lack of validation in opentype.js for SFNS tables.

fpirsch:
Oh :-/
Validation should be improved...

Jolg42:
It's better than nothing 😄 but indeed like @fpirsch said we should improve the regex for the postScriptName & add more validations.

PR #361, opened 2018-10-22
Render arabic rtl text properly

solomancode:
Refactor **featureQuery.js** implementation to favor functions composition over objects orientation 

- #### Gains:
  - _easy to test_
  - _easy to debug_
  - _fails early_
  - _loosely coupled_
---
Unify **Substitution** action + (_class **SubstitutionAction**_)

- #### Gains:
  - _common convention for glyph substitution_
  - _promotes code re-usability_
---
+ added **Latin word** context check (_required for applying Latin features_)
+ **normalize** features application for different scripts 
+ support **Latin ligatures feature** using featureQuery
+ auto apply **essential features** for scripts (arab & latn)
+ roll-back to default features if script was not found

## Types of changes

## Checklist:


src/font.js
```diff
@@ -274,10 +267,14 @@ Font.prototype.getKerningValue = function(leftGlyph, rightGlyph) {
...
-        liga: true,
-        rlig: true
-    }
+    features: [
+        /**
+         * these 4 features are required to render Arabic text properly
+         * and shouldn't be turned off when rendering arabic text.
+         */
+        { script: 'arab', tags: ['init', 'medi', 'fina', 'rlig'] },
+        { script: 'latn', tags: ['liga', 'rlig'] }
```
Jolg42: So if the font is not detected as Latin `liga` & `rlig` will be off?
Like if a font only supports emoji or a certain script?

src/font.js
```diff
@@ -143,6 +142,24 @@ Font.prototype.charToGlyph = function(c) {
     return glyph;
 };
 
+/**
+ * Update features
+ * @param {any} options features options
+ */
+Font.prototype.updateFeatures = function (options) {
+    // TODO: update all features options not only 'latn'.
```
Jolg42: Do you plan to finish this TODO before merging this?

Jolg42:
👍 for all the gains!

Just curious about what happens if a font script is not arab or latn?
Kerning will be on by default and all other features off?
If so it looks like a breaking change from the latest defaults.

src/font.js
solomancode: **Yes**, every registered feature is checked before application. If the **script is missing** or the font **doesn't support this feature**, It'll silently fail to apply this feature.

src/font.js
solomancode: Maybe later. But it'll not affect the current functionality at all. I need find a common model for updating features on UI controls changes / passed options to stringToGlyphs for all scripts.

solomancode:
Just tested it now, Kerning is working fine for Latin script. as for other features. I'm not sure which features should be applied by default for other scripts. for now I added the required features for Arabic **(init, medi, fina, rlig)** and the most important features for Latin **(liga, rlig)**. as for other scripts we might need some help from other contributors to understand which features to apply by default for which script.

solomancode:
Take your time to make sure everything is fine before merging to master, If anything needs clarifying just leave me a comment.

Issue #385, opened 2019-06-16
Merge default rendering options when custom options are given


davidje13:
## Expected Behavior

The following code should return the advance width for a string with hinting enabled:

```javascript
myFont.getAdvanceWidth('foo', 12, {hinting: true});
```

## Current Behavior

An error is thrown:

> ```
> TypeError: Cannot read property 'liga' of undefined
>    at /.../node_modules/opentype.js/src/font.js:159:57
>    at /.../node_modules/opentype.js/src/font.js:159:36
>    at Font.updateFeatures (/.../node_modules/opentype.js/src/font.js:155:47)
>    at Font.stringToGlyphs (/.../node_modules/opentype.js/src/font.js:186:10)
>    at Font.forEachGlyph (/.../node_modules/opentype.js/src/font.js:300:25)
>    at Font.getAdvanceWidth (/.../node_modules/opentype.js/src/font.js:384:17)
> ```

This was introduced as a problem in version 1.0.0, but only because nested objects were used in options for the first time; before that it still did not merge the defaults, so any omitted property was implicitly `false`.

## Possible Solution

Whenever user options are given, they should be merged with the default options. This will avoid the need for users to add every new property as they are introduced. This should be done for any API method which takes optional options. Note that this is a potentially breaking change for users who rely on the implicit `false`-ness of unspecified properties.

## Context

Updating from 0.11.0 to 1.1.0 I had to bisect this error until I found [this readme update](https://github.com/opentypejs/opentype.js/commit/0ba67d6f7934e426569e4fa7d0c0b62c6483bc0e) which offered enough of a clue. Also it was not obvious before that I was implicitly setting `kerning` to `false` when setting `hinting` to `true`.

## Your Environment
* Version used: 0.12.0 -> 1.0.0 transition
* Font used: any
* Browser Name and version: none (node.js)
* Operating System and version (desktop or mobile): any
* Link to your project: https://github.com/davidje13/SequenceDiagram

solomancode:
Hi @davidje13 ,

Fixed, check PR #386

davidje13:
@solomancode thanks for the quick response, though there is an issue with the PR;

This line:

```javascript
options = options ? Object.assign(options, this.defaultRenderOptions) : this.defaultRenderOptions;
```

looks backwards to me; it will overwrite any given options with the default options, rather than the reverse (it will also mutate the passed in options object, though that usually won't matter). Really it should be:

```javascript
[...]

PR #386, opened 2019-06-16
fix issue #385, merge default options with user options

solomancode:


src/font.js
```diff
@@ -295,7 +295,7 @@ Font.prototype.forEachGlyph = function(text, x, y, fontSize, options, callback)
     x = x !== undefined ? x : 0;
     y = y !== undefined ? y : 0;
     fontSize = fontSize !== undefined ? fontSize : 72;
-    options = options || this.defaultRenderOptions;
+    options = options ? Object.assign(options, this.defaultRenderOptions) : this.defaultRenderOptions;
```
davidje13: This line looks backwards to me; it will overwrite any given options with the default options, rather than the reverse (it will also mutate the passed in options object, though that usually won't matter). Really it should be:

```
options = options ? Object.assign({}, this.defaultRenderOptions, options) : this.defaultRenderOptions;
```

Which can be even simpler, since Object.assign can handle nulls/undefineds:

```suggestion
    options = Object.assign({}, this.defaultRenderOptions, options);
```

solomancode:
@davidje13 definitely, You've got a point. maybe I need a break.

jazlalli:
Has this fix found it's way into a release yet? Looking at the tags, the last release was 1st May, and this fix was subsequent to that.

How best to get the version that contains this fix?

gingerchris:
@Jolg42 hi, do you have any plans to create a new opentype.js release/tag containing this PR?  Is there any way we could help?  Thanks

PR #438, opened 2020-06-17
Unify internal Glyph.unicode{,s} usage

yne:
## Description
This commit:
- centralize this data into a the `unicodes` attribute.
- keep the Glyph API for specifying a single `unicode` value.
- update the documentation with both properties examples
- update the tests

## Motivation and Context
Both Glyph.unicode and Glyph.unicodes were used to store the glyph
unicodes values. This duplication led to some bugs when handling the
glyph unicodes.

## How Has This Been Tested?
I re-run the testcases and updates some that where kind of useless lick checking the first unicode of a glyph that match 4000 unicodes (`./node_modules/.bin/mocha --require reify --recursive --bail -g CID`)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ aPath.lineTo(100, 700);
 // more drawing instructions...
 const aGlyph = new opentype.Glyph({
     name: 'A',
-    unicode: 65,
+    unicodes: [65],
     advanceWidth: 650,
     path: aPath
 });
--- a/examples/creating-fonts.html
+++ b/examples/creating-fonts.html
@@ -101,7 +101,7 @@
     aPath.lineTo(200, 600);
     var aGlyph = new opentype.Glyph({
         name: 'A',
-        unicode: 65,
+        unicodes: [65],
         advanceWidth: 650,
         path: aPath
     });
@@ -123,7 +123,7 @@
     bPath.lineTo(200, 300);
     var bGlyph = new opentype.Glyph({
         name: 'B',
-        unicode: 66,
+        unicodes: [66],
         advanceWidth: 650,
         path: bPath
     });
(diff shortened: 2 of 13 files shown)


Jolg42 (approved):
This is really good 😃 

I will need to publish a new version and merge this after as it is a breaking change for users.

yne:
> This is really good 

Thanks :)

> I will need to publish a new version and merge this after as it is a breaking change for users.

I kept the `unicode` option support, to avoid too much breaking changes: https://github.com/opentypejs/opentype.js/pull/438/files#diff-61594601e6ba9344315027f64b17bf28561c1e08d312acc3288970248626a4b2

hope it's okay

Connum:
> I kept the `unicode` option support, to avoid too much breaking changes: https://github.com/opentypejs/opentype.js/pull/438/files#diff-61594601e6ba9344315027f64b17bf28561c1e08d312acc3288970248626a4b2

Just a thought... Should we mark this as deprecated via `console.warn()`, so it can be removed with an upcoming major version?

yne:
> > I kept the `unicode` option support, to avoid too much breaking changes: https://github.com/opentypejs/opentype.js/pull/438/files#diff-61594601e6ba9344315027f64b17bf28561c1e08d312acc3288970248626a4b2
> 
> Just a thought... Should we mark this as deprecated via `console.warn()`, so it can be removed with a future major version?

Sounds good, maybe @Jolg42 con confirm and I'll add a separate commit for that

yne:
I ran into some ligature regression with this version.
I'll understand + fix + add more test + un-WIP the PR when it's done

Issue #440, opened 2020-06-24
advanceWidth of Glyph is NaN after call to Glyph.getBoundingBox 


superelchi:
After writing a font to disk and reading it back advanceWidth for all Glyphs is set to NaN after calling Glyph.getBoundingBox. 

## Expected Behavior
Glyph.advanceWidth should return a number.

## Current Behavior
Glyph.advanceWidth returns NaN for all Glyphs.

## Steps to Reproduce (for bugs)

```
const fs = require('fs');
const opentype = require('opentype.js');
const assert = require('assert');

// Load font here: 
// https://google-webfonts-helper.herokuapp.com/api/fonts/open-sans?download=zip&subsets=latin&variants=regular
let font = opentype.loadSync('./open-sans-v17-latin-regular.woff');
const advanceWidthOne = font.charToGlyph('1').advanceWidth;

assert.strictEqual(font.charToGlyph('1').advanceWidth, advanceWidthOne, "advanceWidth before getBoundingBox");
font.charToGlyph('1').getBoundingBox();
// works
assert.strictEqual(font.charToGlyph('1').advanceWidth, advanceWidthOne, "advanceWidth after getBoundingBox");

const fileName = 'test.woff';
fs.writeFileSync(fileName, Buffer.from(font.toArrayBuffer()));
let newFont = opentype.loadSync(fileName, {});

assert.strictEqual(newFont.charToGlyph('1').advanceWidth, advanceWidthOne, "advanceWidth before getBoundingBox");
newFont.charToGlyph('1').getBoundingBox();
// does not work 
assert.strictEqual(newFont.charToGlyph('1').advanceWidth, advanceWidthOne, "advanceWidth after getBoundingBox");

```
## Your Environment

* Version used: 1.3.3
* Font used: see example above. tested with a couple of .woff and .ttf fonts.
* Browser Name and version: Node v12.18.1
* Operating System and version (desktop or mobile): Linux Mint Cinnamon 18.3

mattlag:
I am also running into this issue.  Some fonts seem to be fine, while others return NaN for all advance widths.  Attached is an OTF file that has this issue.
[BellweatherSans-Regular.zip](https://github.com/opentypejs/opentype.js/files/5071581/BellweatherSans-Regular.zip)

For the attached OTF, I investigated with TTX and the width information seems to be correct there (Looking for Latin Capital A to have an Advance Width of 1407):
[...]

Issue #417, opened 2019-12-16
advanceWidth of Glyph is set to NaN after calling getBoundingBox


adriaanmeuris:
## Expected Behavior
When calling the `getBoundingBox` method for a glyph, the `advanceWidth` property does not change

## Current Behavior
For some fonts, the `advanceWidth` is set to `NaN`.

## Steps to Reproduce (for bugs)
Use this font: [Sugardrop.otf.zip](https://github.com/opentypejs/opentype.js/files/3969461/Sugardrop.otf.zip)

```js
// Require modules
const assert = require('assert');
const opentype = require('./opentype.min.js');

// Load the font
opentype.load('./Sugardrop.otf', function(err, font) {
    if (err) {
        console.error('Font could not be loaded: ' + err);
    } else {
    	// Get glyph for A
    	const A = font.glyphs.glyphs[36];

    	// Advance width is equal to 749
    	assert.equal(A.advanceWidth, 749);

    	// Get the bounding box
    	A.getBoundingBox();

    	// Advance width is not equal to 749 anymore but NaN
    	assert.equal(A.advanceWidth, 749);
    }
});
```

## Your Environment
* Version used: `1.1.0`
* Font used: Sugardrop (see link above)
* Browser Name and version: Node `10.17.0`
* Operating System and version (desktop or mobile): MacOS Catalina
* Link to your project: see snippet above

catmando:
Is there any work around for this?

Connum:
Sounds like this is related to #397 and might have been fixed already, I'm going to check...

ILOVEPIE:
@Connum has this been fixed?

Issue #397, opened 2019-07-23
NaN is given to glyph.advanceWidth in parseCFFCharstring


aihornmac:
[this line of code](https://github.com/opentypejs/opentype.js/blob/master/src/tables/cff.js#L899) override previous value (say `377` for a specific case) to `NaN`

## Expected Behavior

Skip if it is NaN and give previous value default value

## Current Behavior

Write `NaN` anyway

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Font used:
* Browser Name and version:
* Operating System and version (desktop or mobile):
* Link to your project:

aihornmac:
It seems to be lack of `0` glyph. I changed the `makeSubFont` function to this, and the problem disappears.

```ts
export function makeSubFont(font: opentype.Font, content: string) {
  const glyphs = font.stringToGlyphs(Array.from(new Set(content.split(''))).join(''))
  glyphs.unshift(font.glyphs.get(0))
  return new opentype.Font({
    familyName: font.names.fontFamily.en,
    styleName: font.names.fontSubfamily.en,
    unitsPerEm: font.unitsPerEm,
    ascender: font.ascender,
    descender: font.descender,
    glyphs,
  })
}
```

PR #445, opened 2020-08-25
Initialize CFF DICT's _defaultWidthX and _nominalWidthX

ZeroUltimax:
As specified by [CFF specification](https://typekit.files.wordpress.com/2013/05/5176.cff.pdf) (see page 24), the defaultWidthX and nominalWidthX operators should have a default value of 0.

## Description
`gatherCFFTopDicts` parses the CFF table's DICT data. It is in charge of assigning correct defaults for the private DICT operator _subrs_. I assumed we would like to also initialize the default values for __defaultWidthX_ and __nominalWidthX_ here, since other calculations rely on there always being  a value for the topDict's __defaultWidthX_ and __nominalWidthX_.

## Motivation and Context
This solves the problem of OTF fonts that don't have a private DICT having all glyphs end up with a width of NaN, since the width calculation expects values for __defaultWidthX_ and __nominalWidthX_.

- https://github.com/opentypejs/opentype.js/issues/417
- https://github.com/opentypejs/opentype.js/issues/440
- https://github.com/opentypejs/opentype.js/issues/397

## How Has This Been Tested?

The following test used to fail their assertion. They now pass.

- Ran the test case from https://github.com/opentypejs/opentype.js/issues/417, it now works without failing the assert.
- Ran the test case from https://github.com/opentypejs/opentype.js/issues/440, it now works without failing the assert.
    -  Also ran the variant from mattlag's comment
- Ran the following variant of the test from https://github.com/opentypejs/opentype.js/issues/397, it now works failing the assert.

```
const fs = require('fs');
const opentype = require('./dist/opentype.js');
const assert = require('assert');

function makeSubFont(font, content) {
  return new opentype.Font({
    familyName: font.names.fontFamily.en,
    styleName: font.names.fontSubfamily.en,
    unitsPerEm: font.unitsPerEm,
    ascender: font.ascender,
    descender: font.descender,
    glyphs: font.stringToGlyphs(Array.from(new Set(content.split(''))).join('')),
  })
}

let font = opentype.loadSync('./SourceHanSans-Regular.otf');
const char = '上';
const advanceWidthChar = font.charToGlyph(char).advanceWidth;

assert.strictEqual(font.charToGlyph(char).advanceWidth, advanceWidthChar, "advanceWidth before getBoundingBox");
font.charToGlyph(char).getBoundingBox();
// works
assert.strictEqual(font.charToGlyph(char).advanceWidth, advanceWidthChar, "advanceWidth after getBoundingBox");

const subFont = makeSubFont(font, "上与乡互他伊作全公制前包单司客工布带开括旅晓朗机村棒游环球的相着离程管糖网罕联至蕾视足踏迹遍频颗高︐︒﹁﹂");

const fileName = 'test.woff';
fs.writeFileSync(fileName, Buffer.from(subFont.toArrayBuffer()));
let newFont = opentype.loadSync(fileName, {});

assert.strictEqual(newFont.charToGlyph(char).advanceWidth, advanceWidthChar, "advanceWidth before getBoundingBox");
newFont.charToGlyph(char).getBoundingBox();
// does not work
assert.strictEqual(newFont.charToGlyph(char).advanceWidth, advanceWidthChar, "advanceWidth after getBoundingBox");
```

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -384,6 +384,8 @@ function gatherCFFTopDicts(data, start, cffIndex, strings) {
         const topDict = parseCFFTopDict(topDictData, strings);
         topDict._subrs = [];
         topDict._subrsBias = 0;
+        topDict._defaultWidthX = 0;
+        topDict._nominalWidthX = 0;
         const privateSize = topDict.private[0];
         const privateOffset = topDict.private[1];
         if (privateSize !== 0 && privateOffset !== 0) {


Jolg42:
Amazing PR description ❤️ 

Thanks!

PR #447, opened 2020-10-03
Added the support of loading from http url in node environment.

ravishchofficial:
## Description
I've made an extra configuration change. I added support to pass isUrl (boolean) in opt parameter. If passed true it will allow user to load font from an http url in node environment as well.

## Motivation and Context
I myself was using the library and needed to load font from url in backend code written in node js. So now this will help user to use urls even in backend code. And previous behaviour is preserved.

## Types of changes

## Checklist:

### DIFF
--- a/src/opentype.js
+++ b/src/opentype.js
@@ -373,7 +373,7 @@ function parseBuffer(buffer, opt) {
  */
 function load(url, callback, opt) {
     const isNode = typeof window === 'undefined';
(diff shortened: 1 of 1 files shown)


Jolg42:
Thanks for your contribution!

It's not my favorite API but it's good enough for now I guess 👍

dontwork:
How can I use this functionality to load remote url on node?

Jolg42:
This should work
```
const font = await opentype.load(yourUrl);
```

PR #450, opened 2020-10-19
implement writing of coverage format 2

### LINKED ISSUE #223: Can't create coverage table format 2 yet.
Starting from 0.6.5 release, there is now an error when parsing OpenSans-Reulgar.ttf.
Can't create coverage table format 2 yet.

See the fiddle for an example. Version 0.6.4 works fine on the same font file.
https://jsfiddle.net/vd48vyzf/

Connum:
## Description

Implements writing of coverage tables in coverage format 2, which uses glyph ranges instead of a list of glyphs, see https://docs.microsoft.com/de-de/typography/opentype/spec/chapter2#coverage-format-2.

## Motivation and Context

Loading a font that contained a coverage table with coverage format 2 and trying to save it failed with the error message
> Can't create coverage table format 2 yet.

## How Has This Been Tested?

Loaded a font that contains at least one coverage table with coverage format 2 (e.g. this [test font](https://github.com/opentypejs/opentype.js/files/5403238/testfont_coverage_format_2.zip) i created) and save it (using `font.download()` or `font.toArrayBuffer()`). It will fail before the fix and work fine afterwards. I verified that the lookup table is still in the file and it opens fine using Windows Font viewer and FontForge.

Also, a new test case was created for this feature according to https://docs.microsoft.com/de-de/typography/opentype/spec/gsub#EX6 (and extended to three ranges, that's how I found out that I needed to make the changes to `Table()` as well). All other test cases still pass.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) - _this fixes #223, which was closed but never really fixed._

## Checklist:

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -12,9 +12,14 @@ import { encode, sizeOf } from './types';
  * @constructor
  */
 function Table(tableName, fields, options) {
-    for (let i = 0; i < fields.length; i += 1) {
-        const field = fields[i];
-        this[field.name] = field.value;
+    // For coverage tables with coverage format 2, we do not want to add the coverage data directly to the table object,
+    // as this will result in wrong encoding order of the coverage data on serialization to bytes.
+    // The fallback of using the field values directly when not present on the table is handled in types.encode.TABLE() already.
+    if (fields.length && (fields[0].name !== 'coverageFormat' || fields[0].value === 1)) {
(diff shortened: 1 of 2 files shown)


Jolg42 (approved):
Awesome! Thanks. Any example of a font to test this maybe?

PR #451, opened 2020-10-19
Handle undefined chararray

### LINKED ISSUE #240: noto load/save problem with glyphNames
Hello,

I am trying to work with opentype.js and the noto fonts.
The font I use in this example is "NotoSansTagbanwa-Regular.ttf" available in https://noto-website-2.storage.googleapis.com/pkgs/NotoSansTagbanwa-unhinted.zip but the problem also happens with the other noto ttf files.

Basically, the following example outputs an error

```
var opentype = require('opentype.js');
opentype.load("NotoSansTagbanwa-Regular.ttf", function(err, font) {
  if (err) { console.log(err) }
  else {
    font.toArrayBuffer()
  }
})

```

leads to

> /home/x/tmp/noto/node_modules/opentype.js/src/types.js:75
>    for (var i = 0; i < v.length; i += 1) {
>                         ^
> TypeError: Cannot read property 'length' of undefined
>     at encode.CHARARRAY (/home/x/tmp/noto/node_modules/opentype.js/src/types.js:75:26)
>     at Object.encode.OBJECT (/home/x/tmp/noto/node_modules/opentype.js/src/types.js:764:12)
>     at Object.encode.INDEX (/home/x/tmp/noto/node_modules/opentype.js/src/types.js:603:24)
>     at sizeOf.INDEX (/home/x/tmp/noto/node_modules/opentype.js/src/types.js:632:19)
>     at Object.sizeOf.TABLE (/home/x/tmp/noto/node_modules/opentype.js/src/types.js:840:21)
>     at Table.sizeOf (/home/x/tmp/noto/node_modules/opentype.js/src/table.js:50:19)
>     at Object.makeCFFTable [as make] (/home/x/tmp/noto/node_modules/opentype.js/src/tables/cff.js:1095:23)
>     at Object.fontToSfntTable [as fontToTable] (/home/x/tmp/noto/node_modules/opentype.js/src/tables/sfnt.js:293:24)
>     at Font.toTables (/home/x/tmp/noto/node_modules/opentype.js/src/font.js:379:17)
>     at Font.toArrayBuffer (/home/x/tmp/noto/node_modules/opentype.js/src/font.js:393:26)
> 

I drilled down on this issue, and it seems to happen because after loading the font, the glyph names are undefined.

The `font.glyphNames.names` array exists and is empty because `post.version = 3`

I observed the code path in encoding.js 
```
function addGlyphNames(font) {
    var glyph;
    var glyphIndexMap = font.tables.cmap.glyphIndexMap;
    var charCodes = Object.keys(glyphIndexMap);

    for (var i = 0; i < charCodes.length; i += 1) {
        var c = charCodes[i];
        var glyphIndex = glyphIndexMap[c];
        glyph = font.glyphs.get(glyphIndex);
        glyph.addUnicode(parseInt(c));
    }

    for (i = 0; i < font.glyphs.length; i += 1) {
        glyph = font.glyphs.get(i);
        if (font.cffEncoding) {
            glyph.name = font.cffEncoding.charset[i];
        } else if (font.glyphNames.names) {
            glyph.name = font.glyphNames.glyphIndexToName(i);
        }
    }
}

```

and we hit `glyph.name = font.glyphNames.glyphIndexToName(i);` which answers undefined because the array is empty.

then the `CHARARRAY` crashes because it does not know how to serialize undefined.

I would be happy to propose a PR for this but I don't know what would be the best spec-friendly fix. I don't know how to verify if the names exist in the original .ttf, if it is correct or not that a ttf could be created without glyph names, nor if it is acceptable to create a font with empty glyph names

for instance, if the code is modified with

```
        if (font.cffEncoding) {
            glyph.name = font.cffEncoding.charset[i];
        } else if (font.glyphNames.names && font.glyphNames.names.length) {
            glyph.name = font.glyphNames.glyphIndexToName(i);
        } else {
            glyph.name = '';
        }

```

then the load/save cycle works but the fix is maybe too spec-preemptive. There seem to be other problems with the load/save cycle of other noto fonts. When I add the patch above and use the test on `NotoSans-Regular.ttf` I avoid the same problem but then I hit

> /home/x/tmp/noto/node_modules/opentype.js/src/check.js:6
>     throw new Error(message);
>     ^
> Error: Unable to write GSUB lookup type 7 tables.
>     at Object.exports.fail (/home/x/tmp/noto/node_modules/opentype.js/src/check.js:6:11)
>     at Object.exports.argument (/home/x/tmp/noto/node_modules/opentype.js/src/check.js:13:17)
>     at /home/x/tmp/noto/node_modules/opentype.js/src/table.js:180:15
>     at tableList (/home/x/tmp/noto/node_modules/opentype.js/src/table.js:76:68)
>     at new LookupList (/home/x/tmp/noto/node_modules/opentype.js/src/table.js:178:41)
>     at Object.makeGsubTable [as make] (/home/x/tmp/noto/node_modules/opentype.js/src/tables/gsub.js:253:49)
>     at Object.fontToSfntTable [as fontToTable] (/home/x/tmp/noto/node_modules/opentype.js/src/tables/sfnt.js:312:26)
>     at Font.toTables (/home/x/tmp/noto/node_modules/opentype.js/src/font.js:379:17)
>     at Font.toArrayBuffer (/home/x/tmp/noto/node_modules/opentype.js/src/font.js:393:26)
> 

Thanks for your feedback !

Connum:
## Description

If `undefined` is passed as the argument for `encode.CHARSET` or `sizeOf.CHARSET`, treat it as an empty string.

## Motivation and Context

Some fonts contain (or are at least parsed with) glyphs whose name is `undefined`. Saving these fonts will fail with the following error:
> Uncaught TypeError: Cannot read property 'length' of undefined
>     at encode.CHARARRAY (types.js:71)
>     at Object.encode.OBJECT (types.js:871)
>     at Object.encode.INDEX (types.js:710)
>     at sizeOf.INDEX (types.js:739)
>     at Object.sizeOf.TABLE (types.js:946)
>     at Table.sizeOf (table.js:47)
>     at Object.makeCFFTable [as make] (cff.js:1301)
>     at Object.fontToSfntTable [as fontToTable] (sfnt.js:288)
>     at Font.toTables (font.js:483)
>     at Font.toArrayBuffer (font.js:497)
> 

I'll create this PR as a draft for discussion because I'm not sure whether this just fixes a symptom instead of the cause. I wasn't able to find out whether glyphs without a name are actually valid according to the OpenType spec.
In #240 it was proposed to derive the glyph name if it is not set: https://github.com/opentypejs/opentype.js/issues/240#issuecomment-260134716 - that sounds good to me. Should we still implement this fix so that in any case `undefined` as a CHARARRAY is treated as an empty string, maybe loggin an error to the console? That way, the writing process wouldn't simply fail and prevent the user from saving the font.

## How Has This Been Tested?

A font that was parsed with some glyph names being `undefined` could not be saved before the fix, and can be after. I also added a test case for the fix.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) - fixes #240 

## Checklist:

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -67,6 +67,9 @@ sizeOf.CHAR = constant(1);
  * @returns {Array}
  */
 encode.CHARARRAY = function(v) {
+    if (typeof v === 'undefined') {
+        v = '';
+    }
     const b = [];
     for (let i = 0; i < v.length; i += 1) {
         b[i] = v.charCodeAt(i);
--- a/src/types.js
+++ b/src/types.js
@@ -69,6 +69,7 @@ sizeOf.CHAR = constant(1);
 encode.CHARARRAY = function(v) {
     if (typeof v === 'undefined') {
         v = '';
+        console.warn('Undefined CHARARRAY encountered and treated as an empty string. This is probably caused by a missing glyph name.');
     }
     const b = [];
     for (let i = 0; i < v.length; i += 1) {
--- a/src/types.js
+++ b/src/types.js
@@ -83,6 +83,9 @@ encode.CHARARRAY = function(v) {
  * @returns {number}
  */
 sizeOf.CHARARRAY = function(v) {
+    if (typeof v === 'undefined') {
+        return 0;
+    }
     return v.length;
 };
 
--- a/test/types.js
+++ b/test/types.js
@@ -18,6 +18,11 @@ describe('types.js', function() {
         assert.equal(sizeOf.CHARARRAY('A/B'), 3);
     });
 
+    it('can handle undefined as CHARARRAY', function() {
+        assert.equal(hex(encode.CHARARRAY(undefined)), '');
+        assert.equal(sizeOf.CHARARRAY(undefined), 0);
+    });
+
     it('can handle USHORT', function() {
         assert.equal(hex(encode.USHORT(0xCAFE)), 'CA FE');
         assert.equal(sizeOf.USHORT(0xCAFE), 2);


Jolg42:
Hi @Connum 

I see the state of the PR is still in "Draft", is it accurate?

Jolg42:
HI @Connum 😃 

> In #240 it was proposed to derive the glyph name if it is not set: #240 (comment) - that sounds good to me. 

Indeed that's the best solution I think.

> Should we still implement this fix so that in any case undefined as a CHARARRAY is treated as an empty string, maybe loggin an error to the console? That way, the writing process wouldn't simply fail and prevent the user from saving the font.

Indeed it sounds better than failing and it could be interesting to have the logging.

Connum:
Thanks for your feedback! I marked the PR as ready for review, so this fix can be implemented, and the solution proposed in #240 can then be implemented in another PR, if desired. I also added the console warning.

PR #452, opened 2020-11-07
Add UI names for Stylistic Set / Character Variants features

### LINKED ISSUE #399: No mapping of stylistic set names
## Expected Behavior
For a font where OpenType stylistic sets have assigned human readable names, they are contained in the font object in `tables.names`, at indices starting from `256`. Stylistic sets are contained in `tables.gsub.features` where each entry has format `{ tag, feature { featureParams, lookupListIndexes} }`. I would expect a reference to the index of the corresponding entry from `tables.names` somewhere in there.

## Current Behavior
There is no information on which name record belongs to which stylistic set. Technically it is possible to guess that the stylistic set names are stored consecutively but this assumption fails if not all stylistic sets have assigned names, e.g. a font could contain names only for ss01, ss02, ss03, and, why not, ss20. Then the name for ss20 would be wrongly assigned to ss04, and for further sets other metadata like copyright notice, designer name, font names which are also stored in `tables.names` could be wrongly assigned if we don't take precautions.

## Possible Solution
Add index of the name record to the feature object.

## Context
I am building a GUI for manual testing of fonts at https://hyvyys.github.io/Bulletproof/

## My Environment
* Version used: 1.1.0
* Font used: custom
* Link to your project: https://hyvyys.github.io/Bulletproof/

hyvyys:
## Description

When parsing the features list, the `featureParams` offset value is being retrieved but the FeatureParams table itself is not. With these changes, the FeatureParams table will be extracted and the `uiNameId` from it will be matched against the `name` table, and the resulting `name` record will be inserted into the feature definition for features `ss01`–`ss20`.

## Motivation and Context

This is intended to close #399.

## How Has This Been Tested?

I tested the code with several fonts that had stylistic set names assigned as well as with ones that didn't.

I ran the automated tests – I had to adjust the parse.js test to include the `tableOffset` parameter needed to go back to the feature table and add missing data. This implementation is not ideal but I had no better idea due to the highly functional/declarative style of this project (also, my noob skills). Open to suggestions on a better implementation here.

## Types of changes

## Checklist:


dist/opentype.module.js
```diff
@@ -14148,6 +14172,13 @@ function parseBuffer(buffer, opt) {
     if (gsubTableEntry) {
         var gsubTable = uncompressTable(data, gsubTableEntry);
         font.tables.gsub = gsub.parse(gsubTable.data, gsubTable.offset);
+        font.tables.gsub.features.forEach(function (f) {
+            if (f.tag.match(/ss(?:0[1-9]|1\d|20)/)) {
```
Jolg42, line 14176:
```suggestion
			// Match `ss01` to `ss20`
            if (f.tag.match(/ss(?:0[1-9]|1\d|20)/)) {
```

src/tables/gsub.js
```diff
@@ -193,12 +193,20 @@ function parseGsubTable(data, start) {
...
     if (tableVersion === 1) {
-        return {
+        const table = {
             version: tableVersion,
             scripts: p.parseScriptList(),
             features: p.parseFeatureList(),
             lookups: p.parseLookupList(subtableParsers)
         };
+        table.features.forEach(f => {
+            if (f.tag.match(/ss(?:0[1-9]|1\d|20)/)) {
```
Jolg42, line 203:
```suggestion
			// Match `ss01` to `ss20`
            if (f.tag.match(/ss(?:0[1-9]|1\d|20)/)) {
```

Jolg42:
Nice!

I think it could use an example & test with a font that has Stylistic Set,

How did you test this?

vinuel:
One note: Not only Stylistic Sets (ss01–ss20) can have an optional UI Name ID but also the [Character Variant feature](https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-cv01--cv99) (cv01–cv99). Maybe it would be good to implement this as well, once you are working on it?  

The font [Source Code Variabe](https://github.com/adobe-fonts/source-code-pro/releases/tag/variable-fonts) is using both.

hyvyys:
@Jolg42 I tested this in development on the included example page (the one served with the `start` script) with a few simple `console.log`s and by inspecting the GSUB table print on the Font Inspector page – I dragged various fonts onto the file input. Some fonts may have the feature names defined for stylistic sets while others may not. Google fonts seem to have that data stripped while the same fonts elsewhere may retain this data (e.g. Poppins on FontSquirrel – the version from Google still has the `nameId`s in place but the `name` table is brutally trimmed; this causes no issue as far as we are concerned, the `uiName` field is just left undefined).

I've now added automated tests for the added features. As for an example, do you have anything specific in mind? An addition to the example page or the README?

@vinuel Good point, thanks. The FeatureParams table is now being parsed for `ccv01`–`ccv99` as well (with a separate function – [the table's format](https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-cv01--cv99) is dependent on the specific feature). Only the `featUiLabelNameId` field is being resolved against the `name` table though since `featUiLabelName` seems to be more commonly implemented. (Or do you think all fields should be pre-extracted? If someone needs them, they can simply address the `name` table with the `nameId` that is already in place.)

vinuel:
I just wonder, isn't this ready to get published?

ILOVEPIE:
@hyvyys please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

src/opentype.js
```diff
@@ -368,6 +368,18 @@ function parseBuffer(buffer, opt) {
     if (gsubTableEntry) {
         const gsubTable = uncompressTable(data, gsubTableEntry);
         font.tables.gsub = gsub.parse(gsubTable.data, gsubTable.offset);
+        font.tables.gsub.features.forEach(f => {
```
Connum, line 371: I stumbled across "uiName" vs. "UiLabelName" at first, but turns out that's in the spec as well, just in case anyone in the future wonders why this is not the same.

Connum (dismissed):
LGTM! 🚀

Connum:
@ILOVEPIE I only resolved merge conflicts in .js.map files that will be re-generated anyway, but therefore my review does not count towards allowing the merge because I'm technically the last pusher - do you want to review it as well, or shall I proceed with the merge?

ILOVEPIE:
> @ILOVEPIE I only resolved merge conflicts in .js.map files that will be re-generated anyway, but therefore my review does not count towards allowing the merge because I'm technically the last pusher - do you want to review it as well, or shall I proceed with the merge?

We always want to have someone else review first.

ILOVEPIE:
@Connum quick question, when we merge this, will it affect #542 at all? Just curious.

src/opentype.js
ILOVEPIE: this library is already pretty slow, I don't personally like using forEach due to the excess function calls compared with a for-in loop, would it be possible to switch this to a for-in loop?

Connum:
> @Connum quick question, when we merge this, will it affect #542 at all? Just curious.

No, these labels are stored with the features and independent from the platform-specific font names.

src/parse.js
```diff
@@ -241,6 +257,22 @@ Parser.prototype.parseShortList = function(count) {
     return list;
 };
 
+// Parse a list of 24 bit unsigned integers. The length of the list can be read on the stream
+// or provided as an argument.
+Parser.prototype.parseUInt24List = function(count) {
```
ILOVEPIE, line 262: This is probably better named ``parseUInt24Array`` as a List is usually used to refer to a Linked List, but that's a minor sticking point.

src/parse.js
```diff
@@ -391,11 +423,16 @@ Parser.prototype.parseValueRecordList = function() {
...
 };
 
-Parser.prototype.parsePointer = function(description) {
+Parser.prototype.parsePointer = function(description, storeOffset = false) {
     const structOffset = this.parseOffset16();
     if (structOffset > 0) {
         // NULL offset => return undefined
-        return new Parser(this.data, this.offset + structOffset).parseStruct(description);
+        const offset = this.offset + structOffset;
+        const struct = new Parser(this.data, offset).parseStruct(description);
```
ILOVEPIE, line 431: Why do we need to create a new ``Parser`` object here when we're already in one? A better solution might be to add a stack to the parser that we can push the current cursor location onto and jump to a new cursor location and then when we're done there we pop it off to jump back to the original cursor location. Would also reduce the amount of memory allocations used when parsing fonts, speeding up the font parsing.

src/tables/gsub.js
```diff
@@ -209,6 +210,21 @@ function parseGsubTable(data, start) {
         };
     }
 
+    table.features.forEach(f => {
```
ILOVEPIE, line 213: See the above comment about ``forEach``.

ILOVEPIE (commented):
Looks good apart from those comments I made, Idk if we want to make those changes to the Parser first as a separate PR and then include this one, or if we want to approve this PR and then make those changes with a later PR.

src/parse.js
ILOVEPIE: although it looks like there are other functions named this way so i don't think we should break the convention we already have.

ILOVEPIE:
@Connum could you look over the review I did. I'll probably do a separate PR to fix [this](https://github.com/opentypejs/opentype.js/pull/452#discussion_r1096584350). Any other issues, or is it just the forEach loops?

Connum:
Yeah, let's do the optimizations in a separate PR! Other than that, it looks good to go!

ILOVEPIE:
@Connum seeing as this needs a code update anyways to merge, can we change those ``forEach`` calls to for-in loops? I'll do the Parser optimization in a separate PR after this gets merged.

src/opentype.js
Connum: done!

src/parse.js
Connum: Absolutely, let's stick to the naming convention we already have.

ILOVEPIE:
@Connum did we run the tests?

Connum:
> @Connum did we run the tests?

Narf! No we didn't! 😉 I'll look into it!

Connum:
Ok, so the thing is that the "Preserve platform specific names in fonts" feature has broken this one. Now, all the feature labels for character variants and stylistic types are added to each of the platform names as a numeric index.
image

We could just take the first platform with a value for that key that we encounter, but I'm honestly not content with the memory overhead all these duplications are causing. I'm not really sure on the best way to handle this. Put the labels wit ha numeric key into a separate place? But where?

Connum:
So I just read the spec and feature names can indeed be platform-specific as well, so we can't get rid of the overhead that the platform preservation has caused:
https://learn.microsoft.com/en-us/typography/opentype/spec/name#naming-table-header
Sooo... feature.uiName and feature.featUiLabelName should not return an object like
```
{el: 'απλό a', en: 'simple a', ru: 'простой а'}
```

but rather
```
{
  unicode: {el: 'απλό a', en: 'simple a', ru: 'простой а'},
  windows: {el: 'απλό a', en: 'simple a', ru: 'простой а'},
  macintosh: {el: 'απλό a', en: 'simple a', ru: 'простой а'}
}
I guess?

Connum:
> that sounds good but I don't think those numeric indexes are supposed to be showing up in the names object.

Where should we store them instead?

Connum:
Numeric names indexes are by the way already sotored that way, so I don't think this should be fixed in this PR.

PR #461, opened 2020-12-15
Set correctly unicode 0 to ".null" glyph

puria:
as per https://docs.microsoft.com/en-us/typography/opentype/otspec170/recom#first-four-glyphs-in-fonts the unicode number fo the ".null" glyph should be 0 but was set to undefined

## Description

The default value of the unicode property in the glyph object was evaluated as https://github.com/opentypejs/opentype.js/blob/88f7c1755ceec3326cdac4de53bc5ec89d38a615/src/glyph.js#L64
So was set to undefined when the unicode property was set to 0.

Formally should be 

```js
    this.unicode = options.unicode !== undefined || undefined;
```
that we can short circuit to
```js
    this.unicode = options.unicode;
```

## Motivation and Context
No open issues.

## How Has This Been Tested?
Added new unit test

## Types of changes

## Checklist:


khaledhosny:
> as per https://docs.microsoft.com/en-us/typography/opentype/otspec170/recom#first-four-glyphs-in-fonts the unicode number fo the ".notdef" glyph should be 0

That is not what the link is saying. `.notdef` is given _Glyph ID_ 0 and _undefined_ Unicode value. Unicode 0 (i.e. `U+0000`)  is mapped to `.null` glyph.

Anyway, this is a recommendation for font authors not consumers.

puria:
> > as per https://docs.microsoft.com/en-us/typography/opentype/otspec170/recom#first-four-glyphs-in-fonts the unicode number fo the ".notdef" glyph should be 0
> 
> That is not what the link is saying. `.notdef` is given _Glyph ID_ 0 and _undefined_ Unicode value. Unicode 0 (i.e. `U+0000`) is mapped to `.null` glyph.
> 
> Anyway, this is a recommendation for font authors not consumers.

Cool ;) anyway this fixes the `.null` glyph then, cause is not possible to give a 0 value to the unicode attribute within the glyph object... should I update the tests?

I saw it frequently used, like: 

https://github.com/opentypejs/opentype.js/blob/master/test/opentypeSpec.js#L83-L87

Templarian:
Noticed this when looking at the docs, but this PR addresses the same issue.

Looks worth merging still.

Templarian (approved):

ILOVEPIE:
@puria please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

src/glyph.js
```diff
@@ -61,7 +61,7 @@ Glyph.prototype.bindConstructorValues = function(options) {
 
     // These three values cannot be deferred for memory optimization:
     this.name = options.name || null;
-    this.unicode = options.unicode || undefined;
+    this.unicode = options.unicode;
```
Connum, line 64: I would suggest to explicitly set the unicode value to `undefined` if the name is `.notdef`, so that any code copied from the README prior to this fix will still result in a correct glyph. And while we're at it, we could force 0 for the `.null` name as well and maybe even prevent any other glyph from getting assigned a unicode value of 0.

Connum (commented):
Just some minor suggestions, other than that we should get this merged soon, as it addresses something quite elementary when creating fonts.

src/glyph.js
Connum: @Templarian Do you have any objections?

Templarian: @Connum That should always be the case, so that would be a good idea.

Connum:
I added my suggested changes as well as tests for them.

Templarian (approved):
Change looks good as do the new tests.

ILOVEPIE (approved):
looks good to me.

PR #482, opened 2021-07-23
remove "adding CMAP format 12" log

zbryikt:
remove a log.

## Description
A log shown when adding CMAP format 12  is removed.

## Motivation and Context
This seems to be a debug message. It messes up with the program and we don't actually need this when using opentype.js 

## How Has This Been Tested?
not quite tested intensively except `npm run test` but since it only removed one line of log, it should be fine.

## Types of changes

## Checklist:


zbryikt:
the log is removed to prevent this

yne:
this PR contain too much unrelated changes.
there are other logs:
https://github.com/opentypejs/opentype.js/blob/ad8a56761e3bcc1a32e5137a0df56672ecfda74d/src/tables/cff.js

maybe we could use `console.error` instead.

zbryikt:
oops! sorry about that. But not sure if I have time to clean up that
so perhaps I will simply close it for now. :)

Issue #415, opened 2019-12-09
substFormat: 1 is not yet supported for Calibri Font


xornor-web:
## Expected Behavior
SubSet Font not creating for Calibri Font

## Current Behavior
Error: lookupType: 7 - substFormat: 1 is not yet supported
    at FeatureQuery.getLookupMethod (..\node_modules\opentype.js\dist\opentype.js:12239:20)
    at FeatureQuery.lookupFeature (..\node_modules\opentype.js\dist\opentype.js:12295:34)
    at ..\node_modules\opentype.js\dist\opentype.js:12720:43
    at Array.forEach (<anonymous>)
    at Bidi.latinLigature (..\node_modules\opentype.js\dist\opentype.js:12718:28)
    at C:\Users\Xornor\nodeJSpro\pptx2html\node_modules\opentype.js\dist\opentype.js:12911:24
    at Array.forEach (<anonymous>)
    at Bidi.applyLatinLigatures (..\node_modules\opentype.js\dist\opentype.js:12910:13)
    at Bidi.applyFeaturesToContexts (..\node_modules\opentype.js\dist\opentype.js:12932:30)
    at Bidi.processText (..\node_modules\opentype.js\dist\opentype.js:12947:15)

## Steps to Reproduce (for bugs)
1. I am creating calibri.ttf suset font from Node but error occurring 
```
let filePath = "calibri.ttf";
let desFilePath = "sub_calibri.ttf";
var font = opentype.loadSync(filePath);
var subsetFont = undefined;
var glyphs = font.stringToGlyphs("1234567890abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ");
subsetFont = new opentype.Font({
    familyName: font.names.fontFamily.en,
    styleName: font.names.fontSubfamily.en,
    unitsPerEm: font.unitsPerEm,
    ascender: font.ascender,
    descender: font.descender,
    glyphs: glyphs
});
fs.writeFileSync(desFilePath, Buffer.from(subsetFont.toArrayBuffer()));
var fontEnable = opentype.loadSync(desFilePath);
```

Connum:
I'm currently working on writing support for lookup type 6 (format 1 and 3 are done) and hope to find the time to implement all other missing lookup types as well.

devlfm:
I'm having this error an Arial too. I updated to last release 1.3.3 and still getting this error. How can I solve this?

devlfm:
[...]

PR #454, opened 2020-11-12
add tests for parsing lookup type 6 formats 1 and 3

Connum:
## Description

Based on the tests I added for writing lookup type 6 format 1 and 3 tables, I also added the missing tests for the parsing side of the functionality.

## Motivation and Context

Although parsing of those tables was supported for quite some time, there were never tests to cover the functionality, because the specs didn't provide examples.

## How Has This Been Tested?

Ran all tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

_(not applicable, tests only)_

## Checklist:

- [X] I did `npm run test` and all tests passed green (including code styling checks).
- [X] I have added tests to cover my changes.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the **README** accordingly.
- [X] I have read the **CONTRIBUTING** document.

PR #486, opened 2021-08-12
GSUB lookup table type 7 subtable format 1

st0nerhat:
## Description

_In featureQuery.js:chainingSubstitutionFormat3_
- Modify var types of subtable, lookup, and substitutionType to be lets so they can be mutated in the next step.
- Detect substitution type "71" which means lookup table type 7 and format 1. 
- When detected, mutate the substitution type and lookup method based on the linked extension subtable
- Finally update the subtable variable to be the extension table
- Throw an error if the chained substitution type is not handled. This will avoid fonts that use chaining substitutions from silently rendering incorrectly if they use an unhandled changed substitution.

_In featureQuery.js:FeatureQuery.prototype.lookupFeature_

Same changes as above.

## Motivation and Context

Addresses issue #415 

GSUB tables have a lookup table type 7 called the Extension Substitution Subtable. Only one format officially exists (format 1). 
See the [OpenType spec page](https://docs.microsoft.com/en-us/typography/opentype/spec/gsub#lookuptype-7-extension-substitution).

Per the spec, Extension tables are used in fonts that use a large number of substitution subtables. Fonts for scripts where letters are connected are a situation where many substitutions are needed.

## How Has This Been Tested?

So far I have tested this on 3 custom fonts in my library, one of which contains type 7 lookup tables and 2 that do not. Unfortunately I do not have the license to upload these fonts in the PR.

## Types of changes

## Checklist:

### DIFF
--- a/src/features/featureQuery.js
+++ b/src/features/featureQuery.js
@@ -150,7 +150,7 @@ function chainingSubstitutionFormat3(contextParams, subtable) {
             const lookupTable = this.getLookupByIndex(lookupListIndex);
             for (let s = 0; s < lookupTable.subtables.length; s++) {
                 let subtable = lookupTable.subtables[s];
-                let lookup = this.getLookupMethod(lookupTable, subtable);
+                let lookup;
                 let substitutionType = this.getSubstitutionType(lookupTable, subtable);
 
                 if (substitutionType === '71') {
@@ -158,6 +158,8 @@ function chainingSubstitutionFormat3(contextParams, subtable) {
                     substitutionType = this.getSubstitutionType(subtable, subtable.extension);
                     lookup = this.getLookupMethod(subtable, subtable.extension);
                     subtable = subtable.extension;
+                } else {
+                    lookup = this.getLookupMethod(lookupTable, subtable);
                 }
 
                 if (substitutionType === '12') {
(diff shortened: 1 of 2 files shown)


Connum:
> Unfortunately I do not have the license to upload these fonts in the PR.
The tests don't use actual font files to run (see https://github.com/opentypejs/opentype.js/pull/454/commits/3177c13446f79542cc11b5958b5083f6cbce5e48 for example), so it would be great if you could add (a) test case(s) to your contribution!

ILOVEPIE:
@st0nerhat please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

Connum (changes requested):
requires test(s)

Connum:
I just added a PR FilamentGames#2 to the feature branch that adds a test for this parsing feature. @st0nerhat, could you please merge that so we can get this PR merged?

Connum (approved):

Connum:
As I provided the test, a different reviewer should make a review as well, so I tagged @ILOVEPIE

ILOVEPIE (approved):
Looks good to me, apart from the fact that we're still uncertain on the tests/no examples in the microsoft documentation, perhaps we should check how libfreetype handles it?

Connum:
> Looks good to me, apart from the fact that we're still uncertain on the tests/no examples in the microsoft documentation, perhaps we should check how libfreetype handles it?

The functionality is relatively straight-forward. I tried looking at libfreetype, but I didn't know where to look exactly. In my opinion, the test that we have now should be sufficient.

Connum:
I haven't come across a better stripped down example for a test, but I didn't come across any problems either so far, so I'm merging this now.

PR #490, opened 2021-09-19
Add COLRv0 + CPALv0 support

yne:
## Description
This commit add support for COLR and CPAL table for colored glyph support

## Motivation and Context
Every browser now support COLR font. So we can expect traction for this feature anytime soon.

OpenJS can now generate and parse COLRv0 based otf

## How Has This Been Tested?
- I created a custom font with my opentpye.js : [test.otf.gz](https://github.com/opentypejs/opentype.js/files/7192855/test.otf.gz)
- tested in chrome + gnome font viewer => both works
- I added two unit test for building and parsing each table => all pass

image

## Types of changes

## Checklist:


yne:
The `make`/`parse` process is consistent , But I still struggle to understand the relation between `firstLayerIndex` and `numLayers`.

For now, I can only make the 1 colors-per-glyph work if I create a huge palette, than shift the `firstLayerIndex` on every glyph. But I would have prefere to create a palette for each group of glyph.

I'll try to make some more tests but any advice/explanation are welcomed.

yne:
Okay I got It ! There was no bug, but  browsers simply don't offer OTF palette selection, there is no need to define more than 1 palette.

As an demonstration, the following code shall emit a Font with the Belgian flag mapped to the `!` unicode character.
You can use this font in a link `<a>!</a>` for example.

```js
new opentype.Font({
	familyName: "Color Example",
	styleName: "Flags",
	unitsPerEm: 16,
	ascender: 8,
	descender: 0,
	glyphs: [
		new opentype.Glyph({ name: '.notdef', path: new opentype.Path(), advanceWidth: 8 }),
		new opentype.Glyph({ name: '!', unicode: 0x21, path: new opentype.Path(), advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 1, y: 2 }, { type: 'L', x: 3, y: 2 }, { type: 'L', x: 3, y: 6 }, { type: 'L', x: 1, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 3, y: 2 }, { type: 'L', x: 5, y: 2 }, { type: 'L', x: 5, y: 6 }, { type: 'L', x: 3, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 5, y: 2 }, { type: 'L', x: 7, y: 2 }, { type: 'L', x: 7, y: 6 }, { type: 'L', x: 5, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 0, y: 1 }, { type: 'L', x: 8, y: 1 }, { type: 'L', x: 8, y: 7 }, { type: 'L', x: 0, y: 7 }] }), name: '.', advanceWidth: 8 }),
	],
	tables: {
		colr: {
			baseGlyphRecords: [
				{ glyphID: 1, firstLayerIndex: 0, numLayers: 4 }
			],
			layerRecords: [
				{ glyphID: 5, paletteIndex: 0xFFFF },
				{ glyphID: 2, paletteIndex: 0 },
				{ glyphID: 3, paletteIndex: 1 },
				{ glyphID: 4, paletteIndex: 2 },
			]
		},
		cpal: {
			colorRecords: [0x000000FF, 0x00FFFFFF, 0x0000FFFF], // BGRA
		}
	},
});
```

> Note: The `paletteIndex: 0xFFFF` that I used as flag outline does not seems to be honored on webkit but work fine on gekko:

image

> Another fun fact: Webkit render the custom `@fontface` in `<select>` popups, Gekko don't

Screenshot_2021-10-10_17-42-49

The code is ready for review/feedback.

yne:
Everything is running smooth here.
I transformed a bunch of colored country svg flags into SVG/WOFF (see [flags.zip](https://github.com/opentypejs/opentype.js/files/7493033/flags.zip))

image


And they all looks fine on Chrome/Firefox

@Jolg42 : Is it good for merge (+tag) ?

yisibl:
Please use [nanoemoji](https://github.com/googlefonts/nanoemoji) to generate the COLRv0 font for comparison. nanoemoji already has sufficient test cases to ensure the correctness of the generated font.

In addition, the Chinese flag is wrong. see also: https://en.wikipedia.org/wiki/Flag_of_China

yne:
> nanoemoji already has sufficient test cases to ensure correctness.

I used it and it works on both FF and Chrome but :
- it output a TTF (not OTF)
- does not use the same table format for coordinates
- re-scale every glyph to a 1000x1000 grid (even when I change the `--width`)

So it's not a best way to compare/debug.

I also tried iconfont.cn but they need a number to use it.
Can you show me how to use the website to create a font without account ? (like https://icomoon.io/)

> In addition, the Chinese flag is wrong

Yes, my goal is not to be exact, my goal is to have all 193 flags in a small 3KB file. To do that, I work on a 8x8 SVG grid.

yisibl:
nanoemoji is very flexible and has good baseline alignment. Should work for your needs.

- output cff table and OTF file: `--color_format=cff_colr_0` and `--output_file=foo.otf`
- `does not use the same table format for coordinates` I don't understand the meaning of this statement.
- Adjusting the scaling：`--transform='scale(0.86)'`

The full command looks like this:

```shell
nanoemoji --color_format=cff_colr_0 $(find ../foo/ -name 'ad.svg') --family=foo --output_file=test.ttf --transform='scale(0.86)' --width=1000
```

image

Sorry, iconfont.cn currently requires registration to generate color fonts.

yne:
After realizing that even a basic shape would have [different results](https://github.com/yne/picon/wiki/keep-in-mind#rendering-inconsistency), I gave up on trying to understand those COLR0 rendering kirks.
Anyway, this new Opentype.js COLR0 API is so simple (It only add a table) that's it probably not the cause of those kirks, so I don't see any issues for the merge.

rsheeter:
Chrome 101.0.4950.2+ support font-palette (https://www.w3.org/TR/css-fonts-4/#color-font-support) which may make testing this easier. https://roettsch.es/js_palette.html shows an example courtesy of @drott.

In trying to use opentype.js to build a color picker (present one color input per color in palette and then generate css to adjust the colors of [Bungee Spice](https://developer.chrome.com/blog/colrv1-fonts/#show-me-your-colors)) I found the CPAL model here to be a bit inconvenient to work with, particularly the provision of colorRecords as raw ULongs.

If it's OK to add a friendlier representation than the raw contents of the table then perhaps we could expose something like palettes as a list of something easy to get in terms of r,g,b,a values? - for example, to build a test color picker I added palettes as a property of cpal (exploratory hackery @ https://github.com/googlefonts/opentype.js/pull/1/files) with each entry containing a 4-part array of r,g,b,a so it's easier to work with. Happy to send a cleanup up PR if this seems potentially acceptable.

yne:
The last [`check.argument`](https://github.com/opentypejs/opentype.js/blob/2a6c2f61630ec1f5f580f092e42631e95ad8bb48/src/tables/cpal.js#L33) shall be

```js
if (colorRecordIndices.length > 1) {
  check.argument(numPaletteEntries, 'Can\'t infer numPaletteEntries on multiple colorRecordIndices');
}
```

Sorry about that. I'll push a fix.


```js
const fromRGBA = (hrgb_a) => `${hrgb_a}FF`.slice(1, 9).match(/[a-fA-F0-9]{2}/g);
const toBGRA = ([r, g, b, a='FF']) => Number.parseInt(`${b}${g}${r}${a}`, 16);
// toBGRA(fromRGBA("#112233")).toString(16) => '332211ff'
// toBGRA(fromRGBA("#11223344")).toString(16) => '33221144'
```

Previous Belgian flag example with multiple palettes using a 3 lines `ColorPalette0` wrapper.
(Most of the complexity is in toBGRA as it try to parse all color colors format I could think of)

```html
<script src="opentype.js"></script>
<style>
@font-palette-values --zero {font-family: "ColorExample";base-palette: 0;}
@font-palette-values --one {font-family: "ColorExample";base-palette: 1;}
@font-palette-values --two {font-family: "ColorExample";base-palette: 2;}
</style>
<script>
function toBGRA(c){
	if (typeof c == 'number') return c;
	if (c instanceof Array  ) {console.log(c); return new DataView(new Uint8Array([c[2],c[1],c[0],c.length==3?0xFF:c[3]]).buffer).getInt32(0);}
	if (typeof c == 'object') return new DataView(new Uint8Array([c.b, c.g, c.r, 'a' in c ? c.a : 0xFF]).buffer).getInt32(0);
	if (typeof c == 'string') {
		let hex = c.replace(/^#/,'');
		if(hex.length<=4)hex = hex.replace(/(.)/g,'$1$1');// BED => BBEEDD / C0FE => CC00FFEE
		if(hex.length==6)hex += 'FF'; // implicit full opacity
		const [r, g, b, a] = hex.match(/[a-fA-F0-9]{2}/g);
		return Number.parseInt(`${b}${g}${r}${a}`, 16);
	}
}
const ColorPalette0 = (...palettes) => ({
	colorRecordIndices: palettes.reduce(([res,off],pal)=>[res.concat(off),off+pal.length],[[],0])[0],
	numPaletteEntries: palettes.length,
	colorRecords: palettes.flat().map(toBGRA) // the .map part is for "human" input parsing
});
const colorswap = ([r, g, b, o]) => Number.parseInt(`${b}${g}${r}${o}`, 16);
var test = new opentype.Font({
	familyName: "ColorExample",
	styleName: "Flags",
	unitsPerEm: 16,
	ascender: 8,
	descender: 0,
	glyphs: [
		new opentype.Glyph({ name: '.notdef', path: new opentype.Path(), advanceWidth: 8 }),
		new opentype.Glyph({ name: '!', unicode: 0x21, path: new opentype.Path(), advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 1, y: 2 }, { type: 'L', x: 3, y: 2 }, { type: 'L', x: 3, y: 6 }, { type: 'L', x: 1, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 3, y: 2 }, { type: 'L', x: 5, y: 2 }, { type: 'L', x: 5, y: 6 }, { type: 'L', x: 3, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 5, y: 2 }, { type: 'L', x: 7, y: 2 }, { type: 'L', x: 7, y: 6 }, { type: 'L', x: 5, y: 6 }] }), name: '.', advanceWidth: 8 }),
		new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 0, y: 1 }, { type: 'L', x: 8, y: 1 }, { type: 'L', x: 8, y: 7 }, { type: 'L', x: 0, y: 7 }] }), name: '.', advanceWidth: 8 }),
	],
	tables: {
		colr: {
			baseGlyphRecords: [
				{ glyphID: 1, firstLayerIndex: 0, numLayers: 4 } // render glyph[1] as layered glyph[5,2,3,4]
			],
			layerRecords: [
				{ glyphID: 5, paletteIndex: 0xFFFF }, // shall be the current text color
				{ glyphID: 2, paletteIndex: 0 }, // black
				{ glyphID: 3, paletteIndex: 1 }, // yellow
				{ glyphID: 4, paletteIndex: 2 }, // red
			]
		},
		cpal: ColorPalette0(
			['#000', 'FF0', 'FF0000'],
			[[0x00, 0x80, 0x00], 0xFFFFFFFF, [0xFF,0x00,0x00,0xFF]],
			[{r:255, g:0, b:0}, 0xFFFFFFFF, {r:0, g:0, b:255, a:255}],
		)
	},
});
(new FontFace(test.names.fontFamily.en, test.toArrayBuffer())).load().then((ff) => document.fonts.add(ff));
</script>
<body style="font-size:8em;font-family:ColorExample">
<span>!</span> default<br>
<span style=font-palette:--zero>!</span> zero<br>
<span style=font-palette:--one>!</span> one<br>
<span style=font-palette:--two>!</span> two<br>
</body>
```

Preview on chromium 102.0.4967.0
image

rsheeter:
> I'm unsure about adding extraneous "human friendly" argument to the parser/builder. As they won't necessary be reflective (the current parser output can be given as the builder input).

Sorry, I should have clarified that in a tidied up version I would remove some of the current fields and update makeCpalTable to match.

If it's OK to modify the parseCpalTable and update the makeCpalTable accept the same I'd be happy to send a PR that does that. Would that be acceptable? - I'm unclear on whether we are locked into the current structures given that this PR got merged.

yne:
@rsheeter 
Judging by the other `src/tables/*.js` implementations,
`.tables:{}` sub definition are meant to be close to they final result.

But we still could have a nice top-level definition (like `.familyName` ...):

```js
// helper-less palettes+layers example
new opentype.Font({
	familyName: "ColorExample",
	unitsPerEm: 16,
	glyphs: [
		new opentype.Glyph({name:'x',path:new opentype.Path(...), layers:[
			new opentype.Glyph({path: new opentype.Path(...), paletteIndex:0}),
			new opentype.Glyph({path: new opentype.Path(...), paletteIndex:1}),
		]}), // no .paletteIndex == textColor (0xFFFF)
		new opentype.Glyph(...),
	],
	palettes: [
		[0xDEADC0FE, 0xBEEFC0FE], // default font-palette
		[0xDAD1C0FE, 0xBABEC0FE], // using @font-palette-values + font-palette:
	]
});
```

What do you think ?

benoitwimart:
Hi,
I don't understand why name value is '.' and how to know the glyphID number, 
```JS
new opentype.Glyph({ path: Object.assign(new opentype.Path(), { commands: [{ type: 'M', x: 1, y: 2 }, { type: 'L', x: 3, y: 2 }, { type: 'L', x: 3, y: 6 }, { type: 'L', x: 1, y: 6 }] }), name: '.', advanceWidth: 8 }),```

Connum:
That's just some example code. Of course the glyph name can be different. And you can use `font.nameToGlyphIndex()` or `font.charToGlyphIndex()` in order to get the glyph ID.

benoitwimart:
Damn! https://unpkg.com/opentype.js@1.3.4/dist/opentype.js is not working with COLR ; 
I found another version and it's work, so good! Thank you! 
Where is the best version ready to use (with JSDoc) ?

Connum:
Please see the release notes for 1.3.4: https://github.com/opentypejs/opentype.js/releases/tag/1.3.4

PR #491, opened 2021-10-05
GPOS lookupType 9 support

AlexandreRivet:
## Description
Add support for GPOS lookupType 9 to be able to manage other extended lookupType. 🤩 

## Motivation and Context
I noticed, by using this font [WorkSans](https://fonts.google.com/specimen/Work+Sans), I was not able to get kerning values. Diving into this problem, I found it was because GPOS lookupType 9 was not supported and only lookupType 2 was used in getKerningTables.

## How Has This Been Tested?
- Added an unit test included lookupType 9 data
- Tested on WorkSans font. Now we can get kerning values on this one 💪 
- Tested on ProximaNova font to ensure that already lookupType 2 still working as expected

## Types of changes

## Checklist:

### DIFF
--- a/src/position.js
+++ b/src/position.js
@@ -36,6 +36,8 @@ Position.prototype.getKerningValue = function(kerningLookups, leftIndex, rightIn
         const subtables = kerningLookups[i].subtables;
         for (let j = 0; j < subtables.length; j++) {
             const subtable = subtables[j];
+            // Subtables not supported come with an error
+            if (subtable.error) continue;
             const covIndex = this.getCoverageIndex(subtable.coverage, leftIndex);
             if (covIndex < 0) continue;
             switch (subtable.posFormat) {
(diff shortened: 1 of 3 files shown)


ILOVEPIE:
@AlexandreRivet please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

src/position.js
```diff
@@ -36,6 +36,8 @@ Position.prototype.getKerningValue = function(kerningLookups, leftIndex, rightIn
         const subtables = kerningLookups[i].subtables;
         for (let j = 0; j < subtables.length; j++) {
             const subtable = subtables[j];
+            // Subtables not supported come with an error
+            if (subtable.error) continue;
```
ILOVEPIE, line 40: Maybe make it log the error when it comes across one?

ILOVEPIE (changes requested):
Please rebase on the latest code and make requested changes, otherwise this looks good.

Connum:
Pinging @AlexandreRivet

rafallyczkowskiadylic:
This will be covered by https://github.com/opentypejs/opentype.js/pull/557

Connum:
> This will be covered by #557

I was about to close this but tested it before just in case, and with this PR `getKerningTables()` does return a table for the WorkSans font, but with #557 it doesn't. I'm all for fixing this in one PR instead of two, but we have to keep this one open until the other one implements this correctly.

PR #497, opened 2021-11-20
classify glyphset

jimmywarting:
## Motivation and Context
part of #247 as a modernization

## How Has This Been Tested?
`npm run test`

## Types of changes
- Refactor to modern code (no behaviour changes)

## Checklist:

### DIFF
--- a/src/glyphset.js
+++ b/src/glyphset.js
@@ -22,74 +22,76 @@ function defineDependentProperty(glyph, externalName, internalName) {
  * A GlyphSet represents all glyphs available in the font, but modelled using
  * a deferred glyph loader, for retrieving glyphs only once they are absolutely
  * necessary, to keep the memory footprint down.
- * @exports opentype.GlyphSet
- * @class
- * @param {opentype.Font}
- * @param {Array}
  */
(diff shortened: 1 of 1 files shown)


Connum:
I create PR https://github.com/cirbuk/opentype.js/pull/1 on the feature branch to get this reviewed and merged

ILOVEPIE:
@Connum The question is though, do we want to switch to classes? This breaks compatibility with legacy browsers, especially in the standalone non-npm versions of opentype.js. Some applications may depend on opentype.js for polyfilling webfonts in legacy browsers, which means that we'd be breaking those polyfills. Maybe there's a compiler pass we could apply to convert from classes to functions or from classes to Object.create when compiling non-npm versions?

ILOVEPIE:
As long as we can keep it ES5 compatable I don't see an issue, but we need to make sure we can do that.

Connum:
Doesn't rollup with the buble and commonjs plugins make sure that everything's compiled the way we need it to? I haven't used rollup in myself projects before and haven't looked into it more deeply yet, but that's what I understood from a quick glance at it.

PR #503, opened 2022-02-12
Feat: loadFromUrl supports Node.js environment

antoineol:
## Description

This solution provides a node implementation to fetch the font file from the URL, using standard http/https modules (to avoid adding a dependency like axios).

Btw the typing of `load()` does not include the third parameter that allows to use `loadFromUrl`. I don't know who is maintaining it.

## Motivation and Context

loadFromUrl assumes it is run in the browser environment only, but I needed it in Node.js environment. I tried to polyfill XHR2, but the polyfill seems to have a bug: something hangs for 120 seconds. So I provide a nodejs implementation here.

## How Has This Been Tested?

The function loadFromUrl was tested with code like:

```ts
  const url = 'https://github.com/Templarian/MaterialDesign-Webfont/blob/master/fonts/materialdesignicons-webfont.ttf?raw=true';
  const res = await new Promise<ArrayBufferLike>((resolve, reject) => {
    loadFromUrl(url, (error, response) => {
      if (error) reject(error);
      else resolve(response);
    });
  });

  console.log(new TextDecoder().decode(res));
```

And I tested a larger workflow with something like `load(url, undefined, { isUrl: true })`, ensuring it returns the font with the data I expected.

Environment: node. No impact noticed on the rest of the library. All unit tests are still passing (with just 2 warnings that were already there).

N/A

## Types of changes

Counted as new feature, but I'm tempted to say it's a bug, since it uses the existing API in node, which is supposed to be a supported environment. :)

## Checklist:

- [X] I did `npm run test` and all tests passed green (including code styling checks). => 2 warnings for code styling that were already there when I forked the master.


yne:
I just learned that [Node 18](https://nodejs.org/en/) natively support [`fetch()`](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V18.md#fetch-experimental)

Connum:
I just made a [PR on the fork](https://github.com/antoineol/opentype.js/pull/1) to add a test for this.

ILOVEPIE:
> @ILOVEPIE maybe we need some kind of roadmap to keep track of things like this?
@Connum milestones were being used before, we can setup new ones, also we should start using the discussions tab too.

ILOVEPIE:
@Connum we should get this reviewed and merged.

Connum:
Alternatively, instead of waiting for @antoineol, we could merge this PR and add the test with a separate PR.

ILOVEPIE:
@Connum I think it's bests to keep tests in the same PR. I'm not too fond of the idea of committing something to master without testing it.

Connum:
Pinging @antoineol

If we don't hear back, we should make a new PR with these changes and added tests

Connum (commented):
I finally managed to merge the test myself with the help of gh CLI.

Connum:
Just adapted the test because it didn't pass anymore due to the changes in the names object

test/opentypeSpec.js
```diff
@@ -26,6 +26,19 @@ describe('opentype.js', function() {
         });
     });
 
+    it('can load a font from URL in Node context', function(done) {
+        load('https://opentype.js.org/fonts/Roboto-Black.ttf', null, { isUrl: true }).then((font) => {
+            assert.deepEqual(font.names.fontFamily, {en: 'Roboto Black'});
+            assert.deepEqual(font.names.fontFamily, {en: 'Roboto Black'});
```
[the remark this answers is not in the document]
ILOVEPIE: @Connum  why is this checked twice?

test/opentypeSpec.js
```diff
@@ -28,14 +28,16 @@ describe('opentype.js', function() {
 
     it('can load a font from URL in Node context', function(done) {
         load('https://opentype.js.org/fonts/Roboto-Black.ttf', null, { isUrl: true }).then((font) => {
-            assert.deepEqual(font.names.fontFamily, {en: 'Roboto Black'});
-            assert.deepEqual(font.names.fontFamily, {en: 'Roboto Black'});
+            assert.deepEqual(font.names.macintosh.fontFamily, {en: 'Roboto Black'});
+            assert.deepEqual(font.names.macintosh.fontFamily, {en: 'Roboto Black'});
```
ILOVEPIE, line 32: why is this done twice?

ILOVEPIE (dismissed):
We should avoid redundant checks in the code.

test/opentypeSpec.js
Connum: Because I copy-pasted too much! ;) It now checks for the windows name as well, like the other tests do!

Connum (approved):

Connum (approved):

Connum:
added `http` and `https` as externals to make this run with esbuild, and fixed the test that relied on a no longer existing font file.

Issue #183, opened 2016-04-09
Add support for WOFF2


kennethmac2000:
WOFF2 is now at Candidate Recommendation stage and is widely supported by the most popular browsers (including Chrome and Firefox).

Is it time to add support for it to opentype.js?

fdb:
Yes.

Can you recommend a good Brotli library for JS?

fdb:
For reference, the WOFF2 W3 doc: https://www.w3.org/TR/WOFF2/

joseluis:
I believe only the decompressor is needed, right? These are the libs I know of:

1. https://github.com/devongovett/brotli.js/tree/master/dec/ decode.js
  - Instead of using dictionary-data.js use dictionary-browser.js and dictionary.bin. The total extra size may well be less than 100kb.
1. https://github.com/dominikhlbg/brotlijs ([discussion](https://groups.google.com/forum/#!topic/brotli/Y28CS-a7QHo))

fdb:
Coo, I'll take a look.

Encoding would be cool as well as it would allow to write woff2 fonts (we don't do this for WOFF at the moment, but would like to support it there as well).

devongovett:
I'm using my [brotli.js](https://github.com/devongovett/brotli.js) project in fontkit. Recently ported the decoder to JS by hand instead of using Emscripten. That ended up being much faster and smaller.

There's a bit more to decoding WOFF2 fonts than just brotli though. Here's the source code from fontkit: https://github.com/devongovett/fontkit/blob/master/src/WOFF2Font.coffee.

fdb:
Thanks @devongovett! I know it's not just the decompression, but that's the part I didn't want to do myself :-)

nazikus:
can someone provide a code sample how to decompress .woff2 file and parse it with opentype.js in browser?

fdb:
Only woff is supported. Woff2 is not yet supported.

caub:
this other repo seems to support woff2

kontur:
From my superficial understanding woff2 is just a compression format. What in particular needs to get solved for this library to support woff2?

Jolg42:
@kontur so the first official implementation of WOFF2 was not compatible to use from a browser (C++ & Node.js only).

But today thanks to [brotli.js](https://github.com/devongovett/brotli.js) this is now doable!

For me supporting WOFF2 is as easy as integrating brotli.js in the project, so feel free to contribute 😉

kontur:
[...]

PR #518, opened 2022-06-06
Gracefully handle unsupported WOFF2 parsing

yne:
## Description

WOFF2 is becoming more and more popular and is now the default font format in font.google.com

So Opentype.js is more and more expected to parse and generate WOFF2.

However WOFF2 compression (as WASM) would bloat the opentype.js dist file by ~1400 KB.

A more modular approach would be to let the end-user do the WOFF compress/decompress wrapping if they need it.

## Motivation and Context

This PR aim to solve this lack of support by pointing the user to [a wrapping example](https://github.com/opentypejs/opentype.js/issues/183#issuecomment-1147228025) if they try the directly load a WOFF2 file into opentype.js

## How Has This Been Tested?

By loading a WOFF2 into opentype.js without decompresing it beforehand and looking at the reason given by the trown exception.

image

### DIFF
--- a/src/opentype.js
+++ b/src/opentype.js
@@ -207,6 +207,9 @@ function parseBuffer(buffer, opt) {
 
         numTables = parse.getUShort(data, 12);
         tableEntries = parseWOFFTableEntries(data, numTables);
+    } else if (signature === 'wOF2') {
+        var issue = "https://github.com/opentypejs/opentype.js/issues/183#issuecomment-1147228025"
+        throw new Error('WOFF2 require an external decompressor library, see examples at: ' + issue);
     } else {
         throw new Error('Unsupported OpenType signature ' + signature);
     }


herrstrietzel:
@yne : Thank you so much for depicting this woff2 workaround – that's  kind of a game changer!  
First of all, this solution should be at least mentioned in the **current documentation.**     
(I've discovered this approach rather by accident.)  

In the long run, it might be better to include the admittedly complex brotli decompression by default – albeit I always appreciated opentype.js approach to keep the core as compact as possible.  

Highly subjective: I think font parsing is rather a niche for creating graphics based on fonts.    
You rarely need to include opentype.js as a mandatory script asset (like other front end libraries).  
My point: the additional increase in file size might be negligible in favor of woff2 support.

Connum:
This has been discussed before. Brotli decompression will not be integrated into opentype.js, but we do show a console message with a link to a description when trying to load a woff2 file.

PR #523, opened 2022-07-20
Don't attempt to parse meta table tags "appl" and "bild". Fixes #522

### LINKED ISSUE #522: Safari/Chrome/Chromium: RangeError: Maximum call stack size exceeded (parsing  meta Table tags 'appl' and 'bild')
In Google Chrome/Chromium/Safari attempting to parse a font from Apple (which I can't share/attach) I get an Error:

```
RangeError: Maximum call stack size exceeded
        //         [...] at decode.UTF8 [...] "
```

The font in a current macOS Monterey 12.4  is "San Francisco"/"System Font" and located at `/System/Library/Fonts/SFNS.ttf` (Version 17.0d11e1 can be checked via https://wakamaifondue.com/)

It could be argued that Chrome/Chromium/Safari should handle this case better (Firefox does), however, the implementation can improve here IMO as well.

## Expected Behavior

The font should parse without error.

I'm going to send a PR with a possible fix and link here to it: #523

## Current Behavior

In `src/tables/meta.js` the function `parseMetaTable` tries to parse `appl` and `bild` using ` decode.UTF8`.

Both tags, `appl` and `bild` are "Reserved — used by Apple" as of [the spec](https://docs.microsoft.com/en-us/typography/opentype/spec/meta#metadata-tags). Parsing these does currently not yield a usable result, especially because it doesn't seem to be UTF-8 at all.

In Google Chrome/Chromium/Safari I have seen it fail attempting to parse this with: `RangeError: Maximum call stack size exceeded`

This indicates that unknown/proprietary tags should not be attempted to parse this way, also, the `decode.UTF8` implementation  is likely insufficient to parse full UTF-8 (but this would be a different bug), both formats of the the meta tags `dlng` and `slng` are documented as: "Text, using only Basic Latin (ASCII) characters", hence there should be no problem.

## Possible Solution

Skip parsing of the tags `appl` and `bild`, or skip all unknown tags (all but  `dlng` and `slng`).

1. open Google Chrome, Chromium or Safari browser
2. go to https://opentype.js.org/
3. open the Apple System Font from `/System/Library/Fonts/SFNS.ttf`
4. will display message "RangeError: Maximum call stack size exceeded"

## Your Environment

* Version used: latest/HEAD/nightly
* Font used: macOS `/System/Library/Fonts/SFNS.ttf` Version 17.0d11e1
* Browser Name and version:
    * macOS Safari Version 15.5 (17613.2.7.1.8)
    * macOS Google Chrome Version 103.0.5060.114 (Official Build) (arm64)
    * Linux Google Chrome Version 102.0.5005.115 (Official Build) (64-bit)
    * Linux Chromium Version 102.0.5005.115 (openSUSE Build) stable (64-bit)
* Operating System and version (desktop or mobile): macOS Monterey 12.4, GNU+Linux+Gnome3

graphicore:
As described in  #522: Attempting to parse certain fonts that define `appl` and `bild` tags in the `meta` tabke crashes Google Chrome/Chromium/Safari with "RangeError: Maximum call stack size exceeded."

## How Has This Been Tested?

See the detailed explanation in the issue. The font loads on the page served with `npm start` and this version of the code.

## Notes:

* instead of dropping the known incompatible/non-ASCII tags (`appl`, `bild`), we could drop all tags that are unknown (not `dlng`, `slng`)
* instead of dropping the tags, we could keep them, but use the un-parsed binary data. This way, the font could be re-build again without loss

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/meta.js
+++ b/src/tables/meta.js
@@ -21,6 +21,8 @@ function parseMetaTable(data, start) {
         const tag = p.parseTag();
         const dataOffset = p.parseULong();
         const dataLength = p.parseULong();
+        if (tag === 'appl' || tag === 'bild')
+           continue;
         const text = decode.UTF8(data, start + dataOffset, dataLength);
 
         tags[tag] = text;


graphicore:
Here's a quote form the [Apple spec](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6meta.html):


> ### The 'appl' and 'bild' tag
>
>These tags are used internally by Apple. They use binary, and not text, data.
>
>Font designers are not limited to the use of these metadata tags. Any metadata tag beginning with a capital letter (e.g., 'Foob') is to be considered private use. Other cross-platform tags will be added as needed.
>
>(Note that the 'appl' and 'bild' tags are strictly-speaking private use, as they are Apple-internal tags. Their use is retained for historical reasons.)"

PR #530, opened 2022-12-15
create post table from existing data if available

### LINKED ISSUE #529: post table is created with defaults when calling toArrayBuffer
## Expected Behavior
calling toArrayBuffer should create a font that has the same post table as the font it was created from

since the post table is created from scratch with defaults with post.make() things like underlineThickness and underlinePosition
are always set to 0 instead of using the values from the original post table

Here is the offending code:
https://github.com/opentypejs/opentype.js/blob/2a6c2f61630ec1f5f580f092e42631e95ad8bb48/src/tables/sfnt.js#L289

I would expect something like this:
https://github.com/opentypejs/opentype.js/blob/2a6c2f61630ec1f5f580f092e42631e95ad8bb48/src/tables/sfnt.js#L309
https://github.com/opentypejs/opentype.js/blob/2a6c2f61630ec1f5f580f092e42631e95ad8bb48/src/tables/sfnt.js#L312
https://github.com/opentypejs/opentype.js/blob/2a6c2f61630ec1f5f580f092e42631e95ad8bb48/src/tables/sfnt.js#L315

## Current Behavior
font.toArrayBuffer() creates font with post table defaults

## Possible Solution
parse values from the font's post table to create the new set of tables

load a font with a value for tables.post.underlineThickness != 0
call:
font.toArrayBuffer()
use the font with underline, see that it is not underlined

## Context
have working underlining in an SVG

## Your Environment
* Version used: 1.3.4
* Font used: any font
* Browser Name and version: N/A
* Operating System and version (desktop or mobile): N/A
* Link to your project: N/A

cthorner:
copy settings from existing post table if available

## Description
At the moment, when calling toArrayBuffer() on the font, or download() it will create a new ```tables.post``` with all values set to 0(default) instead of using the values for the font.

## Motivation and Context
This fixes the issue with underlines not showing up in SVGs.
This fixes #529.

## How Has This Been Tested?
I verified that this worked with a few different fonts. Before the change the underline would be invisible and after the change it is visible.

## Types of changes

## Checklist:


ILOVEPIE:
@cthorner please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

Connum (changes requested):
LGTM, but could you please bring it up-to-date with the current master and check if the tests still run fine?

src/tables/post.js
```diff
@@ -50,18 +50,28 @@ function parsePostTable(data, start) {
...
+  const {
+    italicAngle = 0,
+    underlinePosition = 0,
+    underlineThickness = 0,
+    isFixedPitch = 0,
+    minMemType42 = 0,
+    maxMemType42 = 0,
+    minMemType1 = 0,
+    maxMemType1 = 0
+  } = postTable || {};
```
ILOVEPIE, line 63: Destructuring is a relatively new feature, does our build system correctly convert it to ES5?

ILOVEPIE (commented):
@Connum little concerned about the use of destructuring, does our build system convert this to ES5? Can we verify that before we merge this?

src/tables/post.js
Connum: Yes, it does! I just verified it today when I checked if template literals are converted to string concatenation, which it also does. So we're free to use newer language features (but it doesn't hurt to check from time to time)!

ILOVEPIE (approved):
Looks good to me.

ILOVEPIE:
@Connum need you to re-review so you don't have "requested changes" status anymore.

Connum (approved):

PR #531, opened 2022-12-17
mark as side effect free

hipstersmoothie:
## Description

Marked the package as side effect free so it can be tree shaken by webpack if not used

## Motivation and Context

I saw that this package was included in my bundle. I'm not actually sure if this dep is side effect free.... is it?

## How Has This Been Tested?

In my app

## Checklist:


ILOVEPIE:
@hipstersmoothie please rebase onto and create a pull request for the new active development fork [here](https://github.com/SABRE-JS/opentype.js)

ILOVEPIE:
@Jolg42 is the library side-effect free, if so this is any easy change we could make.

ILOVEPIE:
> @Jolg42 is the library side-effect free, if so this is any easy change we could make.

@Connum how about you?

Connum:
> @Connum how about you?

Sorry, I can't provide an answer to this one. I don't have an in-depth understanding of tree shaking and side effects and I'm not deep enough into the code base yet to have an understanding of what all the different parts of the code are doing.

me4502:
I've looked at a bunch of files, and it doesn't appear any side effects are present.

Generally what you want to look for is any instance where just the act of importing a file will trigger something to happen. So function calls, etc at the top level. Anything that just sets up a variable to be exported is fine, as that's all repeatable & doesn't modify anything aside from exported entries.

Connum:
> I've looked at a bunch of files, and it doesn't appear any side effects are present.
> 
> Generally what you want to look for is any instance where just the act of importing a file will trigger something to happen. So function calls, etc at the top level. Anything that just sets up a variable to be exported is fine, as that's all repeatable & doesn't modify anything aside from exported entries.

Thanks for the valuable input and for looking into this. I haven't encountered anything that seems to hint at side effects while working on this library so far then. So in my opinion we can merge this.

ILOVEPIE (approved):
Looks good to me, as long as the library is actually side-effect free.

PR #548, opened 2023-02-05
log warning instead of throwing an error when encountering an unknown ClassDef format

### LINKED ISSUE #533: Error `0x963a: ClassDef format must be 1 or 2` since upgrading from `1.3.3` to `1.3.4`
## Expected Behavior
The attached file loads without issues on `opentype.js@1.3.3`

## Current Behavior
It fails to load with `opentype.js@1.3.4` with error `x963a: ClassDef format must be 1 or 2`.

This is the source font file: [96025-ld-soft-serif.ttf.zip](https://github.com/opentypejs/opentype.js/files/10275691/96025-ld-soft-serif.ttf.zip)

```js
import { load } from "opentype.js";

load("src/96025-ld-soft-serif.ttf", function (err) {
  if (err) {
    // when using `opentype.js@1.3.4`
    console.log("Font could not be loaded: " + err.message);
  } else {
    // when using `opentype.js@1.3.3`
    console.log("Successfully loaded");
  }
});
```

## Your Environment

* Version used: `1.3.4`
* Font used: see file attached
* Browser Name and version: Google Chrome @ `108.0.5359.124`
* Operating System and version (desktop or mobile): MacOS Ventura 13.1

Connum:
## Description
This fixes fixes #533 by logging a warning into the console and proceeding to parse the font instead of aborting by throwing an Error.

## Motivation and Context

Some fonts might include an unsupported format identifier (something other than 1 or 2) for a ClassDef table format. These fonts may still render correctly, but prior to this change the lib wouldn't load them at all.

## How Has This Been Tested?

Loaded the font file provided in #533 before and after the change. Loaded other fonts and ran all tests as well as a build to make sure it doesn't introduce new errors.

## Types of changes

## Checklist:


ILOVEPIE:
should we really do this? technically these fonts are invalid.

Connum:
> should we really do this? technically these fonts are invalid.

According to  [you yesterday](https://github.com/opentypejs/opentype.js/issues/533#issuecomment-1416847922), we should! 😉
And I think it's better indeed than not loading the font at all! I was able to load the font in other software without issues (except maybe a warning along the lines of what I'm showing now with that change).

ILOVEPIE:
> According to [you yesterday](https://github.com/opentypejs/opentype.js/issues/533#issuecomment-1416847922), we should! wink

Sorry, i'm a little forgetful at times.

src/parse.js
```diff
@@ -498,7 +498,11 @@ Parser.prototype.parseClassDef = function() {
             })
         };
     }
-    throw new Error('0x' + startOffset.toString(16) + ': ClassDef format must be 1 or 2.');
+
+    console.warn(`0x${startOffset.toString(16)}: Unknown ClassDef format ${format}.`);
```
ILOVEPIE, line 502: The error should probably let them know the font is likely corrupt so that they know to re-acquire it, instead of posting a bug report.

ILOVEPIE (changes requested):
We should make this error a little more descriptive.

Connum (commented):
Should be clearer now

ILOVEPIE (approved):
Looks good to me.

PR #549, opened 2023-02-05
make jshint and jscs ignore the /.git folder

Connum:
I noticed that that jscs showed messages about files in the /.git directory (e.g. WIP files from local branches), so I added it to the ignore lists of both jscs and jshint.


ILOVEPIE (approved):
Looks good to me.

Issue #501, opened 2022-01-06
Fallback to KERN table if GPOS table is empty in `Font.getKerningValue`


miere43:
## Expected Behavior
`Font.getKerningValue` returns kerning value from KERN table if GPOS table is empty.

## Current Behavior
`Font.getKerningValue` returns 0 if GPOS table is empty, even if there are values in KERN table.

## Possible Solution
1. Maybe GPOS parsing is broken and doesn't work for Arial font (not sure).
2. Fallback to KERN table if `font.position.defaultKerningTables.length === 0` is zero in `Font.getKerningValue`. See:
  https://github.com/miere43/opentype.js/commit/6acfd4e41b7aa2901f84cfd49433a6e63bf60cbe

## Steps to Reproduce (for bugs)
```js
const arial = opentype.loadSync('C:\\Windows\\Fonts\\arial.ttf');
const kerning = arial.getKerningValue(arial.charToGlyph('T'), arial.charToGlyph('e'));
console.log('kerning is', kerning);
```
Output:
```
kerning is 0
```
Expected:
```
kerning is -227
```
## Context
Drawing text.

## Your Environment

* Version used: 1.3.4
* Font used: Arial
* Browser Name and version: ---
* Operating System and version (desktop or mobile): Windows 10
* Link to your project: ---

rafallyczkowskiadylic:
@ILOVEPIE this is fixed within https://github.com/opentypejs/opentype.js/pull/557 I tested this using test code snippet

```
describe('Arial fonts test', () => {
        it('should have a correct kerning value', () => {
            const font = loadSync('./fonts/arial.ttf');
            const kerning = font.getKerningValue(font.charToGlyph('T'), font.charToGlyph('e'));
            assert.strictEqual(kerning, -227);
        });
    });
```

unfortunately I cannot commit this because of arial font's License. 

![image](https://github.com/opentypejs/opentype.js/assets/102729629/081b8a68-5d51-4c96-975d-bed44ff266b4)

PR #557, opened 2023-02-06
feat: Support for Mark to Base Attachment Positioning (incl. GPOS 4, 9)

### LINKED ISSUE #515: OpenType.js can't read Google fonts' kerning, eg. EB Garamond
Download the ttf from https://fonts.google.com/specimen/EB+Garamond. use it as a test. The fonts have kerning info, and can be used as regular desktop fonts with proper kerning. However, OpenType.js can't read its kerning data. letter AV kerning should be -140, but OpenType.js returns 0

1. go to https://codepen.io/wendyz/pen/PoQJyoX?editors=0011
2. open the console
3. OpenType.js return AV Kerning to be 0, where it should be -140
image

When I open the EB Garamond font with Robofont, I can see there is kerning info in the kern center. AV kern should be -140
image

Extra info: I download EB Garamond 8pt (another version) from the type designer's Bitbucket repo, and OpenType.js output proper kerning for that font. Included in the codepen example. cc. @davelab6

It may affect other Google Fonts as well, eg.
- DM Serif Text

## Your Environment

* Version used: 1.3.4
* Font used: EB Garamond
* Browser Name and version: Chrome
* Operating System and version (desktop or mobile): Mac and Windows
* Link to your project: https://codepen.io/wendyz/pen/PoQJyoX?editors=0011

rafallyczkowskiadylic:
## Description

Added support for Mark to Base Attachment Positioning (for the DFLT script). Feature is enabled by default because it is a part of a proper glyphs positioning. Includes capabilities implementation for the corresponding parsers (GPOS lookup 4 & 9)

Followed the current solution's convention to introduce new features within the font & layout implementation.

- added GPOS `LookupType 4` parser for `MarkBasePosFormat1` support
- added GPOS `LookupType 9` parser for Extension Positioning `SubtableFormat1` support
- added GSUB support for `multiSubstitutionFormat1` 
- introduced the union lookup feature (`getFeaturesLookups`) for the layout to meet the OT specs
- fixed GSUB substitutions algorithm 
- tests coverage for kerning and positioning features
- tests coverage for all supported and not supported lookups of GPOS

**TODO**: Add support for non DFLT scripts for `MARK` (`src/font.js:344`)
**TODO**: Add a support device offsets for anchor table format 3 (`src/parse.js:515`)
**TODO**: Add MarkToMark Attachment positioning (lookup type 6) functionality.

### Positioning changes

Fonts like Mandarin or Thai to successfully render text with all glyphs (sometimes 3+) representing characters like `ปั` or `ริ่`
requiring the `LookupType 4` for mark to base attachment positioning to do it properly. 

[Extension subtables support](https://learn.microsoft.com/en-us/typography/opentype/spec/gpos#lookuptype-9-extension-positioning) is **required** to support the subtables that exceeds the 16-bit limits of the various other offsets in the GPOS table - that genuinely are being used eg. for the [Apples SF Pro fonts](https://developer.apple.com/fonts/) and many others.

### Substitution changes

There was an issue with the current algorithm where it was collecting all substitutions from all lookup tables and after that each substitution was conducted. This is invalid. Because it should apply substitution for each lookup table before moving to the next lookup table. In other words: substitution should be applied before moving to the next lookup table 

Specification:
> During text processing, a client applies a lookup to each glyph in the string before moving to the next lookup. A lookup is finished for a glyph after the client locates the target glyph or glyph context and performs a substitution, if specified. To move to the “next” glyph, the client will typically skip all the glyphs that participated in the lookup operation: glyphs that were substituted as well as any other glyphs that formed a context for the operation.

Ref: https://learn.microsoft.com/en-us/typography/opentype/spec/gsub#table-organization

### The union lookup feature

The union lookup feature (`getFeaturesLookups`) was required actually to meet the [OpenType specification](https://learn.microsoft.com/en-us/typography/opentype/otspec191alpha/chapter2#lookup-list-table) which guarantees that all enabled features will be processed in a valid order both for positioning and substitutions.
> the list of lookups is the union of lookups referenced by all of those features, and these are all processed in their lookup list order.

## How Has This Been Tested?

Added tests coverage to the `test/parser.js` and `test/tables/gpos.js` including new binary data and testing:
- the [Lookup Type 4: Mark-to-Base Attachment Positioning Subtable specification](https://learn.microsoft.com/en-us/typography/opentype/spec/gpos#lookup-type-4-mark-to-base-attachment-positioning-subtable)
- the [LookupType 9: Extension Positioning](https://learn.microsoft.com/en-us/typography/opentype/spec/gpos#lookuptype-9-extension-positioning)

These lookups uses existing structures like coverage tables, and were tested agains different formats.  Extension positioning lookup was tested agains all currently supported and not supported subtable parsers. 

Added tests coverage to the `test/layout.js` to test the method for union lookups (`getFeaturesLookups`).
Added tests coverage for glyph positioning tests in the `test/font.js` including:
- Kerning feature support with kerning table and lookup table support
- `options.kerning` and  `options.features.kern` testing 
- Mark To Base Attachment feature support to work with kern features
- `options.features.mark` testing
Added tests coverage to the `test/bidi.js` to test `multiSubstitutionFormat1` glyphs decomposition 

Moreover there were many test conducted for the rendered content using `latn` and `arab` fonts.

### Positioning issues

BEFORE (INVALID glyphs pos):
image

AFTER (FIXED):
image

### Substitution issues

BEFORE (INVALID) - very end glyph is a single glyph that should be decomposed into two.
image

AFTER (FIXED) - decomposed into two glyph: 
image

image

## Types of changes

## Checklist:


Connum:
Does this make #491 redundant, or do we have to consolidate the two? Could you please have a look at each other's code @rafallyczkowskiadylic & @AlexandreRivet?

ILOVEPIE:
@rafallyczkowskiadylic I noticed that rlig is missing an implementation for latin scripts and liga is missing an implementation for arabic scripts, is that also fixed in this PR?

rafallyczkowskiadylic:
@Connum https://github.com/opentypejs/opentype.js/pull/491 is related, but not redundant. These MRs have different features covered. But esentially both have a required `rlig` missed :-) 

My proposition for modus operandi:
Let's focus on https://github.com/opentypejs/opentype.js/pull/491 in first place. I will commit and cover the missing `rlig` over there. By the time I will be working on that please review this MR and drop questions or any concerns so we can discuss. 

@ILOVEPIE These two MRs are considered as separate chunks of features. I will fix whatever will be needed starting with the smaller one https://github.com/opentypejs/opentype.js/pull/491

ILOVEPIE (changes requested):
I could only find one issue with the code besides some naming choices for constants. I'm going to get @Connum to review this as well as it's a larger PR and I want to make sure that it's ok.

ILOVEPIE:
Also, @rafallyczkowskiadylic does this PR fix #501, because if you're redoing this code anyways, it might be worth fixing it.

src/substitution.js
```diff
@@ -12,7 +12,10 @@ import Layout from './layout';
  * @constructor
  */
 function Substitution(font) {
-    Layout.call(this, font, 'gsub');
+    Layout.call(this, font, 'gsub', [
+        { featureName: 'rlig',  supportedLookups: [4] }
+        // TODO: Define all supported features to use layout.getFeaturesLookups for a sequence ordered feature lookups
```
Connum, line 17: Is this a TODO for some separate future PR, or did you forget to implement this? How important is this, shouldn't it be in this PR?

src/layout.js
```diff
@@ -194,6 +195,61 @@ Layout.prototype = {
...
+     * @param {string} [script='DFLT']
+     * @param {string} [language='dlft']
+     * @return {Object[]} an ordered lookup list of requested features
+     */
+    getFeaturesLookups: function(requestedFeatures, script, language) {
+        if (!this.font.tables[this.tableName] || !requestedFeatures) {
+            return [];
+        }
+
+        // Fitler out only supported by layout table features
```
Connum, line 217: just y small typo, Fitler => Filter

src/parse.js
```diff
@@ -477,6 +477,74 @@ Parser.prototype.parseCoverage = function() {
...
+            };
+        case 2:
+            return {
+                format,
+                xCoordinate: this.parseShort(),
+                yCoordinate: this.parseShort(),
+                anchorPoint: this.parseUShort()
+            };
+
+        // TODO: Add a support Device offsets
```
Connum, line 533: As with the other TODO, has this been accidentally forgotten, or why is it not implemented with this PR? (not supposed to sound like an accusation/demand, just would like to know what's the reason)

Connum (changes requested):
I have to admit that all that byte-level-parsing stuff is a bit out of my comfort zone (yes, despite me having implemented a few parsing features before ;)), but other than @ILOVEPIE's comments and those I added, there's nothing that caught my eye.

Connum:
@rafallyczkowskiadylic the feature order is quite a critical thing to add before we start working on implementing more features like `calt` etc. It would be great if we could merge this soon!

src/parse.js
```diff
@@ -477,6 +477,74 @@ Parser.prototype.parseCoverage = function() {
...
+        // https://learn.microsoft.com/en-us/typography/opentype/otspec191alpha/gpos_delta#anchor-table-format-3-design-units-plus-device-or-variationindex-tables
+        case 3:
+            return {
+                format,
+                xCoordinate: this.parseShort(),
+                yCoordinate: this.parseShort(),
+                xDevice: 0x00,
+                yDevice: 0x00,
```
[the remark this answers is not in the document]
rafallyczkowskiadylic: Good catch! I will fix shortly

src/parse.js
rafallyczkowskiadylic: @Connum it was intended, because actually this was not required to be supported within the iteration. I covered the planned scope which was not including this bit. If you are concerned, I think I might revisit this

rafallyczkowskiadylic:
@Connum once we have the other [MR](https://github.com/opentypejs/opentype.js/pull/491/files) merged (that covers `rlig` that is missing here as well) I will resolve and address all concerns around this mr shortly.

Connum:
Sure, that's fine! We don't have to support everything at once when implementing a new feature or format. We should just make sure that users are notified about any missing/unsupported functionality.

I just wanted to avoid that anything was misusing that was supposed to be in this PR already.

src/substitution.js
rafallyczkowskiadylic: @Connum this configuration param is used by the `Layout.getFeaturesLookups` that implements the ordered features lookup sequence. The thing is I have improved only Positioning features lookup (`Position.prototype.getPositionFeatures`) leaving substitutions untouched. So for the substitutions still uses the  "old features lookups parsing flow" which is as it was before. Adapting this ordered lookup sequence to the substitution will require a decent refactoring (and then this line will be "in use"). 

Hope this make sense?

rafallyczkowskiadylic: PS. and just to be on the same page, this array defines which lookup tables are supported for which features. This way we can have less redundancy and a single algorithm that parses the lookup tables. That why `Layout.getFeaturesLookups` got implemented.

We might get refactor **substitutions** with better tests coverage. Right now I have not taken this call because I had some concerns about the tests coverage not mature enough for such change to the substitutions (bidi)

rafallyczkowskiadylic:
@Connum Perfect! Great you are on top. So, let's address what is pending now. We stay in touch

src/parse.js
rafallyczkowskiadylic: Assuming this comment https://github.com/opentypejs/opentype.js/pull/557#issuecomment-1453904466 was referring to this?:) Please resolve thread if we are on the same page. Thanks.

src/font.js
```diff
@@ -279,7 +282,8 @@ Font.prototype.defaultRenderOptions = {
          * and shouldn't be turned off when rendering arabic text.
          */
         { script: 'arab', tags: ['init', 'medi', 'fina', 'rlig'] },
-        { script: 'latn', tags: ['liga', 'rlig'] }
+        { script: 'latn', tags: ['liga', 'rlig'] },
+        { script: 'DFLT', tags: ['mark'] },
```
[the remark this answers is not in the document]
rafallyczkowskiadylic: @ILOVEPIE I could not find this. I believe you might have referred to the langSysTable (tag) and this is a script tag? Moreover:
- when looking into specs,[ there are refs to the uppercased values only](https://learn.microsoft.com/en-us/typography/opentype/otspec170/scripttags) or [here](https://learn.microsoft.com/en-us/typography/opentype/spec/chapter2#script-list-table-and-script-record) (screen) 
- when verified with the font source definitions, found also uppercased (screen) 

image

src/layout.js
rafallyczkowskiadylic: Fixed.

src/parse.js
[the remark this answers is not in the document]
rafallyczkowskiadylic: Fixed. Pointer is moving now

rafallyczkowskiadylic:
I will check https://github.com/opentypejs/opentype.js/issues/501

rafallyczkowskiadylic:
#501 is fixed within this MR

rafallyczkowskiadylic:
@Connum @ILOVEPIE added few more commits with the substitution algorithm fixes and a new substitution format support. I haven't refactored the other working code (just the required bits). We can find few spots where we see "similar" functionalities. For example:

1. `src/features/featureQuery.js`  `FeatureQuery.prototype.lookupFeature` collects substitutions - too much responsibility here. I marked this as deprecated and pointed the fixed algorithm
2. `src/substitution.js` `Substitution.prototype.getFeature` - substitution processing
3. `src/bidi.js` - substitution applications 

I believe we could move towards:
- `Layout` - to keep what it has, consume GPOS/GSUB definitions, lookups. Keep important `getFeaturesLookups` method that build a union lookups list for ordered processing.
- `Position` and `Substitution` - shares base functionality to provide features lookups through the `Layout`
- `FeatureQuery` - should be promoted to use this to find and process features (not only in the bidi)
- substitution features per script i.e `src/features/latn/latinLigatures.js` `src/features/arab/arabicRequiredLigatures.js` should not be required since mostly they are simply the same. We basically need word ranges and context, but these are the same methods. These can be removed in favor of  `applySubstitutions.call(this, 'thai', ['liga', 'rlig', 'ccmp']);`

Not sure if `bidi` is really required, maybe for the directional substitutions but I think bidi could be deprecated and it's responsibility could be dropped into the `Font` where bidi is being called and where positioning logic is happening anyways. But basically I think we could use a  single processor  both for substitutions and positioning. 

Well this can open a discussion, but wanted to share some thoughts.

Connum:
I'm totally for unifying duplicate functionality. The way that feature processing is currently structured is something that has deterred me from working on new font features because it frankly seems a bit messy and not very intuitive to me. However, I would see that as a separate step from this PR. Let's get this support in first and consolidate feature processing in the future. Otherwise there might be too much code to review at once, and we have enough open and partly almost-finished PRs lying around as it is.

That's my opinion and not the final say, of course.

ILOVEPIE:
This is blocking quite a few PRs, we should probably get this merged sometime soon.

rafallyczkowskiadylic:
Thank you guys for the comments. Yes, most of suggestions/work has been done.  I have consolidated functionality as described introducing more generic functions to handle/fix specifications. I will resolve conflicts  soon.

Connum:
Fixes #501 according to https://github.com/opentypejs/opentype.js/issues/501#issuecomment-1543727128

@rafallyczkowskiadylic it's been a couple of weeks - just a friendly nudge. :)

Connum (approved):

src/layout.js
```diff
@@ -194,6 +195,61 @@ Layout.prototype = {
...
+     * Returns an ordered, union lookup tables for all requested features.
+     * This follows an ordered processing requirements (specs):
+     * > During text processing, it processes the lookups referenced by that feature in their lookup list order.
+     * > Note that an application may process lookups for multiple features simultaneously. In this case:
+     * > the list of lookups is the union of lookups referenced by all of those features, and these are all processed in their lookup list order.
+     *
+     * https://learn.microsoft.com/en-us/typography/opentype/otspec191alpha/chapter2#lookup-list-table
+     *
+     * @param {string[]} requestedFeatures
+     * @param {string} [script='DFLT']
```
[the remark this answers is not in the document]
Connum: according to the spec it's uppercase: https://learn.microsoft.com/en-us/typography/opentype/spec/scripttags

src/font.js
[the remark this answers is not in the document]
Connum: https://learn.microsoft.com/en-us/typography/opentype/spec/scripttags it's lowercase for all other tags, but DFLT is uppercase in the docs

Connum (dismissed):

Connum:
I was finally able to test this more thoroughly... I merged it into my experimental fork after resolving some conflicts.

The chars in the GPOS–3 test of the unicode test suite still rendered wrong, and they also didn't work in our demo page. 
image 
image


So I had to make a few changes to make it work, see
https://github.com/Connum/opentype.js-next/commit/95688c6f54ec55274cfe2eee86eb8493788ca632

image
image
Hey presto!

However, these calculations are quite expensive, even before my additions, because they are done for each single glyph. I don't think that's necessary and we should implement ways to at least cache this, or best only call the feature update when something has changed.

Connum:
Ok, now the tests for disabling the mark feature are failing of course, because I implemented it so that default features cannot be disabled. We'll have to talk about how we want to handle this.

Connum:
The more I look at this, the more confused I get... 😅

I think we still need some cleaning up before we can finally merge this, but I'm lost here without you @rafallyczkowskiadylic. I tried to fix https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#GSUB-1, but I got caught up in spaghetti code for several hours trying to follow the path of the substitution mechanism and got confused between old and new functionality. e.g. we should really get rid of this:
> - substitution features per script i.e `src/features/latn/latinLigatures.js` `src/features/arab/arabicRequiredLigatures.js` should not be required since mostly they are simply the same. We basically need word ranges and context, but these are the same methods. These can be removed in favor of  `applySubstitutions.call(this, 'thai', ['liga', 'rlig', 'ccmp']);`

Regarding this substitution case, I started wondering why we split into words in the first place? The space glyph being part of a substitution is perfectly fine, there's no rule against it to my knowledge.

I know that's the crux of open source, but I wouldn't be such a pest if it wasn't blocking the progress so much. A clear structure for substitution and positioning as well as supporting all the different formats is really crucial for implementing further features.

Connum:
By the way, trying to get rid of the latn-specific `applyLatinLigatures`, I replaced it with
```
function applyLatinFeatures() {
    checkGlyphIndexStatus.call(this);
    applySubstitutions.call(this, 'latn', ['liga', 'rlig', 'ccmp', 'calt']);
}
```
in analogy to `applyThaiFeatures()`, and called it in `Bidi.prototype.applyFeaturesToContexts`:
```
    if (this.checkContextReady('latinWord')) {
        applyLatinFeatures.call(this);
    }
```

But then all ligatures stopped working...

rafallyczkowskiadylic:
Hey @Connum  I am on it. Let me resolve conflicts and see what is on the code. I will revert back by the end of the weekend

rafallyczkowskiadylic:
@Connum Conflicts resolved. 

Pending items:

1. Could you share this `TestShapeEthi.ttf` file with me so I can verify the outcomes? 
2. I understand. I had similar issue, with some redundancy to get into the logic, that is why I didn't wanted to deprecate old code but rather leverage a new core lookup features. The most important functions is: Layout.getFeaturesLookups that implements common logic for fetching features, for substitutions and for positioning. 

> I started wondering why we split into words in the first place?

It was implemented like that, and the reason for this is that within single sentence there may be characters from a different language context. Example: `我的iPhone是最好的` where "iPhone" is for the latin context word. 

4. For the `applyLatinLigatures` issue, could you send me specific case you are testing? Including exact string, expected result, and source font file. I will check what may be happening because I see no issues with the latin fonts I am testing currently. 

Moreover, please keep in mind supported substitution formats.

```
/**
 * Supported substitutions
 */
const SUBSTITUTIONS = {
    11: singleSubstitutionFormat1,
    12: singleSubstitutionFormat2,
    63: chainingSubstitutionFormat3,
    41: ligatureSubstitutionFormat1, 
    21: multiSubstitutionFormat1
};
```

Different fonts' source files may have these definitions in a different formats that eventually may not be supported yet.

-----

@Connum if you could share all your use cases, with results you expect to see and source font files - I can review all of them and eventually address. 

-----

For the clean up I think we could just mark functions as deprecated for now. I'd refactor only bits covered with reliable tests so we can actually rely on outcomes before/after we change the code (this was also the reason I'd rather applied the code and "promoted" than removing or refactoring old functioning pieces).

rafallyczkowskiadylic:
Is there easy way to tests svg using demo page? I wrapped up simple compiler so I am using this a

`npm run font:preview -- --font="Roboto-Black.ttf" --content='Lorem ipsum dolor sit amet. ąłćżółźżćśćłóęą '`

that generates `preview.html` that compares HTML (browser) rendered content and SVG path. 

image

rafallyczkowskiadylic:
Ok. @Connum currently I do not have pending items, unless you share your failing test cases (font names, conetnt tested, etc) with me according [to this](https://github.com/opentypejs/opentype.js/pull/557#issuecomment-1837164457)

Connum:
I don't really like the idea of having both approaches in the code base in parallel. The arab and latn features should work with the new approach as well, so we should move everything to work with it. I understand if this is more than you're willing to do. Could you mark the PR as allowing maintainers to make changes? Then me or someone else could look into fixing this for all the scripts.

Connum:
Shouldn't I be able to change `applyLatinLigatures()` to
```
function applyLatinLigatures() {
    applySubstitutions.call(this, 'latn', ['liga']);
}
```
and it should just work? But the ligature tests then fail.

rafallyczkowskiadylic:
> Shouldn't I be able to change `applyLatinLigatures()` to
> 
> ```
> function applyLatinLigatures() {
>     applySubstitutions.call(this, 'latn', ['liga']);
> }
> ```
> 
> and it should just work? But the ligature tests then fail.

Actually I checked and found few tests failing for your cases, for `latn` they are related to manipulating font's definitions (`Substitution.prototype.addLigature`). These implementations are still using old lookupTables search, and manipulating local references (

In other word, addLigature is using `Layout.getLookupTables`, manipulates this method's output references (expecting the test will also use `.getLookupTables` method). Thing is my solution does not use `.getLookupTables` at all because of `getFeaturesLookups` 

```
Substitution.prototype.addLigature = function(feature, ligature, script, language) {
    const lookupTable = this.getLookupTables(script, language, feature, 4, true)[0];
    let subtable = lookupTable.subtables[0];
    ...
    subtable.ligatureSets = .... // <- manipulation
});
```

yet `Layout.getLookupTables` creates own copy of subtables refs (that is why we can't see them within new feature) 

image

If `.addLigrature` would instead manipulate definition on the `this.font.tables[this.tableName].lookups` level (not local references) everything will work as expected. This is how I would approach this, to deprecate any use of `getLookupTables` and if there are definitions updated I would try to make them as "high" as possible (ideally around specific table def source, `this.font.tables[this.tableName]....`)

PR #562, opened 2023-02-12
make cffEncoding.charset fall back to CIDs instead of undefined

### LINKED ISSUE #524: parseCFFCharset returns undefined array elements for CIDKeyed CFF fonts with SID -> CID mapping
CFF fonts don't always provide strings for each SID in the TopDICT.charset table. In this case the SID for a given GID is actually a CID. `parseCFFCharset` implicitly assumes that a string will exist for every SID in the charset table, which is incorrect. This will require minor changes to `cffEncoding.charToGlyphIndex` too.

## Expected Behavior
`parseCFFCharset` should return an array of strings *and* CIDs (depending on whether a string was found for the given SID)

## Current Behavior
`parseCFFCharset` returns an array with names from the relevant encoding (standard, expert, etc.) and `undefined` for other GIDs.

## Possible Solution
```js
 charset.push(getCFFString(strings, sid));
```
Should test the returned value for undefined:
```js
charset.push(getCFFString(strings, sid) || sid);
```

## Steps to Reproduce
1. Load a CIDKeyed CFF font like [Noto Sans HK](https://fonts.google.com/noto/specimen/Noto+Sans+HK?subset=chinese-hongkong)
2. Inspect `font.cffEncoding.charset`

Connum:
## Description

see #524

## Motivation and Context

Fixes #524

## How Has This Been Tested?

Loaded *FiraSansOT-Medium.otf* before and after: former `undefined` entries are now the corresponding CID instead.

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/cff.js
+++ b/src/tables/cff.js
@@ -421,14 +421,14 @@ function parseCFFCharset(data, start, nGlyphs, strings) {
     if (format === 0) {
         for (let i = 0; i < nGlyphs; i += 1) {
             sid = parser.parseSID();
-            charset.push(getCFFString(strings, sid));
+            charset.push(getCFFString(strings, sid) || sid);
         }
     } else if (format === 1) {
         while (charset.length <= nGlyphs) {
             sid = parser.parseSID();
             count = parser.parseCard8();
             for (let i = 0; i <= count; i += 1) {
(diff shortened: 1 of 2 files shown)


ILOVEPIE (approved):
Looks good to me.

PR #563, opened 2023-02-14
Gsub lookup 71 writing

### LINKED ISSUE #561: Add writing support for lookup table 7 (Extension Substitution Subtables)
PR #486 will add parsing support for the extension tables, but we'll still need writing support.

In detail we'll have to make sure:
1. That tables marked as format 7 will be written instead of throwing an error that they are unsupported
2. when writing tables (e.g. added manually/programmatically), we'll have to check whether any table offset would go beyond the 16-bit range and if so, add a lookup table 7 for that table

Connum:
## Description
THIS IS A DRAFT ONLY, SEE BELOW!

## Motivation and Context

Based on #486 adding read support, I thought I could implement writing support for LookupType 7 (Extension Substitution).
However it seems I'm way out of my comfort zone and stuck on finding the right approach to this.
What I've done so far is adding a subtableMaker for ExtensionSubstFormat1 tables and returning the extension as a new table to append after all the other subtables.

However, I'm not sure if the extensions are appended in the correct location ([the spec](https://learn.microsoft.com/en-us/typography/opentype/spec/gsub#lookuptype-7-extension-substitution) is not clear on where they are supposed to be located - after the LookupList? Or at the end of the GSUB table? -, or at least I don't understand it) and how to replace the offset value in the corresponding extension substitution table afterwards. I tried to do it similar to the subsitution table offsets, but I wasn't able to figure it out and gave up.

Maybe this is still a good starting point for someone more clever than me to implement this.


src/tables/gsub.js
```diff
@@ -402,6 +402,21 @@ subtableMakers[6] = function makeLookup6(subtable) {
     check.assert(false, 'lookup type 6 format must be 1, 2 or 3.');
 };
 
+subtableMakers[7] = function makeLookup7(extensionSubstitutionTable) {
```
comesuccingfuccsloot, line 405:
```JavaScript
import { encode } from '../types.js';

// ...

subtableMakers[7] = ({ substFormat, lookupType, extension }) => {
	check.assert(substFormat === 1, 'lookup type 7 format must be 1.');

	const innerTable = subtableMakers[lookupType](extension);

	return new table.Table('extensionSubstitutionTable', [
		{ name: 'substFormat', 			type: 'USHORT', 	value: 1 },
		{ name: 'extensionLookupType', 	type: 'USHORT', 	value: lookupType },
		{ name: 'extensionOffset', 		type: 'ULONG', 		value: 8 },
		{ name: 'extensionData', 		type: 'LITERAL', 	value: encode.TABLE(innerTable) },
	]);
};
```

PR #535, opened 2022-12-31
feat: Added a thai context to the bidi and substitution features

rafallyczkowskiadylic:
## Description

feat: Added a thai context to the bidi and substitution features. 
Followed the current solution's convention to introduce thai script's features within the bidi implementation.

- added the thai word context using a unicode characters range for the thai range
- added the support for thai `liga`, `rlig`, `ccmp` substitution features
- updated the font default render options for thai scripts features support
- tests coverage

## Motivation and Context

Thai fonts / context had no support for glyph substitutions features that are required to render a valid thai contents:
- [Ligatures (liga)](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ko#tag-liga)
- [Required Ligatures (rlig)](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt#tag-rlig)
- [Glyph Composition/Decomposition (ccmp) ](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae#tag-ccmp)

## How Has This Been Tested?

Added tests coverage to the `test/bidi.js` conducted agains the NotoSansThai-Medium font source ([SIL OFL](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL)) including: 
- should apply glyph composition feature
- should apply ligatures feature
- should apply required ligatures feature
- should match thai words in a given text (including mixed thai and english content)

## Screenshots (if appropriate):

image

## Thai font updated:

OFL font has been updated to add and test `rlig` feature as follows:
image (1)

with successful tests results WITH `rlig` support as:
image (2)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I did `npm run test` and all tests passed green (including code styling checks).
- [x] I have added tests to cover my changes.
- [x] My change requires a change to the documentation.
- [x] I have updated the **README** accordingly.
- [x] I have read the **CONTRIBUTING** document.

PR #564, opened 2023-02-14
Handle rlig for latin script

### LINKED ISSUE #479: Ligatures not rendering
## Expected Behavior
The font should be rendered with ligatures:
image

## Current Behavior
The font is rendered without ligatures, although they are being detected in the `gsub` table as far as I understand:

image

image

## Possible Solution
I don't know actually, I don't have much experience on the font file internals, but I'm open to working on the issue if someone is willing to guide me a little

1. Go to https://opentype.js.org/
2. Upload the TTF font [Pecita.zip](https://github.com/opentypejs/opentype.js/files/6703435/Pecita.zip)
3. Make sure ligatures are enabled
4. Write the word "quick" or any other word, the font is not rendered with ligatures

## Context
I have an app where users can write text on the browser using any font and export the text as SVG, it works great for most fonts, some of them have issues, converting them from TTF to OTF or vice-versa or re-exporting them with FontForge fixes most issues, in this case, I couldn't fix the issue by doing that in this case.

## Your Environment
* Version used: Opentype.js 1.3.1
* Font used: [Pecita.zip](https://github.com/opentypejs/opentype.js/files/6703435/Pecita.zip)
* Browser Name and version: Chrome Version 91.0.4472.114
* Operating System and version (desktop or mobile): Windows 10 21H1 x64

### LINKED ISSUE #554: Rlig should always be active, for all scripts and should not be able to be disabled.
## Expected Behavior
rlig stands for required ligature, liga is optional ligature

## Current Behavior
rlig is only on for arabic script.

## Possible Solution
Add rlig to all scripts or just hard-code it to be on by default.

see #479 for reproduction steps.

## Context
Some fonts use rlig instead of liga because they don't want the ligatures to be toggleable.

Connum:
## Description

This handles the `rlig` feature for latin scripts.

## Motivation and Context

Despite being enabled for latin scripts even by default, `rlig` ligatures were only handled for Arabic script.

I did some minor refactoring of the ligature application to diminish code duplication betweeen arabic and latin script ligatures and the new rlig implementation. They now use the same method in a new `src/features/commonFeatureUtils.js` file, without breaking the existing file/module structure, with parameters for the script, feature and range. This might come in handy as we progress to implement new font features. Speaking of which, we should probably overhaul feature handling as we do so. This PR only enables `rlig` for the latin script, which fixes #554 (and fixes #479), while other scripts might need it as well and there are other features that might not be script-specific.

## How Has This Been Tested?

Tested with the Pecita font provided in #479 and added corresponding tests for it (the font is under SIL Open Font License so I was able to use it directly for the tests), not breaking any other tests.

before:
image
after (note the "qu" and "ck" shapes):
image
The rest of the differences to how it should look according to #479 can be attributed to not supporting the `calt` feature.

## Types of changes

## Checklist:


ILOVEPIE:
This might conflict with #535 and #557. Which should we merge first, one of those two PRs reworks the feature system (can't remember which one).

ILOVEPIE:
@Connum do we want to merge this first or rework it once #535 and #557 are merged?

Connum:
Good question... Both of them made changes to bidi.js which I also altered, and both of them could probably benefit from the common helper function that I introduced. I think it's easier to merge those two first and have me adapt the rlig code to those changes instead of heaving to incorporate mine into both of the PRs.

Connum:
up-to-date with current master, blocked by #535 and #557

src/features/commonFeatureUtils.js
```diff
@@ -0,0 +1,37 @@
+import { ContextParams } from '../tokenizer';
```
Connum, line 1: I just noticed that the file extensions are missing, but are now required since we changed the build process and the library may be used in module form on the web, where we need the file extensions. I'm fixing it right now.

Connum:
I somehow managed to mess up this branch... Going to convert to draft until I find time to fix this.

ILOVEPIE:
@Connum can we get this updated?

Issue #326, opened 2018-03-05
Fix test cases for CMAP, GPOS and GSUB


brawer:
Thanks to @fpirsch’s fix in #324, it’s again possible to run the [Unicode test suite for text rendering engines](https://github.com/unicode-org/text-rendering-tests) on OpenType.js.

## Expected Behavior
All test cases should pass.

## Current Behavior
Currently, OpenType.js is failing 62 test cases. See [test report](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html).

## Possible Solution
Consider fixing at least [CMAP-1](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#CMAP-1), [CMAP-2](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html#CMAP-2), and all the GPOS and GSUB cases. For inspiration, have a look at [fontkit](https://github.com/devongovett/fontkit), another OpenType implementation in JavaScript, which is passing most tests in the test suite (see [test report](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/fontkit.html)).

## Steps to Reproduce
See [README for the test suite](https://github.com/unicode-org/text-rendering-tests/blob/master/README.md). Until OpenType.js cuts its next release, you’ll need to manually patch @fpirsch’s fix from #324 into `node_modules/opentype.js/bin/test-render`.

Jolg42:
Hi @brawer Thanks for reporting that!

I merged PR #324 but a new release on npm is still needed I guess?

fpirsch:
@Jolg42 last release was several months ago, a whole lot of bugs have been fixed since. It would be nice to have a new release :-)

Jolg42:
@fpirsch @brawer 0.8.0 is out!

brawer:
Updated test report is [here](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html). Happy debugging!

fpirsch:
Actually `bin/test-render` has its own text rendering function which doesn't handle any font feature. It uses opentype.js only to get individual glyph SVG paths. This is why most tests fail.

brawer:
[...]

PR #568, opened 2023-02-18
Apply SVG path optimizations, create Path from SVG

### LINKED ISSUE #369: Path.toPathData and Path.toSVG - X Axis is flipped
I'm using OpentypeJS to create SVGs for single glyph characters. However, when I draw the SVGs, the glyph is flipped on it's X axis ( upside down ). Even the ./bin/test-render in the repo generates flipped SVG path data.

## Expected Behavior

1) Load font
2) Get glyph path data with glyph.path.toSVG()
3) Use the path data to render proper glyph

## Current Behavior

1) Load font
2) Get glyph path data with glyph.path.toSVG()
3) Use the path data to render glyph, the glyph is rendered upside down.

## Possible Solution

Fix path commands. 

Just run the /bin/test-render  from the repo

## Your Environment

* Version used:  0.11.0
* Font used: Every font

Connum:
## Description

I'm working on some SVG optimizations that will lead to more [unicode-org/text-rendering-tests](https://github.com/unicode-org/text-rendering-tests/issues) tests passing.
This will also add the capability to make a Path() from an SVG path string or element.

Still WIP as I need to test some more and write actual tests, but input is already welcome!

### Todo
- [x] change boolean optimize option to options object
- [x] provide option to flip axis on SVG output (fix #369)
- [X] don't flip y axis in bin script for test report
- [X] provide option to flip axis for reading svg
- [x] Take a look at #326

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
Will break code that relies on SVG data being upside down, but that behaviour was wrong in the first place

## Checklist:


Connum:
Just with some SVG path optimizations (removing duplicate/unnecessary commands), we now pass 56 of the unicode test cases instead of 37 with the current master: https://connum.github.io/opentypejs-reports/reports/test-connum_svg-optimize-099bc17ecfe9030edcd1530b2792063cb7981c56.html
The optimizations also produce smaller path sizes, so I decided it makes sense to implent these by default.

Connum:
I improved the SVG rendering of the test script, so that glyph substitution and positioning should now work exactly like when drawing on a canvas.

Now we pass 69 Tests, and all tests still failing should really be cases that we don't support or handle incorrectly right now (Note that sometimes the paths are just off by a few pixels, so they look like they should pass, but are still actually incorrect):
[Connum/opentype.js:optimize-svg#7e4de2c](https://connum.github.io/opentypejs-reports/reports/test-connum_svg-optimize-7e4de2c2ce9e822161b61349cf2a7ec55c7fd05e.html) (passing: 69, failing: 676)

Connum:
Superseded by #569 (after I completely messed up the commit history trying to clear it up...)

PR #569, opened 2023-02-21
SVG-related improvements and fixes

### LINKED ISSUE #369: Path.toPathData and Path.toSVG - X Axis is flipped

### LINKED ISSUE #526: Support using a plain SVG file as a glyph
## Expected Behavior

I don't know if this would be considered out of scope or not, but I'd really like if this library supported importing a raw SVG string as a glyph, which would cut out all sorts of intermediary dependencies one currently needs when generating an .odt font.

## Current Behavior

It doesn't seem possible to import raw SVG.

## Possible Solution

Maybe the Glyph class could get an additional `svg` property that internally extracts the path from the SVG?

Connum:
This PR supersedes draft #568

## Description

* implements loading of SVG path data into Path(), including basic modification like offset and scaling (fix #526)
* fixes SVG paths being flipped upside-down (fix #369)
* Helps the library pass more [unicode-org/text-rendering-tests](https://github.com/unicode-org/text-rendering-tests/issues) tests (see #326)
  * implements path optimizations for shorter/smaller path data *
  * improves the SVG rendering in the test-render script *

## Motivation and Context

The library failed several tests in the unicode test tool, despite technically supporting the underlying features.
Exporting SVG paths was possible, but the paths were flipped upside down due to inverted Y-axes in SVG vs font paths. Importing SVG paths was not supported.

## How Has This Been Tested?

We now pass 69 tests instead of 37 in the unicode test tool, see report: https://connum.github.io/opentypejs-reports/reports/test-connum_svg-optimize-7e4de2c2ce9e822161b61349cf2a7ec55c7fd05e.html
We didn't have any tests for path.js so I added several to test the basic and new features.
I played around with loading, exporting and modifying SVG paths in existing fonts.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
Will break code that relies on SVG data being upside down, but that behaviour was wrong in the first place

## Checklist:


src/path.js
```diff
@@ -16,6 +16,283 @@ function Path() {
...
+    }
+    commands = [].concat.apply([], subpaths); // flatten again
+    return commands;
+}
+
+/**
+ * Returns options merged with the default options for parsing SVG data
+ * @param {object} options (optional)
+ */
+function defaultSVGParsingOptions(options) {
```
ILOVEPIE, line 65: wouldn't it be better to call this createSVGParsingOptions, because if we're overriding the defaults it's not default.

src/path.js
```diff
@@ -16,6 +16,283 @@ function Path() {
...
+    };
+    const newOptions = Object.assign({}, defaultOptions, options);
+    return newOptions;
+}
+
+/**
+ * Returns options merged with the default options for outputting SVG data
+ * @param {object} options (optional)
+ */
+function defaultSVGOutputOptions(options) {
```
ILOVEPIE, line 82: Same here.

src/path.js
```diff
@@ -16,6 +16,283 @@ function Path() {
...
+        });
+    }
+
+    function makeRelative(buffer) {
+        if (!this.commands.length) {
+            return buffer;
+        }
+        const lastCommand = this.commands[this.commands.length - 1];
+        for (let i = 0; i < buffer.length; i++) {
+            buffer[i] += lastCommand[i % 2 ? 'y' : 'x'];
```
ILOVEPIE, line 138: Just a small optimization, ``x % 2`` can always be rewritten as ``x & 1`` which runs faster, not that big of a deal, but processors are a lot slower at modulus (due to the division).

ILOVEPIE (changes requested):
I'd like to see someone else's opinion on this, the changes I've suggested are relatively minor, this is however a breaking change PR, so we should probably lump it into 2.0.0

src/path.js
Connum: True, I'll change that!

src/path.js
Connum: Interesting! That's good to know. We could probably optimize performance further by using a generator function and yield, but our current build process doesn't transpile that. I'll add that as a todo comment.

Connum:
> this is however a breaking change PR, so we should probably lump it into 2.0.0
I was hesitant whether to mark it as breaking or not. Having upside down output is obviously not what anyone would expect, on the other hand it's been like that for years, so there's probably code out there that relies on it and flips it itself.

src/path.js
```diff
@@ -16,6 +16,285 @@ function Path() {
...
+            throw new Error('Unexpected character: ' + token + ' at offset ' + i);
+        }
+    }
+    applyCommand.apply(this);
+
+    if (options.optimize) {
+        this.commands = optimizeCommands(this.commands);
+    }
+
+    let flipY = options.flipY;
```
yne, line 268: is this variable a number or a boolean ?

src/path.js
```diff
@@ -16,6 +16,285 @@ function Path() {
...
+
+    return this;
+};
+
+/**
+ * Generates a new Path() from an SVG path element or path notation
+ * @param  {string|SVGPathElement}
+ * @param  {object}
+ */
+Path.fromSVG = function(path, options) {
```
yne, line 293: Don't offer much gain, not sure about it usefulness

src/path.js
Connum: See options documentation: "whether to flip the Y axis of the path data, because SVG and font paths use inverted Y axes. (true: calculate from bounding box, false: disable, integer/float: a base value for the calculation. default: true)"

It can be useful to calculate the actual base value from the font's ascender and descender values or the viewbox of an SVG you want to add the generated path to.

src/path.js
Connum: Well, we use such shorthands in other places of the codebase too, and I liked the shorter notation without having to create an empty Path first.

src/path.js
```diff
@@ -16,6 +16,285 @@ function Path() {
...
+function createSVGOutputOptions(options) {
+    // accept number for backwards compatibility
+    // and in that case set flipY to false
+    if (parseInt(options) === options) {
+        options = { decimalPlaces: options, flipY: false };
+    }
+    const defaultOptions = {
+        decimalPlaces: 2,
+        optimize: true,
+        flipY: true
```
yne, line 91: we shall keep it `false`, and flip it back to `true` in 2.0.0
so we could merge this PR now

Connum: Is there a plan to release a 1.x version before 2.0.0? Because we already have the breaking change regarding the platform-agnostic `font.table.name` values merged in our main branch

src/path.js
```diff
@@ -230,17 +509,18 @@ Path.prototype.draw = function(ctx) {
...
  * See http://www.w3.org/TR/SVG/paths.html#PathData
- * @param  {number} [decimalPlaces=2] - The amount of decimal places for floating-point values
+ * @param  {object|number} [options={decimalPlaces:2, optimize:true}] - Options object (or amount of decimal places for floating-point values for backwards compatibility)
  * @return {string}
  */
-Path.prototype.toPathData = function(decimalPlaces) {
-    decimalPlaces = decimalPlaces !== undefined ? decimalPlaces : 2;
+Path.prototype.toPathData = function(options) {
+    // set/merge default options
+    options = createSVGOutputOptions(options);
```
yne, line 517: no need for a separate function, I think this wold also work
```js
Path.prototype.toPathData = function(options = 2) {
  const def = { decimalPlaces: 2, optimize: true, flipY: true};
  options = Object.assign(def, isNaN(options) ? options : {decimalPlaces:options});
}
```

Connum: My code was structured differently at first, and I wanted to be able to use the options in the glyph.js wrapper functions as well, in order to pass the base value for flipY from the glyph metrics and already have the default values there. But I didn't want to make things more complicated with this PR as they already are. We might need that in the future though, that's why I left them in a function.

src/path.js
yne: I was referring to the left side `let flipY`

src/path.js
yne: We need to clarify this road map (with @fdb ?)
If the next versions is going to be 2.0.0(-alpha1) then no problem for me

yne (approved):
Very nice addition, but we shall clarify ASAP if the next release is going to be 2.0.0, in which case some function argument could be completely migrated (number => object).

src/path.js
Connum: It just takes on the value of the corresponding option. I can rewrite it if that's considered bad practice?

yne: I feel a litle bit confused when I see
```js
let flipY = options.flipY // boolean
```
then
```js
flipY = boundingBox.y1 + boundingBox.y2 // number ?!
```
but
```js
flipY !== false // boolean
```
and
```js
flipY - cmd[prop] // number again
```
I think there is 2 usecase for the same variable
the "offset" (number) version would be better in a separate variable I think.

src/path.js
Connum: Chromium browsers seem to do some optimizations that make the difference marginal (and sometimes modulus runs even a tiny bit better than bitwise), but the story is different in Firefox. I had differences of up to ~15%:

https://jsbench.me/2zlehfyyi2/1

image

yne: I confirm that, on chromium-based browser get almost identical result for both `%` and `&`.

The PR is (re)open so you can merge it (no issue for me :+1: ).

Let's just not forget that `premature optimization is the root of all evil` ;)

src/path.js
Connum: Makes sense that it might be confusing. Changed it with a new commit.

Connum: Woops, pushed to the wrong repo, but now it's in.

src/path.js
```diff
@@ -16,6 +16,286 @@ function Path() {
...
+
+    if (options.optimize) {
+        this.commands = optimizeCommands(this.commands);
+    }
+
+    let flipY = options.flipY;
+    const doFlipY = !!flipY || flipY === 0;
+    if (options.flipY === true) {
+        const boundingBox = this.getBoundingBox();
+        flipY = boundingBox.y1 + boundingBox.y2;
```
yne, line 272: were is `flipY` declared ?

src/path.js
yne: I still don't get why does we mixup in the same option/variable
- the offset
- the enabling
I would have used `{flipY:false}` `{flipY:true, flipOffsetY: 0.2}` if I get your algorithm correctly.

Connum: So we *are* talking about the option itself, not just the variable?
I think it's just simpler to have one option for this. Either you disable it, have it calculated automatically, or you provide a value. You would never set `{flipY: false, flipOffsetY: 123}`, so in my mind it didn't make sense to have this constellation.
@ILOVEPIE what do you think? In the end, it doesn't really matter to me, so I'll go with the majority.

Connum: split into `flipY` and `flipYBase` now, upated README accordingly

Connum:
up-to-date with the current master and ready for re-review

ILOVEPIE:
> I was hesitant whether to mark it as breaking or not. Having upside down output is obviously not what anyone would expect, on the other hand it's been like that for years, so there's probably code out there that relies on it and flips it itself.

This is a breaking change as it changes some public facing apis, not just wither the SVG is upside down.

Connum:
You can still pass a number instead of the options object and it will be handled accordingly, or am I missing something else?
We also have the platform agnostic names table change in the master already and had the version field bumped to 2.0.0 before it was removed with the overhauled build/release system. So I think we should be able to merge this?
If we really need a hotfix 1.x release before 2.0.0 is ready, we can still branch it off from an earlier point before all the new features that were added recently.

yne (approved):

yne:
@Connum 
I rebased-squashed your branch so it shall be mergable now,
please avoid merging master into your branch as it make the rebase more painful.

Connum:
Can we please talk about the roadmap and how to proceed with this? I'm not that keen on resolving endless conflicts should this get merged after other PRs with big changes. 🙂 Also this SVG fix is very helpful when testing new features/enhancements/fixes with the Unicode test suite.

It was my understanding that we were going to release 2.0.0 next anyway (though I have to admit I'm not sure what that feeling is based on, as we never explicitly talked about it before) but if that's not the plan, we could at least have a dedicated 2.0.0 branch? Then I'll also scrap the partial backwards compatibility.

yne:
@ILOVEPIE can we process?

src/path.js
ILOVEPIE: @yne the reason why it's the same on chrome is that chrome rewrites the code to actually be ``x & 1`` when it sees ``x % 2``.

src/path.js
```diff
@@ -16,6 +16,290 @@ function Path() {
     this.strokeWidth = 1;
 }
 
+function roundDecimal(float, places) {
+    return +(Math.round(float + 'e+' + places) + 'e-' + places);
```
ILOVEPIE, line 20: The type conversions here may slow this down. Might I suggest
```js
const shift = Math.pow(10,places);
return (Math.round(float*shift)/shift);
```
instead.

ILOVEPIE (approved):
There is one comment I had regarding performance/readability, but it looks good.

src/path.js
Connum: That seems to have introduced a rounding issue

Connum: I'll take a closer look at this and run the unicode test suite to see what impact it has. Maybe it could even make things better in some cases. We'll see.

yne: From what I see, the `roundDecimal` is used at 2 places, for different reasons: 
- at path parsing, where all precision shall be kept (?)
- at serialisation, where it shall treated number as (truncated) String, not a rounded Number (which is sometime impossible due to floating point precision)

Connum: So when I adapt the test case, we don't fail any other tests and the number of passed tests in the Unicode test suite doesn't change, but I really think `M0.575` should change to `M0.58` when changing the `decimalPlaces` option from 3 to 3, and not `M0.57`. Why would we ever want to cut off decimal places instead of actual rounding?
I think that's why I chose to do it the way I did in the first place, because I always ended up with some weird rounding errors. But I'm open to a solution that feels less hacky and still produces correct values.

Connum: for now, I reversed the change and rebased on master

ILOVEPIE: That should be functionally equivalent to the rounding function you had originally as they do the exact same thing.

Connum: I don't know why the test is failing with a wrong value then, either 🤔

Connum: I was able to get around that rounding issue by adding `Number.EPSILON` before multiplying
```js
const shift = Math.pow(10, places);
return (Math.round((float + Number.EPSILON) * shift) / shift);
```
which makes all tests pass again - but when I tried with different values, `10.075` rounds to `10.07` instead of `10.08`.
My original implementation was the only one so far that seems to work for all cases. I'd say we leave it like that before introducing precision issues.

PR #577, opened 2023-02-26
fix: add missing `type` field to `package.json`

FractalHQ:
## Description
added `"type": "module"` to `package.json`

I'm not 100% certain this won't cause problems with legacy CJS builds / pipelines because I've always used ESM exclusively, but out of the dozens of dependencies in my project, this is the only one that causes this error.  I'm hoping to at least open a discussion about this with some of you who are very experienced with both CJS and ESM so I can learn more about how to navigate / assist in this predicament.

## Motivation and Context
This fixes the following errors when consuming `opentype`:
```bash
(node:4803) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
..../node_modules/.pnpm/opentype.js@1.3.4/node_modules/opentype.js/dist/opentype.module.js:14458
export default opentype;
^^^^^^

SyntaxError: Unexpected token 'export'
    at compileFunction (<anonymous>)
```

## How Has This Been Tested?
No additional tests.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "type": "git",
     "url": "git://github.com/opentypejs/opentype.js.git"
   },
+  "type": "module",
   "main": "./dist/opentype.js",
   "browser": "./dist/opentype.js",
   "module": "./dist/opentype.module.js",


ILOVEPIE:
Type field is automatically generated by the build process. If you wish to use opentype.js as an esmodule, please use the opentype.module.js  file or the opentype.mjs file

FractalHQ:
Ah that makes sense.  It seems the real issue is the library I'm using that uses this as a dependency is using the .module.js instead of the .mjs, resulting in the error.  I'll open an issue there.  Thanks!

PR #581, opened 2023-03-02
Parse cmap table format 14 and handle Unicode Variation Sequences (UVSes)

Connum:
## Description

This PR implements parsing of cmap table format 14 and applying Unicode Variation Sequences (UVSes) during rendering.
Writing is not yet implemented and will be added later in a separate PR.

## Motivation and Context

This will get rid of the `.null` glyph being displayed for variation selectors in text and swap the base character to the desired variation if supported by the font. We'll pass tests CMAP-1 and CMAP-2 of the Unicode test report tool as a result.

## How Has This Been Tested?

Added tests, tested with fonts, ran the Unicode test report (with #568 implemented in a test branch).

image
image

## Types of changes

## Checklist:


ILOVEPIE:
I'll review it after work today

src/bidi.js
```diff
@@ -178,6 +182,13 @@ function applyLatinLigatures() {
     });
 }
 
+function applyUnicodeVariationSequences() {
+    const ranges = this.tokenizer.getContextRanges('unicodeVariationSequence');
+    ranges.forEach(range => {
```
ILOVEPIE, line 187: forEach can be slow due to extra function calls, might be a better idea to use a for-in loop.

src/features/unicode/contextCheck/variationSequenceCheck.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+function isVariationSequenceSelector(char) {
+    if (char === null) return false;
+    const charCode = char.codePointAt(0);
+    return (
+        // Mongolian Variation Selectors
+        (charCode >= 0x180B && charCode <= 0x180D) ||
+        // Generic Variation Selectors
+        (charCode >= 0xFE00 && charCode <= 0xFE0F) ||
```
ILOVEPIE, line 12: Are we accounting for variation selectors on emoji?

src/parse.js
```diff
@@ -130,6 +135,13 @@ Parser.prototype.parseF2Dot14 = function() {
     return v;
 };
 
+
+Parser.prototype.parseUInt24 = function() {
+    const v = getUInt24(this.data, this.offset + this.relativeOffset);
```
ILOVEPIE, line 140: There's another PR that implements this function, I can't remember which one but that PR will need to be refactored if this is merged.

src/tables/cmap.js
```diff
@@ -78,25 +78,95 @@ function parseCmapTableFormat4(cmap, p, data, start, offset) {
...
     let offset = -1;
     for (let i = cmap.numTables - 1; i >= 0; i -= 1) {
         const platformId = parse.getUShort(data, start + 4 + (i * 8));
         const encodingId = parse.getUShort(data, start + 4 + (i * 8) + 2);
         if ((platformId === 3 && (encodingId === 0 || encodingId === 1 || encodingId === 10)) ||
             (platformId === 0 && (encodingId === 0 || encodingId === 1 || encodingId === 2 || encodingId === 3 || encodingId === 4))) {
             offset = parse.getULong(data, start + 4 + (i * 8) + 4);
-            break;
+            // allow for early break
+            if (format14Parser) {
```
ILOVEPIE, line 160: Did we mean for this to only break if format14 parser is present, because this changes the behavior from before...

src/tables/cmap.js
```diff
@@ -116,6 +186,11 @@ function parseCmapTable(data, start) {
         throw new Error('Only format 4 and 12 cmap tables are supported (found format ' + cmap.format + ').');
```
ILOVEPIE, line 186: This should read ``4, 12 and 14``

ILOVEPIE (commented):
I've written a few comments/questions, once those are discussed I will do an approving review. They aren't necessarily show-stoppers so I'm not going to request changes, but we should discuss them.

src/tables/cmap.js
Connum: Yes, because all other cmap table formats are exclusive, we were able to break as soon as we encountered the first one that we support. But now we don't know whether a format 14 table might follow, so we have to proceed with the loop. But if we encounter a supported format and already have a table with format 14, we can indeed break to save following interations.

src/parse.js
Connum: ok, whatever gets merged first, the other one will have to remove it

src/features/unicode/contextCheck/variationSequenceCheck.js
Connum: All combinations in https://www.unicode.org/Public/UCD/latest/ucd/emoji/emoji-variation-sequences.txt use the variation selector FE0E, so they are included here. UVSes only handle combinations of one base character and a variation selector. More complex emoji variations using zero-width non-joiner etc. are something different and not covered by this.

ILOVEPIE: from wikipedia:

They are currently used to specify standardized variation sequences for mathematical symbols, [emoji](https://en.wikipedia.org/wiki/Emoji) symbols, ['Phags-pa](https://en.wikipedia.org/wiki/%27Phags-pa_script) letters, and CJK unified ideographs corresponding to [CJK compatibility ideographs](https://en.wikipedia.org/wiki/CJK_compatibility_ideographs). At present only standardized variation sequences with VS1, VS2, VS3, VS15 and VS16 have been defined; VS15 and VS16 are reserved to request that a character should be displayed as text or as an [emoji](https://en.wikipedia.org/wiki/Emoji) respectively.

ILOVEPIE: or am i reading something different

src/tables/cmap.js
Connum: fixed

src/features/unicode/contextCheck/variationSequenceCheck.js
Connum: We'll support all current and future selectors in the three ranges according to
https://www.unicode.org/glossary/#variation_selector and https://freetype.org/freetype2/docs/reference/ft2-glyph_variants.html  But maybe I just don't understand what you're getting at? 😄

src/bidi.js
yne: I personally prefers a more readable code (i.e. forEach) since we can't predict what optimization will appear in the upcoming JS engines...

yne (approved):

src/bidi.js
yne: @Connum how many items are expected to be returned (on average) by this this.tokenizer.getContextRanges('unicodeVariationSequence') ?
if more than 100 then I do agree that we should use for() loop.
if lower than 100 then a readable code is more important.

Connum: Unfortunately, I don't have any empirical values, but from the extent of the lists, I would say that quite a lot could come together if a font were to implement all or at least many of them. But I don't know how likely or common that is. I have no problem changing this to a for loop, in fact I use for loops for opentype.js ever since @ILOVEPIE commeted that, despite using forEach loops everywhere else because I seldomly have real performance concerns when I'm mainly doing trivial web stuff.

Connum: On the other hand, I copied those lines from another feature handler, so we should probably either change it to for() everywhere, or use forEach() everywhere, especially when we should strave to reduce code duplication in favor of common helper methods to handle se same thing with just a few different parameters.

Connum:
rebased on current master

src/bidi.js
Connum: OK, so I wanted to know exactly and I created a benchmark. And I would not have thought that the difference is so significant, even with very few iterations.

Even with one element we are at 2-3%:
https://jsbench.me/rjlet6fft9/1

10 elements - >11%:
https://jsbench.me/rjlet6fft9/2

100 elements - >26%:
https://jsbench.me/rjlet6fft9/3

1000 elements - 91.5% (!):
https://jsbench.me/rjlet6fft9/4

While I do agree that it's a tradeoff with readability/clarity, I also do think that in a library like this we should put more emphasis on performance. So we should probably replace all uses of forEach() with for() - but I would tackle that in a separate PR, and not let it hold this and #535 back, what do you think?

We should also add it to our contribution guidelines (by the way, the PR template references a CONTRIBUTING file, but it seems to have been removed - what happened?) and probably even implement a linting rule so forEach doesn't sneak in in the future. Yes, there may be optimizations in browsers/engines in the future, but as it is now, I think the difference is crucial enough to ban forEach() for the time being.

src/features/unicode/contextCheck/variationSequenceCheck.js
ILOVEPIE: OK, nevermind was just making sure.

src/bidi.js
ILOVEPIE: Yeah we can do that as a separate PR, all modern processors do branch prediction/speculative execution to try to counteract how slow function calls are, they're one of the slowest operations the CPU can do.

ILOVEPIE: I'm not saying you should write everything in 1 function, but you should try to avoid doing unnecessary function calls in a loop.

ILOVEPIE: This is also the reason why compilers will sometimes opt to inline a function call.

Connum: OK, let's definitely do that then once these PRs are through.

ILOVEPIE (approved):
Looks good to me.

PR #576, opened 2023-02-24
Parsing and writing support for variable font tables fvar, STAT and avar

Connum:
## Description

- [X] add missing `encode.F2DOT14`
- [X] add `encode.FLOAT` as a wrapper to `encode.FIXED` when handling actual float numbers
- [X] adapt `fvar` writing to new platform-agnostic names structure (see #542)
- [X] fix `fvar` table not being written (fix #573)
- [X] add parsing and writing support for `STAT`
- [X] add parsing and writing support for `avar`
- [X] add stub for gvar parsing and writing

## Motivation and Context

This PR lays the groundwork for variable fonts support. Actual coordinate scaling and normalization (see https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#CSN) for rendering is not yet implemented, but several tables needed for supporting it can now be parsed and written.

I decided against also implementing `gvar`, because it introduces a set of new concepts/data types and seems to be the most complex of them. That would have bloated the PR even more.

### Further resources for future rendering implementation:
* arrowtype/blog/issues/6
* https://www.axis-praxis.org/samsa/

## How Has This Been Tested?

Loading and saving different variable fonts, added tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I did `npm run test` and all tests passed green (including code styling checks).
- [X] I have added tests to cover my changes.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the **README** accordingly.
- [X] I have read the **CONTRIBUTING** document.

PR #584, opened 2023-03-06
added local rule for banning forEach(), replace forEach() with for()

### LINKED ISSUE #583: Ban forEach, use for loops instead.
See https://github.com/opentypejs/opentype.js/pull/581#discussion_r1125017810

Connum:
## Description

Fix #583 by introducing a new local eslint rule that will throw an error when a call to `.forEach()` is encountered.
Also adds another local eslint rule to enforce file extensions in import paths to ensure the library working as a module in browser context (see https://github.com/opentypejs/opentype.js/pull/576#issuecomment-1453715852).

## Motivation and Context

As per the [discussion here ](https://github.com/opentypejs/opentype.js/pull/581#discussion_r1125145076), `.forEach()` has a significantly worse performance than a `for()` loop. As parsing and rendering should be as fast as possible, every little bit of performance boost we can get is good, as it can quickly add up with larger fonts or larger texts to render.

Also, we need to make sure that we explicitly inclue the file extension in import paths to ensure the library working as a module in browser context.

## How Has This Been Tested?

All tests still pass after changing all instances of `.forEach()` to `for()` equivalents. Linting will output an error if `forEach()` is encountered in the code.

## Types of changes

- [X] performance improvement

## Checklist:


ILOVEPIE:
Before merging this, we should compare the performance with and without the changes.

yne:
very good initiative @Connum.
we shall keep this benchmarker in the /docs/ so anybody could test the parsing performance.

I was about to say that the jsbench example only show the foreach overhead. in real life the body of the for would be much more large and the performance delta much lower.

again, i'm not a big fan of enforcing this kind of "optimization" rules because
- it does not appear often anyway
- it is only relevant for the current JS engine optimization
- it prevent acceptable cases (ie small list)

A programming guideline would have been more appropriate.

at best, this foreach->for transpilation shall be handled at bundling stage. but again, that's a lot of efforts/enforcement for a low benefit.

Connum:
I get your point of view, but regarding

> it does not appear often anyway

while there may not be many instances in the code, the loops can be called for many glyphs and many different features/lookups, so they may end up being executed many times, which will add up in the end

yne:
what I mean is that there are many way to write sub-efficient or unreadable code.

PR review is the way to prevent those issues (according to a guideline) so I wouldn't spend time on ESlint rules for every possible cases (.map .filter ...)

PR #586, opened 2023-03-08
add glyph character to the glyph inspector for easier copying

manylegged:
## Motivation and Context

I'm trying to copy glyphs from an icon font into source code. Displaying the glyph directly in the web page text makes it easy to load up an icon font and copy the characters I need.

## How Has This Been Tested?

I copied glyph-inspector.html and opentype.js onto my own server.

Screenshot 2023-03-08 115810

## Types of changes

## Checklist:


docs/glyph-inspector.html
```diff
@@ -130,6 +130,7 @@ <h1>Free Software</h1>
     var glyph = font.glyphs.get(glyphIndex),
         html = '<dl>';
     html += '<dt>name</dt><dd>'+glyph.name+'</dd>';
+    html += '<dt>glyph</dt><dd>'+String.fromCodePoint.apply(null, glyph.unicodes) + '</dd>';
```
Connum, line 133: When there are multiple code points for a glyph, all characters are just concatenated. A comma separated list would be preferrable. Also "glyph" is not really the correct term as a label here, because it's not the glyph we're showing, but readable characters for codepoints that are matched to that glyph.

Connum (changes requested):
Please see my comment! Other than that, good addition!
After your changes, please rebase on the current master and request re-review.

Connum:
Hey @manylegged, I would like to get this merged to reduce the number of open PRs. So if you find the time, it would be great if you had a look at my comment.

PR #591, opened 2023-04-01
Add WOFF2 loading support to demo pages

yne:
- modernise the pages to use `<script type=module>` version of opentype.js (better loading time, better lighthouse score, better page-ranking etc...)
- Add [wawoff2](https://github.com/fontello/wawoff2) "on-demand" loading for WOFF2 support on all demo pages and in the README

image


yne:
@Connum @ILOVEPIE : I think I'll merge it without waiting for review since it does not touch opentype.js cobase, only the HTML docs.

PR #596, opened 2023-04-16
added support for friendly stylistic set names.

moontypespace:
## Description
Added featureParams to get friendly stylistic set name via given nameID if given.
For reference, see OT spec: https://docs.microsoft.com/en-us/typography/opentype/spec/features_pt#tag-ss01---ss20

## Motivation and Context
It's possible to give stylistc set features name name via ID. They are stored in the name table. This was not implemented into opentype.js, yet.

## How Has This Been Tested?
I added unittests.

## Types of changes

## Checklist:

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -600,7 +600,9 @@ Parser.prototype.parseFeatureList = function() {
     return this.parsePointer(Parser.recordList({
         tag: Parser.tag,
         feature: Parser.pointer({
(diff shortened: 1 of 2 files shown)


Connum:
This will break the writing of the value in https://github.com/opentypejs/opentype.js/blob/master/src/table.js#L170 - could you implement that as well?

Connum:
I just biticed

PR #610, opened 2023-08-03
Delete cubic/quadratic interpolation in BoundingBox class

fatih-erikli:
Any cubic/quadratic curve will always be contained inside of the convex hull of their control points. Adding the control points to the bounding box will be enough.


fatih-erikli:
My patch will speed up the cubic/bezier curve generation when generating SVG. I closed the pr, there may be some cases when it's needed. I will keep it in my fork.

Issue #117, opened 2015-05-17
Not rendering OTF correctly


kuckir:
Hi, I have this font file http://www.ivank.net/AbrilFatface-Regular.otf and opentype.js does not render characters from Latin Extended-A group. However, charToGlyph() returns a glyph.

Here is an image: http://www.ivank.net/abril.png

kuckir:
Hi, can somebody take a look at it? I really need some library to render OTF fonts.

photopea:
Ok, so I am writing my own parser and I ran into the similar issue.

For a letter "ř", U+0159, index 291 in the previous font, there are following commands:
[44, 136, 154, 247, 6, 247, 99, 14]
44=95, 136=-3, 154=15, (247,6)=114, (247,99)=207, 14 = operator endchar

Do you know, how should be these commands interpreted, to draw "ř" ?

Connum:
Here's the text in question for anyone trying to debug/fix this in the future: Příliš žluťoučký kůň úpěl ďábelské ódy

Connum:
I just had a look at this font - that font does have the glyphs defined, but they contain no path data when opened with our library:
![image](https://github.com/opentypejs/opentype.js/assets/13076806/68de08bd-96fa-4194-a13d-f79aa4c0c5c0)
![image](https://github.com/opentypejs/opentype.js/assets/13076806/4699c12a-2883-4457-baac-78a2019567bc)
![image](https://github.com/opentypejs/opentype.js/assets/13076806/577062e3-ae0c-4972-bdef-a0de87fc01df)
Fontforge does render them
![image](https://github.com/opentypejs/opentype.js/assets/13076806/b3328b77-6d8f-4b79-9ab8-d7b31ea3b6cb)
but outputs a bunch of warnings when opening the font:
```
...
rcaron's SEAC-like endchar operator is deprecated for Type2
sacute's SEAC-like endchar operator is deprecated for Type2
scircumflex's SEAC-like endchar operator is deprecated for Type2
scaron's SEAC-like endchar operator is deprecated for Type2
...
```

So it looks like this could be the culprit. Even though it's deprecated, we should get these paths like other font software does.

Connum:
The glyphs in question are also composite glyphs, so it seems like we don't support those under specific circumstances...

Connum:
This test case is now part of https://github.com/unicode-org/text-rendering-tests. For opentype.js, it will be fixed via #633

PR #616, opened 2023-09-21
fix(charset): Implemented charset according to specification

ln-north:
## Description

Implemented Charsets in accordance with the specification

- Enable retrieval of corresponding CID for GID in CID fonts
- Utilize Predefined Charsets from Adobe TN #5176

## Motivation and Context
- In CID fonts, we use the CID obtained from Charsets for glyph names. Without this, development and rendering of CID fonts become difficult.
- Aligned with [fonttools](https://github.com/fonttools/fonttools/blob/532a771b985356ca9e69b94cc57f6c926b2c7442/Lib/fontTools/cffLib/__init__.py#L1617-L1662) implementation as it was significantly different.

## How Has This Been Tested?
- Tested that in CID Keyed fonts, the glyph name changes from gid2 to cid00002.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/src/encoding.js
+++ b/src/encoding.js
@@ -45,6 +45,71 @@ const cffStandardStrings = [
     'Uacutesmall', 'Ucircumflexsmall', 'Udieresissmall', 'Yacutesmall', 'Thornsmall', 'Ydieresissmall', '001.000',
     '001.001', '001.002', '001.003', 'Black', 'Bold', 'Book', 'Light', 'Medium', 'Regular', 'Roman', 'Semibold'];
 
+// Strings below index 392 are standard CFF strings and are not encoded in the font.
+
+const cffISOAdobeStrings = [
+    '.notdef', 'space', 'exclam', 'quotedbl', 'numbersign', 'dollar', 'percent', 'ampersand', 'quoteright',
+    'parenleft', 'parenright', 'asterisk', 'plus', 'comma', 'hyphen', 'period', 'slash', 'zero', 'one', 'two',
+    'three', 'four', 'five', 'six', 'seven', 'eight', 'nine', 'colon', 'semicolon', 'less', 'equal', 'greater',
+    'question', 'at', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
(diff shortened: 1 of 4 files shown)


Connum:
I did a quick review and it looks fine to me, however due to limited time at the moment I might have overlooked something and would like to have another reviewer do a more thorough review and testing.

Connum:
I had hoped that this might also fix #117, but unfortunately it does not. Could you take a look at this and see if there's an easy fix for this or if it's a separete problem?

ln-north:
Thanks for your review, I got it. I will try it.

Connum:
I just added the dummy text used to #117 for debugging/fixing

Connum:
btw, #117 had a different cause and has meanwhile been fixed

ln-north:
Understood, this PR is now awaiting review.

yne (approved):
can't test right now
but found some unnecessary spaces/tabs/newline
and also some unrelated `"` to `'` changes that would be cleaner if removed.
but otherwise seems good 👍

yne (approved):

Connum:
What specification was that change from gid* to cid000...* based on?
We are now failing some of the unicode tests that we passed before, because the tests actually expect a name of "gid*":
CFF-1 and CFF-2 are no longer passing, despite the paths matching, because of the difference in glyph names.

yne:
@Connum shall we add those test to the CI ? `npm run test` went fine.

Connum:
I already have a workflow file prepared on my fork that will run the tests for PRs and add a comment with the results, as well as save the report as a build artifact. I just have to think of a way to make sure that it will mark the workflow as failed when a test fails that previously passed. We'll have to store this info somehow. Could do it by automatically updating a file, but that would pollute the commit history, and I'm not sure if we want that.

How shall we go about the naming issue? I think we should stay in line with all the other implementations, right?

yne:
I like the idea of a larger test set (even if non-fully-passing)

but there is no need for a file to store the delta. just run the test twice:
- one on the latest PR commit
- one on master

(more failure = CI return fail)

no opinion on the naming, maybe @ln-north have more insight

Connum:
That's a good idea. We just have to make sure that it catches when something let's a new test pass but breaks an old one,which would result in no change in numbers, so we'll have to compare the actual test results.

Connum:
I mean we could probably adapt the script used for rendering the glyphs for the test suite to generate the gid* names for the output, but that doesn't seem to be necessary for any of the other engines. I just want to make sure we don't break anything.

PR #621, opened 2023-10-19
add tests for and fix #613, test toDOMElement() via mocked elements

### LINKED ISSUE #613: Path properties do not translate in Path.toSVG(2)
## Expected Behavior
The fill color, stroke color and stroke thickness properties in the Path object should be included in the conversion of path object to path XMLElement. If the object contains these properties, the XMLElement should contain these same properties, as shown in the example here: https://www.w3.org/TR/SVG/paths.html#PathData

## Current Behavior
Currently, the properties are touched upon in the readme, but not actually included in the conversion when running the path.toSVG() function. I noticed that they are included in the code, but an incorrect if statement has an unwanted outcome that is not actually tested in the test cases. 

## Possible Solution
Resolve the contradicting code in path.js and add test cases for the additional properties. 

1. Load a .ttf font
2. Obtain the path object of a specific glyph by calling glyph.getPath(0, 0, 72)
3. Do not add a fill color (or set path.fill = 'black') and do console.log(path.toSVG(2)). The fill property is not included. Similarly, visualizing yields a white canvas. 

## Context
Visualize the font. 

## Your Environment
* Version used: 
* Font used:
* Browser Name and version:
* Operating System and version (desktop or mobile):
* Link to your project:

Connum:
## Description

This fixes #613 which was caused by an if statement that could never be fulfilled, leading to `Path` properties not all making their way to the generated SVG output of `toSVG()`. It also fixes the properties being completely ignored by `toDOMElement()` and allows to test that function in the first place by mocking `document.createElementNS()`.

## Motivation and Context

We weren't able to test `toDOMElement()` in Node context before, that's why we missed that the properties were completely ignored. There was also no test case that would have caught the missing property for `toSVG()`.

## How Has This Been Tested?

`npm run test`

## Types of changes

## Checklist:


yne (approved):
LGTM

Issue #623, opened 2023-10-20
Opentype.js can only draw black text.


ILOVEPIE:
## Expected Behavior
There should be a way to specify the color of the text being drawn.

## Current Behavior
opentype.js is the Henry Ford of font rendering "Any [user] can have [their text] any color that [they want] so long as it is black."

## Possible Solution
Allow a color to be selected in the draw call for Font and Glyph objects.

## Context
My library has to manually draw the path to work around this, granted I have to do this anyways as I need to stroke the text too, but my point stands.

axkibe:
My take on this is, maybe opentype.js shouldn't draw at all. Why duplicate API? Have opentype.js just make the path on the canvas an the caller can fill and stroke it.

ILOVEPIE:
> My take on this is, maybe opentype.js shouldn't draw at all. Why duplicate API? Have opentype.js just make the path on the canvas an the caller can fill and stroke it.

We could do it either way.

Shmacked:
> My take on this is, maybe opentype.js shouldn't draw at all. Why duplicate API? Have opentype.js just make the path on the canvas an the caller can fill and stroke it.

For me, I replaced the native text component for fabric.js with opentype.js drawing command. I had to rewrite the drawing command in order to be able to apply stroke and fill (and other effects, if I wanted) on it, since this isn't native to fabric.js, yet. I could convert them to paths, but I wanted the ability to to edit text and fonts.

My fix is not perfect, but it pretty dang close to getting accurate bounding boxes on the text with a given font.

PR #626, opened 2023-10-20
implement styling of fill/stroke/strokeWidth when using Font.draw() or Glyph.draw()

Connum:
## Description

Implements a new method `Path.applyStyles()`, which applies valid style properties (fill, stroke, strokeWidth) to the Path. This is used in `Glyph.getPath()` and `Font.getPath()` and allows to specify styling options e.g. when using `Font.draw()` or `Glyph.draw()`.

## Motivation and Context

Implements #623.

## How Has This Been Tested?

Temporarily modifying `docs/index.html`

options in `renderText()` of `docs/index.html`
image
result:
image

## Types of changes

## Checklist:


ILOVEPIE:
I'll review this later today.

ILOVEPIE (changes requested):
One discrepancy here with regard to other font renderers is the fact that lineWidth draws half the line inside the fill area, which means our stroke is actually half-width this way (at least with canvas, not sure about SVG). Also, Miter type becomes important when rendering the stroke on text, so we need to do some additional research with regards to that before this can go through. In my code base I'm using  ``lineCap = "round"`` and ``lineJoin = "round"`` but this may only be correct for my use case.

ILOVEPIE:
> I'll review this later today.

I ended up having a little time to go through it now before work.

Connum:
I would treat the outline positioning as a separate follow-up issue. It comes into play during the actual canvas drawing and we'll have to take it into account there, but at least the values get passed along with the Path now.

But good point with the line behaviour. We should take a look at that and find good values for it, and probably add an option to influence it according to one's needs.

ILOVEPIE:
In my codebase I'm doing a kind of dirty hack by multiplying the outline width by 2 then filling the text with ``destination-out`` compositing mode to cut out the inner part of the stroke.

Connum:
> In my codebase I'm doing a kind of dirty hack by multiplying the outline width by 2 then filling the text with `destination-out` compositing mode to cut out the inner part of the stroke.

That's what I thought of doing. Doesn't even feel too dirty to me.

ILOVEPIE:
> > In my codebase I'm doing a kind of dirty hack by multiplying the outline width by 2 then filling the text with `destination-out` compositing mode to cut out the inner part of the stroke.
> 
> That's what I thought of doing. Doesn't even feel too dirty to me.

The problem with that is that it doesn't work for SVG output, unless we want to use masking.

Issue #464, opened 2021-01-11
Can we release a new version?


wtagain:
The lasted released version is 9 months ago，during this time， some issues have been fixed，so can we release a new version?

wtagain:
@Jolg42 Hi, can we release a new version?

Connum:
An additional year later, I'd love to see the support for more lookup tables I added released as well! :)

Jolg42:
Hi there!
I'll try to check that in the next days, though I didn't follow what happened here in a while and could use help if someone wants to be a maintainer 😃

saragee3:
Bumping this — will there be a new release soon? Looks like there are some commits after 1.3.4 that addresses performance improvements, additional glyph support, etc. Thanks!

ILOVEPIE:
If this project needs a maintainer, I'd be happy to help, I've got a project that depends on this one and really need some of these features and PRs to be merged.

ILOVEPIE:
If I don't hear back from @Jolg42 within a week or two on this, I'll just continue development under my organization's fork.

Connum:
> If I don't hear back from @Jolg42 within a week or two on this, I'll just continue development under my organization's fork.

That would be a real shame!

ILOVEPIE:
> > If I don't hear back from @Jolg42 within a week or two on this, I'll just continue development under my organization's fork.
> 
> That would be a real shame!

Why? It would still be open source.

Connum:
> Why? It would still be open source.

Sure, but this could be a chance to breathe new life into this seemingly abandoned project, and most people will probably not be looking at the forks, so having a new maintainer would be preferable to abandoning the main repo and moving active development to a lesser known fork.

ILOVEPIE:
Still waiting on @Jolg42, he apparently saw a message I sent him on mastodon, but so far no movement yet.

ILOVEPIE:
Seeing as it's been several months with no movement, I will be continuing development of opentype.js on my fork which is available [here](https://github.com/SABRE-JS/opentype.js). If and when @Jolg42 returns I would be more than happy to shift development back to the main repository, but it looks like he's abandoned this project so it doesn't seem likely.

 I will need some co-maintainers at least temporarily to help go through old PRs and apply ones that are worth applying, anyone who wishes to help with that process can reply to this message, or contact me via discord where i go by ILOVEPIE#9201.

Connum:
That's great news!
[...]

PR #630, opened 2023-10-26
Fixed metadata that enables fonts to be recognised as a family (tested!)

mattlag:
This is an updated version of #606 with a few additions that allow the code to pass `npm run test`. @Connum, If this gets merged, we can also close #606 .

From @Lmpessoa
I've found out that a few fields that might be passed to opentype.js were not being used while producing the font. I have been able to pinpoint those fields and where they should be applied to produce fonts that are recognised as a family. Tested only with a couple of fonts but changes were regonised by Windows and other apps.

An issue regarding the fields hhea.caretSlopeRise and hhea.caretSlopeRun for italic and oblique fonts is still pending review but does not prevent font usage or recognition as a family (caret slope had a weird inclination while testing on Word but not on Affinity Designer).

These same changes were applied to the copy of the library used by Glyphr Studio (v2) where the tests were conducted.

## Types of changes

## Checklist:


src/font.js
```diff
@@ -87,11 +87,43 @@ function Font(options) {
...
+        } else if (this.italicAngle > 0) {
+            selection |= this.fsSelectionValues.OBLIQUE;
+        }
+        if (this.weightClass >= 600) {
+            selection |= this.fsSelectionValues.BOLD;
+        }
+        if (selection == 0) {
+            selection = this.fsSelectionValues.REGULAR;
+        }
+        if (options.fsSelection) {
```
Connum, line 105: If `options.fsSelection` is set, all the other checks above are not necessary, so we should check that first and only perform the other checks otherwise.

Connum (changes requested):
See my one comment, otherwise it looks good.
However, while the current tests are now passing, there are no tests to test the new functionality itself. We should test for
* the values being calculated/set correctly
* the panose fallback working and its values and the resulting os2 properties taking on the corresponding values
* the `macStyle` value being written correctly

src/font.js
mattlag: I was able to address 3 out of your 4 comments:
* I added an initial check to see if `fsSelection` was specified, and only calculate it's value if it wasn't specified.
* Added a test to make sure Panose generates default values
* Added a test to make sure, in the case where `fsSelection` was not specified, that the auto-calculation happens 

The one thing I couldn't figure out is how to access `macStyle` to test it. It seems that, after a font is generated, `font.tables.head.macStyle` does not exist. I tried to add a test for this, but it failed (due to it not existing)

Do you know how to access `macStyle` after the font is generated, so I can add a test for it?

Connum: When you create a new font object, there is no head table being created automatically. What we want to test here is if the provided data is reflected in the generated table data when saving the font, i.e. when `toTables()` is called (which happens internally when downloading the font).

```js
// after creating your font object
font.tables.head = { macStyle: 1 };
const head = font.toTables().tables.find(table => table.tableName === 'head');
assert.equal(head.macStyle, 1);
```

Please also test the panose values when they are specified, right now just the default value fallback is tested. Thanks!

mattlag: Got it - just did some more investigation and I found the `toTables` method. I also looked through the OTF spec to see what the `macStyle` bits should be, and discovered they should match the `fsSelection` bits... so I updated how `macStyle` is calculated.

mattlag: All right - just updated the tests and added a test where panose values are specified.

... anything else? Let me know!

Connum: Just one thing: When the spec says the `fsSelection` and `macStyle` bits should "agree", that doesn't mean that they should be exactly the same. Just the corresponding meaning must match (i.e. if `fsSelection` says "bold italic", `macStyle` should say "bold italic" as well). But the bits used actually differ, see
https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fsselection (here you can even see the bits used for `macStyle` in the table)
and https://learn.microsoft.com/en-us/typography/opentype/spec/head
> Bit 0: Bold (if set to 1);
> Bit 1: Italic (if set to 1)
> Bit 2: Underline (if set to 1)
> Bit 3: Outline (if set to 1)
> Bit 4: Shadow (if set to 1)
> Bit 5: Condensed (if set to 1)
> Bit 6: Extended (if set to 1)
> Bits 7–15: Reserved (set to 0).
It would be best to add these values in analogy to `Font.prototype.fsSelectionValues = {...}` in `font.js` and make use of them like that.

src/font.js
```diff
@@ -52,7 +52,8 @@ function createDefaultNamesInfo(options) {
  * @property {Number} ascender
  * @property {Number} descender
  * @property {Number} createdTimestamp
- * @property {string=} weightClass
+ * @property {Number} weightClass
+ * @property {Number} italicAngle
```
Connum, line 56: Oh and I think the slope property is missing here

src/tables/head.js
```diff
@@ -52,7 +52,7 @@ function makeHeadTable(options) {
         {name: 'yMin', type: 'SHORT', value: 0},
         {name: 'xMax', type: 'SHORT', value: 0},
         {name: 'yMax', type: 'SHORT', value: 0},
-        {name: 'macStyle', type: 'USHORT', value: 0},
+        {name: 'macStyle', type: 'USHORT', value: options.macStyle},
```
Connum, line 55: might be good to add a default value `|| 0` just in case

Connum:
I added two more comments for some minor things. If that and the `macStyle` bit are resolved, we should be good to go!

src/font.js
mattlag: Oh, good catch - I misinterpreted the meaining.  I'll make sure the `macStyle` bits are accurate based on the reference you linked to above.

src/font.js
mattlag: I looked through the code (which I picked up from the other pull request) `slope` was being added as a property to the font object, and also added to the `hhea` table... but there is no `slope` property on the `hhea` table, and the font property was not being used for any calculation. So, I removed `slope` from the PR.

src/font.js
mattlag: Bit names added as you suggested, and now used while calculating `macStyle`

src/tables/head.js
mattlag: I resolved this by adding the line 

```javascript
let macStyle = options.macStyle || 0;
```

mattlag:
@Connum I marked our conversations as resolved, but please let me know if you have any further comments!

src/font.js
Connum: But it's a property of the font object, so it should be added here. Or am I missing something?

Connum: Oh, now I get what you're saying. So the slope shouldn't have been part of this PR in the first place, got you!

Connum (approved):

Connum:
Could you please merge the master branch back in and resolve the conflicts? Then this would be ready to be merged back in!

src/font.js
mattlag: Yeah, I think Lmpessoa added it, thinking he would use it to also work on the `hhea` table values `caretslopeRise` and `caretSlopeRun`... but he decided not to do that part as it wasn't necessary for fonts being recognized as a family. `slope` was declared but never used anywhere.

mattlag:
Sorry, my inexperience showing again. I'm not sure exactly how to merge the master branch back in.

I see this at the base of the pull request:
image

And in GitHub desktop I see this:
image

Connum:
It's all good! I wanted to do a rebase first, but squash and merge worked fine. Thanks for your contribution, really appreciated!

mattlag:
Hey, @Connum, I want to give you a huge thank-you for helping me out. I'm essentially a UX Designer pretending to be a JavaScript developer 😅 so the extra time you put in reviewing my code and helping me out - I really appreciate it!

mattlag:
One last question you may know the answer two - when does opentype.js ship a 'release'? v1.3.1 was released in 2020, what is the criteria for releasing 1.3.2? I want to use this new feature that just got merged. Should I just use a random build, or is 1.3.2 a thing that will happen soon?

Connum:
That's a good question, and unfortunately this is still to be discussed. Only about a year ago was this project revitalised thanks to @ILOVEPIE 's initiative (see #464), but as it often happens, life and paid work got in the way and active work on the project was dormant for several months again.

Only recently have I found some time to review PRs and issues again and fix bugs myself. There are already some bigger features and refactorings in the master since the last version was released, and several in line in the form of outstanding PRs. Some internally used methods have changed that might have been exposed and used externally, therefore the next release will most probably be 2.0.0.

Meanwhile, I would use the current master as a github dependency (`npm install "https://github.com/opentypejs/opentype.js.git#master"`), keeping in mind that even though the master branch is protected by required code reviews for each PR and disallowing direct code changes, there might still be the case that something gets messed up or even broken when there is no test catching it. Although I'd say that risk is only marginally higher than with any proper release...

Would be lovely to see more contributions in the future, I'm always happy to help if my time allows!

mattlag:
This is great, thank you for all the context!

src/font.js
```diff
@@ -87,11 +88,45 @@ function Font(options) {
...
+
         this.tables = Object.assign(options.tables, {
             os2: Object.assign({
                 usWeightClass: options.weightClass || this.usWeightClasses.MEDIUM,
                 usWidthClass: options.widthClass || this.usWidthClasses.MEDIUM,
-                fsSelection: options.fsSelection || this.fsSelectionValues.REGULAR,
+                bFamilyType: options.panose[0] || 0,
+                bSerifStyle: options.panose[1] || 0,
+                bWeight: options.panose[2] || 0,
+                bProportion: options.panose[3] || 0,
```
Finii, line 122: This PR introduces names for the individual Panose values.
I think this is problematic.

When one sees a property `bWeight` one might assume it has something to do with the font weight. But the Panose values are ... a mess. The meaning of a certain value depends on the 1st Panose value.

For example with the 4th Panose value `panose[3]` contains
* `Proportion` for Latin Text
* `Spacing` for Latin Script and Latin Picture
* `Aspect` for Latin Decorative

So I would not call it `bProportion` as that is an oversimplification. I believe it is best to keep just the array and let interested uses decode it by for example this excellent compilation that I usually use for reference:

image
_https://forum.high-logic.com/postedfiles/Panose.pdf_

Connum:
pinging @yne @ILOVEPIE

yne:
not sure if I can be of any help here

Finii:
Sorry, "I think this is problematic." is kind of an exaggeration.
I just stumbled about this when I tried to fix Panose.

The real "problem" is described in the new PR, \<will be filled in in a minute\>, which fixes the Panose write operation.

PR #631, opened 2023-10-29
add tests for LONGDATETIME parsing/encoding

Connum:
## Description

add tests for parsing and encoding the LONGDATETIME type

## Motivation and Context

This was not tested before, and a FIXME comment suggested to do so once it was implemented (which it has been for a long time).

## Types of changes

- [X] Testing

## Checklist:

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -181,6 +181,7 @@ Parser.prototype.parseTag = function() {
 // JavaScript and unix timestamps traditionally use 32 bits, so we
 // only take the last 32 bits.
 // + Since until 2038 those bits will be filled by zeros we can ignore them.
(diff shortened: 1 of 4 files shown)


ILOVEPIE (approved):
This looks good, but we should probably finish implementing LONGDATETIME, or at least add it to the issue tracker. We should also start maintaining a separate issues list on trello or github projects.

Connum:
> we should probably finish implementing LONGDATETIME

I already looked into that but we'd need to use `BigInt()`, which means we would either loose backward compatibility or have to use a library to handle the 64bit timestamps. All that for an issue that we'll face in 15 years from now, where it should be possible to use `BigInt()` by then.

PR #635, opened 2023-11-04
fix setting preferred subfamily

StreakInTheSky:
## Description

Fixed how a font's `preferredSubfamily` is set.
Currently in the `tables/sfnt.js` file inside `fontToSfntTable`, when setting a font's `preferredSubfamily` it's looking for `preferredSubFamily`, so if it's not already set it will always be `undefined`. Changed the `F` to an `f`

## Motivation and Context

Fixes `undefined` preferredSubfamily if it's not explicitly set

## How Has This Been Tested?

Testing a feature from this [repo/site](https://github.com/eigilnikolajsen/commit-mono). which parses font files which does not have `macintosh.preferredSubfamily` set, but has `windows.preferredSubfamily` set. Specifically fonts from [here](https://github.com/eigilnikolajsen/commit-mono/tree/main/src/fonts/fontlab).  Launched the site locally using a simple http server and ran the customization feature from the site, while testing how the font object's `name` props change.

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/sfnt.js
+++ b/src/tables/sfnt.js
@@ -316,15 +316,15 @@ function fontToSfntTable(font) {
     }
 
     if (!names.unicode.preferredSubfamily) {
-        names.unicode.preferredSubfamily = fontNamesUnicode.fontSubFamily || fontNamesMacintosh.fontSubFamily || fontNamesWindows.fontSubFamily;
+        names.unicode.preferredSubfamily = fontNamesUnicode.fontSubfamily || fontNamesMacintosh.fontSubfamily || fontNamesWindows.fontSubfamily;
     }
 
     if (!names.macintosh.preferredSubfamily) {
-        names.macintosh.preferredSubfamily = fontNamesMacintosh.fontSubFamily || fontNamesUnicode.fontSubFamily || fontNamesWindows.fontSubFamily;
+        names.macintosh.preferredSubfamily = fontNamesMacintosh.fontSubfamily || fontNamesUnicode.fontSubfamily || fontNamesWindows.fontSubfamily;
     }
 
     if (!names.windows.preferredSubfamily) {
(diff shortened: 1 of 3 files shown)


Connum:
Thanks for your contribution! I think the first link is incorrect, because it links to the repo root of the font of your second link, and not some testing tool.

Could you also add a test case for this issue, please?

StreakInTheSky:
Sorry, didn't really write a standalone script or test to test this out. Since I was trying to fix a particular issue in that repo, just did it manually through my fork of that site.

But I can write a test case for it. I do want to clarify of how I should go about it. I was thinking of creating a new test file for `sfnt.js`, And then just make some test instances of `Font` with the relevant data inside the `names` property. Then I'd test if the Font's `name` values are correct. after running through the function.

Connum:
Basically we want to test `fontToSfntTable` in `sfnt.js`, but there's a wrapper on the Font object called `toTables`, so I think we could add it there as well. There is already a test making sure that table data is being serialized, the code of which is a good example of how to access the resulting structure. The new test should then get the name table and check for the correct value.

StreakInTheSky:
Hi, so I added another script in package.json to just run mocha so I could specify the test files to run. Would you like this committed or named something else?

```json
"scripts": {
    ...,
    "test": "npm run build && npm run dist && mocha --require reify --recursive && npm run lint",
    "test:local": "mocha --require reify --recursive",
    ...
},
```

But as an update, I finally got the tables parsed correctly so I can check the values for equality. Hopefully can get this done in the next day or so.

Connum:
> Would you like this committed or named something else?
Let's not include that with this PR

StreakInTheSky:
Hey sorry, took a bit longer. Pushed the tests for review.

Connum (approved):
LGTM! 🚀

PR #642, opened 2023-11-16
performance: implement caching for roundDecimal() and prevent unnecessary recalculations

### LINKED ISSUE #641: Bad performance of roundDecimal() in path.js
This is more a note for me... `roundDecimal()` in `path.js` that I added for SVG optimization decreases the perfocmance drastically (see #588). It was however the only calculation method I found so far that handles rounding with decimal places reliably and precise enough to pass the tests in the unicode test suite. We'll need to find a remedy for this before releasing 2.0.0. I'm thinking of a cache for the rounding of the decimal places (that should be quite repetitive among the glyph paths).

Connum:
## Description

Implement caching for decimal rounding by storing the results for the decimal part of floats only and adding them to the significant part. Also avoid unnecessary recalculations by storing the value for reuse.

fixes #641 

## Motivation and Context

Using our performance test tool, we could see a huge drop (~1s vs ~6s) in performance from 1.3.4 to the current master. The culprit seems to have been the function `roundDecimal()` in `path.js` alone. That calculation seems to be very performance-hungry, however, it's the only way that I found so far to get reliable rounding of floats without losing precision.

## How Has This Been Tested?

Using the performance test tool in `docs/benchmark.html` comparing to version 1.3.4 before and after the optimization. Also ran `npm run test` as well as the unicode test tool to make sure it doesn't break passing tests.

before
image

after
image

## Types of changes

## Checklist:

- [ ] <s>I have added tests to cover my changes.</s> (not applicable)

### DIFF
--- a/src/path.js
+++ b/src/path.js
@@ -16,8 +16,25 @@ function Path() {
     this.strokeWidth = 1;
 }
 
+const decimalRoundingCache = {};
+
 function roundDecimal(float, places) {
-    return +(Math.round(float + 'e+' + places) + 'e-' + places);
+    const integerPart = Math.floor(float);
+    const decimalPart = float - integerPart;
+
+    if (!decimalRoundingCache[places]) {
+        decimalRoundingCache[places] = {};
+    }
+
+    if (decimalRoundingCache[places][decimalPart] !== undefined) {
+        const roundedDecimalPart = decimalRoundingCache[places][decimalPart];
+        return integerPart + roundedDecimalPart;
+    }
+    
+    const roundedDecimalPart = +(Math.round(decimalPart + 'e+' + places) + 'e-' + places);
+    decimalRoundingCache[places][decimalPart] = roundedDecimalPart;
+
+    return integerPart + roundedDecimalPart;
 }
 
 function optimizeCommands(commands) {
@@ -525,10 +542,11 @@ Path.prototype.toPathData = function(options) {
     options = createSVGOutputOptions(options);
 
     function floatToString(v) {
-        if (Math.round(v) === roundDecimal(v, options.decimalPlaces)) {
-            return '' + roundDecimal(v, options.decimalPlaces);
+        const rounded = roundDecimal(v, options.decimalPlaces);
+        if (Math.round(v) === rounded) {
+            return '' + rounded;
         } else {
-            return roundDecimal(v, options.decimalPlaces).toFixed(options.decimalPlaces);
+            return rounded.toFixed(options.decimalPlaces);
         }
     }
 


yne:
I'm still puzzled how a hashtable lookup can be faster that truncating a stringified float.

Is this truncating function really needed for standard compliance ? or is it for a cosmetic purpose ?

`floatToString` alone seems to call 3 times this function with the same argument each times, so another improvement would have be to use a variable once.

anyway, it's good to see some love about performance.

Connum:
The costly part is `+(Math.round(decimalPart + 'e+' + places) + 'e-' + places)`. So it's not a truncation or a simple `Math.round()`, because that will result in loss of precision. The rounding is done for every value of every point/command in every glyph in the string that is to be rendered. So I'm not really surprised this is quite costly! 😃

Connum:
> `floatToString` alone seems to call 3 times this function with the same argument each times, so another improvement would have be to use a variable once.

I'm already storing that in the variable `rounded`

ILOVEPIE:
I'll take a look at this after my system's done updating.

yne:
> The costly part is `+(Math.round(decimalPart + 'e+' + places) + 'e-' + places)`. So it's not a truncation or a simple `Math.round()`, because that will result in loss of precision. The rounding is done for every value of every point/command in every glyph in the string that is to be rendered. So I'm not really surprised this is quite costly! 😃

I'm struggling to understand the need for a loss precision function. can you show me an example of expected input-output for this ? hopefully it will help me grasp the idea.

Connum:
> > The costly part is `+(Math.round(decimalPart + 'e+' + places) + 'e-' + places)`. So it's not a truncation or a simple `Math.round()`, because that will result in loss of precision. The rounding is done for every value of every point/command in every glyph in the string that is to be rendered. So I'm not really surprised this is quite costly! 😃
> 
> I'm struggling to understand the need for a loss precision function. can you show me an example of expected input-output for this ? hopefully it will help me grasp the idea.

I tried out several rounding methods at the time, starting with a simple `Math.round()` in conjunction with multiplicater and subsequent division in order to round to a specific number of decimals. That resulted in wrong values e.g. when running the unicode test suite, as far as I remember. So did `toFixed()`.

Here is a short blog post about the issues with floating-point arithmetics: https://www.jacklmoore.com/notes/rounding-in-javascript/
It's from 2014 and suggests the same approach we're using, but I haven't found anything more reliable.

ILOVEPIE:
It might be possible to do this using a typed array/array buffer to do it correctly.

yne (approved):

PR #648, opened 2023-11-24
implement graceful loading via logging severity levels

### LINKED ISSUE #337: Corrupted font file handling
The more different tables we parse in opentype.js, the more we come across corrupted font files.
Most of them are still usable with other software, but opentype.js currently throws errors and stops parsing the font.
Shouldn't we try to skip (sub)tables containing errors ?
Should we report errors in some way ? Maybe just keeping a log of ignored tables ?
What do you think ? @fdb @Jolg42 @Pomax and others ?

### LINKED ISSUE #370: TTF not contains name and cmap table parse error
like title,  I parse some ttf maybe from some pdf file, they are not found name and cmap table. so opentype.js throw error
I can open them using fontforge
[font_21.zip](https://github.com/opentypejs/opentype.js/files/2864117/font_21.zip)

### LINKED ISSUE #426: Support of subset-fonts
[Fonts.zip](https://github.com/opentypejs/opentype.js/files/4258838/Fonts.zip)

The attached file BAAAAA+Bauhaus93.ttf was extracted from a PDF.
It was build as subset of BAUHS93.TTF and contains only the glyphs used by "Hello World" (=> 'H', 'e', 'l', 'o', ' ', 'W', 'r', 'd').

## Expected Behavior

Is it possible to extend opentype.js to support such subset-fonts also.

## Current Behavior

Currently an error-messae occured: `Error: No valid cmap sub-tables found.`

## Possible Solution

[fontforge](https://fontforge.org/en-US/) can open BAAAAA+Bauhaus93.ttf and view the glyphs of it.

1. Open https://opentype.js.org/index.html
2. Open BAAAAA+Bauhaus93.ttf 
3. The error-message occured
4.

## Your Environment

* Version used:
* Font used: BAAAAA+Bauhaus93.ttf 
* Browser Name and version: FireFox 72.0.1 (64bit)
* Operating System and version (desktop or mobile): Linux
* Link to your project:

### LINKED ISSUE #607: Font validate is incomplete
## Expected Behavior

`Font.validate` should return a boolean or the list of warnings that have been generated.

## Current Behavior

`Font.validate` doesn't do anything.

## Possible Solution

After asserting return the warnings array.

## Context
I'm referencing [this](https://github.com/opentypejs/opentype.js/blob/d9e369e8b0e9d081dc1d752cfca89967d0bae8d8/src/font.js#L471) method.

### LINKED ISSUE #643: Error parsing some cut font files using 'opentype. parse(...)'
**Firstly, I only know a little English. I hope you can understand it. Thank you again!!

I have some font files with unknown sources, but I have a rough understanding that they should have been cut, leaving only some text used in the files. But I don't know how they cut this file. Some font files couldn't be opened by double clicking with the mouse, and I tried to parse them using `opentype. parse (...)` from opentype.js. Not surprisingly, different errors will be reported. But to my knowledge, there are some font parsing libraries that can be opened (possibly belonging to the backend). I really need to parse these font files correctly without using only opentype.js, because these cut fonts cannot be directly introduced into the browser. I need to obtain the missing character codes to depict the corresponding text to complete my task.

So, if anyone knows how to solve this problem, or can modify the source code of opentype.js from the perspective of error reporting, please let me know or provide me with relevant guidance. I am very worried that opentype.js will not be able to solve my problem, and I really need your reply!

Here are some screenshots examples of errors and problematic font files:**
Snipaste_2023-11-20_14-36-15
Snipaste_2023-11-20_14-36-42
Snipaste_2023-11-20_14-44-12
[fonts.zip](https://github.com/opentypejs/opentype.js/files/13407817/fonts.zip)

Looking forward to your reply！！

Connum:
## Description

**_This is a work in progress, but I wanted to get feedback before adding tests, changing all the instances of `console.*` and `throw` and adding to the documentation._**

This implements a logger that allows to log messages with different severities (`ERROR`, `WARNING`, `DEPRECATED`). By default, everything gets logged to the console, and only messages of type `ERROR` get thrown. This can be set via `setLogLevel()` and `setThrowLevel()`, using binary flags, e.g. `opentype.ErrorTypes.ALL ^ opentype.ErrorTypes.DEPRECATED` to log everything but `DEPRECATED` messages.

When adding a message to the log, this will also trigger a custom `opentypejs:message` event, so you can catch all the different message types, e.g. for displaying them on our demo pages. Speaking of which, those have also been adapted to handle missing values more gracefully and log multiple messages instead of just the last error thrown.

During initial parsing, many formerly thrown error messages have been changed to messages of type `WARNING`. This allows incomplete and corrupted fonts to load. To test one of those, I also implemented parsing of standalone CFF fonts, because it was fairly simple. Those are basically standalone CFF tables, so we had everything ready for parsing and just needed to handle some font information that we get from other tables in OpenType CFF files. Might need to extend that a bit further in the future, but it's OK for getting basic glyph and meta data.

Last but not least, the same logging functionality as above is used for `Font.validate()`. It uses a separate logger instance so the log and throw levels can be set differently from the global logger, and in any case the collected messages will be returned as an array so they can be handled separately.

## Motivation and Context

fix #643, fix #426, fix #370, fix #337 (all related to subset font data)
fix #607 (`Font.validate()`)

## How Has This Been Tested?

Tested all the subset files provided (unfortunately can't simply use those for testing due to unclear licensing).

**TODO**
* add some general tests regarding error logging/handling
* find or produce some usable test files if I can
* add tests for `Font.validate()`

image
`font.validate()` finally being useful

image
Showing multiple warning messages in the demo

image image
parsing subset fonts

image
parsing standalone CFF (subset) fonts

## Types of changes

## Checklist:

- [ ] [TODO] *I have added tests to cover my changes.*
- [ ] [TODO] *I have updated the **README** accordingly.*

### DIFF
--- a/docs/font-inspector.html
+++ b/docs/font-inspector.html
@@ -144,11 +144,23 @@ function showErrorMessage(message) {
     var el = document.getElementById('message');
     if (!message || message.trim().length === 0) {
         el.style.display = 'none';
+        el.innerHTML = '';
     } else {
         el.style.display = 'block';
+        el.innerHTML = `<p class="message-type-1">${message}</p>`;
     }
-    el.innerHTML = message;
 }
+
+function appendErrorMessage(message, type) {
+    var el = document.getElementById('message');
+    el.style.display = 'block';
+    el.innerHTML += `<p class="message-type-${type}">${message}</p>`;
+}
+
+document.addEventListener('opentypejs:message', function(event) {
+    const message = event.detail.message;
+    appendErrorMessage(message.toString(), message.type);
+});
 
 function sortKeys(dict) {
     var keys = [];
@@ -257,10 +269,11 @@ async function display(file, name) {
     }
     try {
         const data = await file.arrayBuffer();
-        onFontLoaded(opentype.parse(isWoff2 ? Module.decompress(data) : data));
         showErrorMessage('');
+        onFontLoaded(opentype.parse(isWoff2 ? Module.decompress(data) : data));
     } catch (err) {
         showErrorMessage(err.toString());
+        throw err;
     }
 }
 form.file.onchange = function(e) {
--- a/docs/glyph-inspector.html
+++ b/docs/glyph-inspector.html
@@ -92,11 +92,23 @@ function showErrorMessage(message) {
     var el = document.getElementById('message');
     if (!message || message.trim().length === 0) {
         el.style.display = 'none';
+        el.innerHTML = '';
     } else {
         el.style.display = 'block';
+        el.innerHTML = `<p class="message-type-1">${message}</p>`;
     }
-    el.innerHTML = message;
 }
+
+function appendErrorMessage(message, type) {
+    var el = document.getElementById('message');
+    el.style.display = 'block';
+    el.innerHTML += `<p class="message-type-${type}">${message}</p>`;
+}
+
+document.addEventListener('opentypejs:message', function(event) {
+    const message = event.detail.message;
+    appendErrorMessage(message.toString(), message.type);
+});
 
 function pathCommandToString(cmd) {
     var str = '<strong>' + cmd.type + '</strong> ' +
@@ -273,13 +285,19 @@ function renderGlyphItem(canvas, glyphIndex) {
     var cellMarkSize = 4;
     var ctx = canvas.getContext('2d');
     ctx.clearRect(0, 0, cellWidth, cellHeight);
-    if (glyphIndex >= window.font.numGlyphs) return;
+    const nGlyphs = window.font.numGlyphs || window.font.nGlyphs;
+    if (glyphIndex >= nGlyphs) return;
 
     ctx.fillStyle = '#606060';
     ctx.font = '9px sans-serif';
     ctx.fillText(glyphIndex, 1, cellHeight-1);
-    var glyph = window.font.glyphs.get(glyphIndex),
-        glyphWidth = glyph.advanceWidth * fontScale,
+    const glyph = window.font.glyphs.get(glyphIndex);
+    if (!glyph.advanceWidth) {
+        // force calculation of path data
+        glyph.getPath();
+    }
+    const advanceWidth = glyph.advanceWidth;
+    let glyphWidth = glyph.advanceWidth * fontScale,
         xmin = (cellWidth - glyphWidth)/2,
         xmax = (cellWidth + glyphWidth)/2,
         x0 = xmin;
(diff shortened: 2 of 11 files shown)


ILOVEPIE:
This is a good idea. We need a more robust logging system and one that's more compatible with other logging systems.

ILOVEPIE:
We should make it so that event.preventDefault makes it so that the message does not get printed if it would have been. The reason for that is so that it can be used to integrate with other logging systems.

Connum:
Can you be more specific about "other logging systems"? Anything particular I should take look at as a reference?

ILOVEPIE:
there are JavaScript logging and log capturing libraries that are used to manage logs or aggregate them from clients to a central logging server. By providing an event like this, we could then have the warning from our system be suppressed and passed on to the other logging library.

Connum:
Just pushed a change to make the event cancelable via `preventDefault()`.

docs/font-inspector.html
```diff
@@ -144,12 +144,24 @@ <h1>Free Software</h1>
     var el = document.getElementById('message');
     if (!message || message.trim().length === 0) {
         el.style.display = 'none';
+        el.innerHTML = '';
     } else {
         el.style.display = 'block';
+        el.innerHTML = `<p class="message-type-1">${message}</p>`;
```
yne, line 150: use innerText and set the class to el directly

yne:
This feel over engineered.
As I understand this try to fix 2 issues:
- the ability to customize logging
- the ability to parse more resiliently


So why not allow to set at Font.parse() time:
- a logger object (that default to the platform agnostic `console`)
- a `resilientParse` boolean
?

Connum:
@yne we don't have access to the current font object in all the places where we might throw or log an error, in order to access an option like `resilientParse`. And we would have to check that option every time we want to output an error, so we would need a wrapper for the default console anyway.

docs/font-inspector.html
Connum: Then we'd loose the wrapping paragraph element and set the class to the outer message area. But you're right, we definitely don't want the message to be interpreted as HTML. so I'll have to create a new paragraph element and set the class and text on that.

ILOVEPIE:
Is this ready for review? If so, we should mark the PR as no longer being a draft in the description.

Connum:
> Is this ready for review? If so, we should mark the PR as no longer being a draft in the description.

Nope, #704 implements the graceful loading part of this, and there was never a fruitful discussion about the logging aspect of this PR

PR #653, opened 2023-11-27
better browser/node detection

### LINKED ISSUE #174: Using web workers
I'm trying to include opentype support into another open project (openJScad) that does just about everything in web workers. Because of this, opentype.load() defaults to loadFromFile (because there's no window). I've worked around this by just forcing load() to use loadFromUrl(), but I'm now getting a cross origin error when trying to load the file from my server.

"XMLHttpRequest cannot load . Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https, chrome-extension-resource."

Is there any way to get around this?

Also, is trying to use opentypejs in a web worker just futile? Should I look for another solution altogether?

Thanks

Connum:
## Description

Made browser/node context detection more reliable, so we'll pick the correct loading methods even when running in a WebWorker context.

## Motivation and Context

close #174 

## How Has This Been Tested?

Can't really add a test for this, but checked that our tests (= Node) and docs (= Browser) still work

## Types of changes

## Checklist:

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,9 +1,16 @@
 function isBrowser() {
-    return typeof window !== 'undefined';
+    return (
+        typeof window !== 'undefined' ||
+        typeof WorkerGlobalScope !== 'undefined'
+    );
 }
 
 function isNode() {
-    return typeof window === 'undefined';
+    return (
+        typeof window === 'undefined' &&
+        typeof global === 'object' &&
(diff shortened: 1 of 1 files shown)


ILOVEPIE (approved):
I can't help but feel there's a better way to handle this, but this should be good for now.

PR #656, opened 2023-11-27
add eslint-import rule no-cycle

Connum:
## Description

Added the eslint-import plugin and activated the `no-cycle` rule, so we'll catch any accidental circular dependencies.

## Motivation and Context

I saw the comment in **glyph.js** regarding the circumvention of a possible circular dependency with **tables/glyf.js** and wanted to make sure we catch similar situations in the future.

## How Has This Been Tested?

`npm run lint`

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
       "devDependencies": {
         "esbuild": "^0.19.6",
         "eslint": "^8.54.0",
+        "eslint-plugin-import": "^2.29.0",
         "mocha": "^8.4.0",
         "reify": "^0.20.12"
       }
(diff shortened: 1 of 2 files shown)


ILOVEPIE (approved):
This looks good to me.

yne (commented):
isn't it already supposed to be detected at ci level (when building or running the tests ?)

Connum:
I think I tested it by producing a circular dependency in the glyf file and esbuild didn't complain about it iirc. But it's still an antipattern.

yne:
I added a `import {parse} from './opentype.mjs';` in `font.mjs` (and added a fake `if(parse)parse;` in a font function)

but `npm run lint` did not reported it.

Did I missunderstood it usage ? Or did it simply never worked ?

Connum:
It must have worked at one point, or I wouldn't have made the PR :) But maybe some changes lead to it not working anymore, as we don't have any tests for the linting

PR #662, opened 2023-12-21
Add glyphName, htmlCode, cssCode to glyph inspector.

ahaoboy:
## Description

Referencing the site [website](https://beautifulwebtype.com/fira-code/glyphs/?i=1), adding glyphName, glyph, htmlCode, cssCode to the glyph inspector for easy copying. Due to no response on https://github.com/opentypejs/opentype.js/pull/586, I separately submitted this PR.

For cases involving multiple unicode, the use of String.fromCodePoint.apply also works.
```js
var str = '\uD83D\uDC04';
console.log(str)

var str2 =  String.fromCodePoint.apply(null, [55357, 56324])
console.log(str2)
```

## Motivation and Context
For easier copying of unicode characters

## How Has This Been Tested?
Local testing ensures proper functionality when copying and pasting into HTML and CSS.

image

## Types of changes

## Checklist:


Connum (changes requested):
Why go the route loading the font via CSS instead of getting the data directly from the `Font` object? This also has the same issue as #586

ahaoboy:
> Why go the route loading the font via CSS instead of getting the data directly from the `Font` object? This also has the same issue as #586


I'm not familiar with the font API, so I referred to https://github.com/opentypejs/opentype.js/pull/586. I also couldn't find a way to directly obtain the corresponding string from glyphs. If there's a better solution, please guidance me. Additionally, how should we add the string corresponding to the glyphs to their attributes?

image

Issue #669, opened 2024-01-24
hinting broken on the demo page


axkibe:
If you go to the opentype.js demo page and upload a ttf font and activate hinting it will break (see in console exports.DEBUG accessing undefined).

I guess this happened when moving opentype.js to esm things, previously in CJS world I designed it that way so the DEBUG statements get removed from minified builds. This needs to be repaired.
 
A PR is following in a minute.

PS: I recently noted that hinting is doing something wrong with NotoSans-Regular.ttf (as the result will look slightly "jumpy" where it should not) I will have to debug into that issue...

axkibe:
revert PS: the hinter is actually fine. (I'm glad I dont have to go further down this rabbit hole again). I just had a bug in my caching mechanism for rendered glyphs with non-integer bounding boxes.

axkibe:
Can you explain why you closed it, but didn't accept the PR?

Connum:
Indeed we should close this after the PR is merged

ILOVEPIE:
Sorry, I thought you were saying that the bug was with your fork of the library and not with the main branch.

Connum:
This should be fixed now, I encountered it during development for #699.
@axkibe please confirm, so we can close this

Connum:
Checked it again and hinting is working now.

PR #670, opened 2024-01-24
fixing hintingtt DEBUGing for esm builds

axkibe:
Fixes issue #669


Connum (approved):
Good catch! This was also one of the reasons why the hinting checkbox in the docs/index.html was broken. Now the minified builds don't contain the corresponding code anymore, resulting in a smaller file size as well.

PR #672, opened 2024-01-31
feat: add SVG  table support

### LINKED ISSUE #106: Support for 'SVG ' table
The `SVG` table has been added to the ISO OFF and Microsoft OpenType specification. It would be really great to support this in opentype.js :)

kinolaev:
## Description

This PR adds support for ['SVG ' table](http://www.microsoft.com/typography/otspec/SVG.htm).

## Motivation and Context

Fixes #106, #193

## How Has This Been Tested?

I added tests to `test/svg.js` file. These tests parse and encode [Colortube](https://www.fontspace.com/colortube-font-f28146) font.

## Types of changes

## Checklist:


ILOVEPIE:
Is this to implement apple-style emoji fonts? I'll pull in connum on this one as I'm a little out of the loop but this looks good to go.

ILOVEPIE:
One issue I do notice for this is the lack of integration with the drawing code.

kinolaev:
Hello @ILOVEPIE! It is to parse and make color fonts like [Colortube](https://www.fontspace.com/colortube-font-f28146) or [Fattern](https://www.fontspace.com/fattern-font-f48474) where SVG documents are embedded directly into a font file instead of being converted into glyphs.
By drawing code you mean [`Font.prototype.draw`](https://github.com/opentypejs/opentype.js/blob/1.3.4/src/font.js#L398) method, right? In my opinion parsing and making SVG tables are useful features even without the ability to draw SVGs on canvas. So maybe we could leave the drawing part out of this PR? By the way SVG fonts usually have glyphs as fallback.
I'll look into it and try to implement drawing SVGs on canvas in a separate PR if I have time.

ILOVEPIE:
The purpose of the library is to parse, generate and draw fonts. This PR is incomplete imho.

kinolaev:
Ok, I see your point and will try to find time to implement drawing as part of this PR.

Connum:
> The purpose of the library is to parse, generate and draw fonts. This PR is incomplete imho.

IMHO it's OK to implement parsing and writing first and drawing with an additional PR. We have other features where parsing works but writing doesn't, so here we even have parsing and writing in one go.

Connum:
But what I *am* missing is a method to retrieve the SVG document in plaintext, using `extDecoder()` as is done for the tests. This should be done in a deferred way using `defineProperty()` and getters/setters, so we only have to do that calculation when the data is needed, like we already do in other places like glyph paths.

Connum:
Nice! However, rendering of the Colortube-lWAd.otf is working in the font inspector but not the glyph inspector for me. Can you look into this?

Two optimizations we should implement for performance:
- Only de-gzip and build the image when it is requested for a given glyph
- Use the native DecompressionStream instead of tiny inflate if available (recent browser and node versions)

We can implement the second one independently from this PR at a later time, but it would be great if we could have the lazy loading from the start.

Connum:
There also seem to be scaling issues when rendering SVG data in the following fonts from https://github.com/unicode-org/text-rendering-tests/tree/main/fonts
* [TestSVGgradientTransform.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGgradientTransform.otf)
* [TestSVGgzip.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGgzip.otf)
* [TestSVGmultiGlyphs.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGmultiGlyphs.otf)

kinolaev:
I've added the ability to draw SVGs. Before drawing SVG images must be prepared by calling `await font.prepareSVGImages()`. It will render SVGs as HTMLImageElements and cache them.
It's also possible to pass an array of SVGDocumentRecord to that function to prepare images only for specific records.

I've exported `decodeDocument` function to retrieve the SVG document in plaintext.
As SVGs can be gzipped, I've also added utility functions to determine and uncompress a gzip buffer (using `tiny-inflate`).

@Connum, you wanted to have a getter `svg` on `Glyph`, is it right? The problem is that one SVG document can contain multiple glyphs. If you want to get an SVG document for a specific glyph, it must also be transformed after decoding (see `makeGlyphSvgTemplate` function). Transformed SVG documents in plaintext are available in data URL of prepared images (`font.tables.svg.images.get(glyphId).image.src`). Could you please explain a bit further, where you want me to add a getter and what exactly it should return?

Connum:
When getting the `path` for a glyph, we do this via a getter that will calculate the path once when it is requested for the first time, and then returns a cached version of it. We should do the same for the SVGs per glyph instead of preparing all of them at once, which could take quite some time depending on how large and complex the SVG font is.

Connum:
For example, [NotoColorEmoji-SVG.otf](https://github.com/adobe-fonts/noto-emoji-svg/raw/main/fonts/NotoColorEmoji-SVG.otf) is freezing the browser for a considerable amount of time.

kinolaev:
Thank you @Connum for the feedback! It's difficult to prepare images lazily, because `draw` is synchronous and `prepareSvgImages` is asynchronous... But the problem is clear to me now and I'll try to figure it out.

Connum:
I'm currently working on COLR/CPAL font rendering (so far we only support parsing and writing the tables), where I will also implement lazy loading of the layers. I hope to have this ready in the next days, then you could take some inspiration from it.

Connum:
I hope I'll finish the cpal/colr overhaul today or tomorrow. Maybe it will help you fix the performance issue.

Connum:
Haven't created a PR yet, but you can already take a look at my current work in progress here:
https://github.com/Connum/opentype.js-next/tree/feature/colr-cpal-rendering

regarding the lazy loading, especially:
https://github.com/Connum/opentype.js-next/blob/feature/colr-cpal-rendering/src/glyphset.js#L136

kinolaev:
Thank you, @Connum , I've added `svgImage` getter. Does it looks good to you?
I've also implemented decompression using native `DecompressionStream` when it's available.

ILOVEPIE:
We should definitely support all of the different methods of doing emoji that have been implemented into fonts, although I will say that only Apple supports the SVG table currently.

Connum:
> We should definitely support all of the different methods of doing emoji that have been implemented into fonts, although I will say that only Apple supports the SVG table currently.

I'll have a PR for CPAL/COLR v0 rendering ready by (hopefully) tomorrow. v1 is quite complex though.

kinolaev:
> There also seem to be scaling issues when rendering SVG data in the following fonts from https://github.com/unicode-org/text-rendering-tests/tree/main/fonts
> 
> * [TestSVGgradientTransform.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGgradientTransform.otf)
> * [TestSVGgzip.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGgzip.otf)
> * [TestSVGmultiGlyphs.otf](https://github.com/unicode-org/text-rendering-tests/blob/main/fonts/TestSVGmultiGlyphs.otf)

Scaling issues have been resolved. Is there anything anything else you'd like me to improve?

Connum:
Can you please resolve the conflicts with the current master in order to do a rebase? Then we can have another look at it! Thanks!

kinolaev:
The conflicts are resolved. I took inspiration from your work and made three changes:
- added `drawSVG` font render option
- added `getSvgImage` method to `Glyph` class
- moved all svg image creation code to `SVGImageManager` class

Connum:
Thank you! I already spotted a few things, but I'll do a full review soon, hopefully in the course of this week.

docs/font-inspector.html
```diff
@@ -261,9 +261,19 @@ <h1>Free Software</h1>
 }
 
 function onFontLoaded(font) {
+    if (window.font) {
+        window.font.onGlyphUpdated = null;
+    }
     window.font = font;
     renderText(font);
     displayFontData(font);
+    font.onGlyphUpdated = (glyphId) => {
```
Connum, line 270: That's an interesting idea using events to trigger redrawing. I would pefer to make this more versatile so we can use it in other places as well (colr/cpal updates, glyph outline changes, etc.) Something like this: https://medium.com/@aryakvn/emitting-events-from-custom-classes-in-javascript-ea635d8023c6

I would normally tend to do that in a separate PR, however, if we provide `onGlyphUpdated` for now, we'd at least have to document it, just to deprecate it when implementing the optimized event handling... On the other hand, it will still be a while until 2.0.0 and we'll probably be able to implement the event handling before release... I'm unsure and would like to get input by @ILOVEPIE @yne

src/svgimages.js
```diff
@@ -0,0 +1,230 @@
...
+            svgDocCacheEntry = createSvgDocCacheEntry(svgBuf);
+            this.cache.set(svgBuf, svgDocCacheEntry);
+        }
+
+        let svgImageCacheEntry = svgDocCacheEntry.images.get(glyphIndex);
+        if (svgImageCacheEntry === undefined) {
+            svgImageCacheEntry = createSvgImageCacheEntry(this.font, svgDocCacheEntry.template, glyphIndex);
+            svgImageCacheEntry.promise.then((svgImage) => {
+                svgImageCacheEntry.image = svgImage;
+                if (typeof this.font.onGlyphUpdated === 'function') {
```
Connum, line 77: See also here regarding the event handling question above @ILOVEPIE @yne

src/path.js
```diff
@@ -415,6 +415,21 @@ Path.prototype.quadTo = Path.prototype.quadraticCurveTo = function(x1, y1, x, y)
     });
 };
 
+/**
+ * @function drawImage
+ * @memberof opentype.Path.prototype
+ */
+Path.prototype.drawImage = function(image, x, y, width, height) {
+    this.commands.push({
```
Connum, line 423: I don't like the idea of handling this as a Path command. The commands should be actual commands for drawing outlines. I'd prefer to handle this spearately, like I did for colr/cpal drawing. Also the outlines shouldn't be drawn below the SVG image at all, rather if a glyph has an SVG (or COLR/CPAL) representation, that should entirely replace the base glyph's outline.

I'd also suggest the following: If a glyph has COLR layers as well as an associated SVG, I'd say we prefer the SVG version, as we currently only support COLRv0 with solid colors. Unless `options.drawSVG` is set to `false` of course. That way, by turning it off, you can force the COLRv0 version, or switch that off as well in order to render the outlines.

src/path.js
```diff
@@ -490,6 +505,14 @@ Path.prototype.getBoundingBox = function() {
                 prevX = startX;
                 prevY = startY;
                 break;
+            case 'I':
```
Connum, line 508: see above

src/path.js
```diff
@@ -526,6 +549,8 @@ Path.prototype.draw = function(ctx) {
             ctx.quadraticCurveTo(cmd.x1, cmd.y1, cmd.x, cmd.y);
         } else if (cmd.type === 'Z' && this.stroke && this.strokeWidth) {
             ctx.closePath();
+        } else if (cmd.type === 'I') {
```
Connum, line 552: see above

src/tables/svg.js
```diff
@@ -0,0 +1,110 @@
...
+    const svgDocumentListOffset = data.byteOffset + offset + p.relativeOffset;
+    const numEntries = p.parseUShort();
+    const svgDocMap = new Map();
+    for (let i = 0; i < numEntries; i++) {
+        const startGlyphID = p.parseUShort();
+        const endGlyphID = p.parseUShort();
+        const svgDocOffset = svgDocumentListOffset + p.parseOffset32();
+        const svgDocLength = p.parseULong();
+        let svgDoc = svgDocMap.get(svgDocOffset);
+        if (svgDoc === undefined) {
```
Connum, line 32: I think "cutting" the svg document out of the buffer would be an excellent candidate for deferred parsing as discussed here: 
https://github.com/opentypejs/opentype.js/pull/698#issuecomment-2046612897
https://github.com/opentypejs/opentype.js/pull/698#issuecomment-2046626397

See here for the code for gvar parsing:
https://github.com/opentypejs/opentype.js/blob/be0d4417a04d92d43178e075273048e926164abf/src/parse.js#L873

Because with many and more complex SVG documents, right now everything would be parsed into memory at once, even though it might not be needed at all, bloating the memory unnecessarily.

Connum (changes requested):
FInished review with a few requested changes.

Could you also please add a TODO comment and console message to
https://github.com/opentypejs/opentype.js/blob/946f2555a2d2d14f232c5c3efe794182d905ac0c/src/glyph.js#L454
similar to https://github.com/opentypejs/opentype.js/blob/946f2555a2d2d14f232c5c3efe794182d905ac0c/src/path.js#L636 for COLR/CPAL?

Because we can't simply output the whole SVG document as it is, we should first sanitize it to make sure it doesn't contain any malicious scripts or features not supported in SVG fonts. (By the way, the latter about unsupported features would also be true for rendering in general, even on canvases, but we don't have to do that in this PR! See https://learn.microsoft.com/en-us/typography/opentype/spec/svg#svg-specification)

docs/glyph-inspector.html
Connum, line 1: Forgot to add in my last review that SVG glyphs are still not drawn in the larger canvas when a glyph is selected from the grid of glyphs on a page.

Connum (changes requested):

kinolaev:
Thanks for the detailed review! I'll try to address all the issues you mentioned by the end of next week.

docs/font-inspector.html
yne: Agree,
I'm in no hurry to move to a centralized event handling and do more lazy parsing. So the current onGlyphUpdated API seems fine

src/path.js
kinolaev: What do you think about making a layer with an SVG image? Something like this
```js
const p = new Path();
if ( options.drawLayers ) {
  const svgImage = this.getSvgImage(font);
  if ( svgImage ) {
      const layer = new Path();
      layer._image = {
          image: svgImage.image,
          x: x + svgImage.leftSideBearing * scale,
          y: y - svgImage.baseline * scale,
          width: svgImage.image.width * scale,
          height: svgImage.image.height * scale,
      };
      p._layers = [layer];
      return p;
  }
  const layers = this.getLayers(font);
  // ...
}
```
Or do you prefer to keep images in a separate array?

src/tables/svg.js
kinolaev: Just to be sure we are on the same page. By "cutting the svg document out of the buffer" you  mean `svgDoc = new Uint8Array(buf, svgDocOffset, svgDocLength);`? As I understand this is a cheap operation, because no memory is copied. So we spend only 24 bytes per document here (`sizeOfPointer(buf) + sizeOfNumber(svgDocOffset) + sizeOfNumber(svgDocLength)`). An SVG table consumes `(sizeOfNumber(glyphId) + sizeOfPointer(svgDoc)) * countOfSVGGlyphs + 24 * countOfSVGDocuments`. Therefore, an SVG table with 1000 SVG documents with one glyph per document would be about 40kb. Is it worth to parse it lazily?
Or did you mean something different by "an excellent candidate for deferred parsing"?

Connum: I'm not too familiar with the internals of js regarding data views and typed arrays, but if the array is really just a pointer to that part of the DataView in memory, then of course that's totally fine!

src/path.js
Connum: I think we could do it that way! We should just make sure that there are separate options for colr/cpal emojis and svg emojis to toggle and that svg is preferred if both exist for a glyph and svg rendering is not disabled

test/fonts/Colortube-lWAd.otf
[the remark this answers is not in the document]
kinolaev: I replaced Colortube font with three fonts from unicode-org/text-rendering-tests you mentioned above.

src/path.js
kinolaev: Done c1bd51e785635739c405fe8de2bc3d03a8460c78

docs/glyph-inspector.html
kinolaev: Done a1d76b31e6e717a02c60344f3f4d6e4dedee9e88

kinolaev:
TODO and warning are added to `Path.prototype.toSVG` method 2d56eb26a1e99f8ca79138210bef8559e8f9141e.
I think I've answered all your comments. Please let me know if I've missed anything.

src/tables/svg.js
kinolaev: Yes, when constructed with an ArrayBuffer as the first parameter, an Uint8Array points to the specified buffer. Here is the relevant doc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#parameters

Connum: I learned something today, thanks! :)

src/util.js
```diff
@@ -90,4 +92,48 @@ function binarySearchInsert(array, key, value) {
...
+function isGzip(buf) {
+    return buf[0] === 31 && buf[1] === 139 && buf[2] === 8;
+}
+
+/**
+ * [GZIP File Format Specification](https://datatracker.ietf.org/doc/html/rfc1952#section-2.3)
+ * @param {Uint8Array} gzip
+ * @returns {Uint8Array}
+ */
+function unGzip(gzip) {
```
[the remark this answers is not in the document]
Connum: We should absolutely keep the fallback, it doesn't hurt!

Connum (approved):
Looking good! I'll add the licenses for the three new test fonts to the LICENSE file in a follow-up PR, so we don't have to take another round.

Connum:
Thanks again for your work and your patience. I'd love to see more contributions in the future!

kinolaev:
Thank you for your time and attention. It was a pleasure to collaborate with you on this task!

Issue #682, opened 2024-03-05
Writing slow (30+ seconds) for certain fonts (CJK, Noto Chinese)


Balearica:
For most fonts, Opentype.js can write a font file using `font.download()` or `font.toArrayBuffer()` in a fraction of a second.  However, certain fonts that include many characters are subject to extreme runtimes.  

This occurs with both Simplified Chinese variants of the Noto family.  Using `NotoSansSC-Regular.ttf` (available [here](https://fonts.google.com/noto/specimen/Noto+Sans+SC)), running `font.toArrayBuffer()` from the console after loading in the [example site](https://opentype.js.org/) took 36 seconds.  Similarly, trying to write `NotoSerifSC-Regular.otf` (available [here](https://fonts.google.com/noto/specimen/Noto+Serif+SC)) running `font.toArrayBuffer()` took 33 seconds.

Not all CJK fonts are this slow--I tested a couple others from Google Fonts that took 4-5 seconds.  However, as I've never experienced anything like this with fonts that only support Latin characters, I'm guessing it's related to the number of glyphs in these fonts.

I will investigate more myself and try and fix in a PR. 

## Your Environment

* Version used: 3fce140bda37e557d2ecb206094428240bfd1615
   * Latest commit as of the time of this writing
* Font used: Noto Sans Simplified Chinese, Noto Serif Simplified Chinese
* Browser Name and version: Chrome
* Operating System and version (desktop or mobile): Windows 10

Balearica:
This ended up being due to excessive re-allocation of memory by using `d = d.concat(...)` rather than `push`.  This is fixed in the PR linked above.

Connum:
Closed via #683

PR #513, opened 2022-05-12
Replaced concat with push for performance

Balearica:
## Description
Replaces instances of `d = d.concat(...)` with `push` within encoding functions for improved performance per #512.  Should not impact results. 

## Motivation and Context
This version is faster, particularly on Firefox.  As a test, I ran `window.font.download()` in the console of the https://opentype.js.org/ site.  In Firefox the change reduced runtime from 1,341ms to 190ms.  Chrome experienced a more modest improvement from 203ms to 166ms.  The ability to quickly generate custom font files is important for my project [Scribe OCR](https://github.com/Balearica/scribeocr), and I'm sure others as well. 

## How Has This Been Tested?
The unit tests pass and I am using this version in my live site.  Additionally, the change is conceptually simple and narrow in scope. 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I did `npm run test` and all tests passed green (including code styling checks).
- [ ] I have added tests to cover my changes.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the **README** accordingly.
- [x] I have read the **CONTRIBUTING** document.

PR #683, opened 2024-03-05
Replaced concat with push for better performance per #682

Balearica:
## Description
Replaces instances of `d = d.concat(...)` with `push` within `cmap` functions for improved performance per https://github.com/opentypejs/opentype.js/issues/682. Should not impact results.  Conceptually speaking, this is almost identical to #513.

Additionally, `t.segments.sort` was being run *every time* a new element was added within a loop; I edited to only run once after all elements are inserted. 

## Motivation and Context
See #682.  In short, the performance of `font.toArrayBuffer()` can be poor, including >30+ second runtimes for certain large fonts.  This PR fixes that.

On my system (Chrome, Windows 10), this PR reduces the runtime for writing NotoSansSC from 36 seconds to 8 seconds (78% reduction), and reduces the runtime for NotoSerifSC from 33 seconds to 8 seconds (76% reduction).  

## How Has This Been Tested?
The unit tests pass and I am using a version with these changes in my project. Additionally, the change is conceptually simple and narrow in scope.

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/cmap.js
+++ b/src/tables/cmap.js
@@ -277,14 +277,14 @@ function makeCmapTable(glyphs) {
     ];
 
     if (!isPlan0Only)
-        cmapTable = cmapTable.concat([
+        cmapTable.push(...[
             // CMAP 12 header
             {name: 'cmap12PlatformID', type: 'USHORT', value: 3}, // We encode only for PlatformID = 3 (Windows) because it is supported everywhere
             {name: 'cmap12EncodingID', type: 'USHORT', value: 10},
             {name: 'cmap12Offset', type: 'ULONG', value: 0}
         ]);
 
-    cmapTable = cmapTable.concat([
+    cmapTable.push(...[
         // CMAP 4 Subtable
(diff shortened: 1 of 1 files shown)


Connum:
Could you add a performance test by measuring the time it takes to run `toArrayBuffer()` and making sure it's below a certain threshold that it was above without the change?

src/tables/cmap.js
```diff
@@ -361,12 +360,22 @@ function makeCmapTable(glyphs) {
     t.entrySelector = Math.log(t.searchRange / 2) / Math.log(2);
     t.rangeShift = t.segCountX2 - t.searchRange;
 
-    t.fields = t.fields.concat(endCounts);
+    for (let i = 0; i < endCounts.length; i++) {
```
Connum, line 363: I was about to refactor this using a single `push()` call combined with the spread operator, but it turns out this is way more performant (see https://jsperf.app/cabimi). That's great!

Connum (approved):
Looking good!

Issue #511, opened 2022-05-02
Support multi-character emoji


forresto:
This issue with monochrome Noto Emoji is distinct from the color emoji issue (#193).

#338 added support for non-Basic-Multilingual-Plane (BMP) characters, but uses `Array.from`, which doesn't account for combined emoji.

It seems that Opentype.js has the glyph information needed, but the initial text-to-glyph translation is the issue:

![image](https://user-images.githubusercontent.com/395307/166242754-65e95f63-79cb-4ad3-8e4c-f87310a8aefc.png)
https://opentype.js.org/glyph-inspector.html

## Expected Behavior

Calling `notoEmojiFont.draw(context, "👨‍👩‍👧‍👦")` should render
![image](https://user-images.githubusercontent.com/395307/166223470-afd9ff31-c504-420b-aa71-b59d949e7d42.png)

## Current Behavior

Calling `notoEmojiFont.draw(context, "👨‍👩‍👧‍👦")` renders
![image](https://user-images.githubusercontent.com/395307/166223510-9ca5f5db-839f-40eb-a450-69527a2d245c.png)

## Possible Solution

0. If "ccmp" is not supported yet and would cover this, this issue can be closed as a duplicate of https://github.com/opentypejs/opentype.js/issues/443.

1. [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) is a native solution, but [isn't supported by Firefox yet](https://caniuse.com/mdn-javascript_builtins_intl_segmenter).

```js
const splitSegmentArray = (string) => Array.from(new Intl.Segmenter().segment(string)).map(x => x.segment);
console.log(splitSegmentArray("😅👨‍👩‍👧‍👦💖👩‍💻💔👩‍🌾🧡👨🏽‍🌾💜🖖🏾🌈"))
```

2. [graphemer](https://github.com/flmnt/graphemer) is a library-based solution. (It is a fairly big library.)

3. [twemoji-parser](https://github.com/twitter/twemoji-parser) is focused on parsing emoji sequences, so it's smaller than graphemer.

## Steps to Reproduce (for bugs)

Live demo: https://gm69qn.csb.app

1. Call `notoEmojiFont.stringToGlyphs("👨‍👩‍👧‍👦")` and get glyphs for "👨👩👧👦" interspersed with the combiner (`"uni200D"`) instead of the one glyph for the combined family.

[...]

Issue #374, opened 2019-02-27
Lets Make a List of ToDos and RoadMap file


bawarkamalqader:
Just wanted to ask if it was possible to make a ToDo list , so that new contributors will be able to know where to start, and what to do , like a ROADMAP.

things like ::
- GPOS Table Writing
- GSUB Subtables Writing : 2,5,6,7,8,9
- others ...

brawer:
Fix [test failures](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/OpenType.js.html)?

bawarkamalqader:
that goes into implementing remaining tables, currently 18 tables are supported, we need to add the remaining tables ::

Remaining Tables  :
 'acnt' , 'ankr' , 'avar' , 'bdat' , 'bhed' , 'bloc' , 'bsln' , 'cvar' , 'cvt ' , 'EBSC' , 'fdsc' , 'feat' , 'fmtx' , 'fond' , 'fpgm' , 'gasp' , 'gcid' , 'gvar' , 'hdmx' , 'just' , 'kerx' , 'lcar' , 'mort' , 'morx' , 'opbd' , 'prep' , 'prop' , 'sbix' , 'trak' , 'vhea' , 'vmtx' , 'xref' , 'Zapf' , 

Important Opentype Fonts
 BASE,CBDT,CBLC,COLR,CPAL,DSIG,EBDT,EBLC,GDEF,HVAR,JSTF,LTSH,MATH,MERG

bawarkamalqader:
we need to sort the remaining tables by priority , i mean the tables that need to be supported earlier and implemented immediately .

brawer:
I'm in no position to define a roadmap, but personally I'd recommend fixing the security bugs first; then fixing the missing parts of cmap; then fixing shaping (GSUB/GPOS). The latter will need GDEF (easy), various barely-documented script-specific shaping engines (painful) plus the Universal Shaping Engine USE. The fontkit project has ported HarfBuzz to JavaScript, so have a look at the fontkit sources. Once all that is done, OpenType.js will be able to correctly render Unicode text.

Personally, I would recommend looking at variable fonts (gvar, hvar, avar, HVAR), color fonts, AAT tables, etc. only after the basics of text rendering are working without bugs. Just my personal two cents of course; others might have different opinions.

Jolg42:
Having a list sorted by priority would be great indeed! Thanks @brawer for your input.

[...]

PR #688, opened 2024-03-22
Feature/ccmp

TonyJR:
Support noto-emoji
## Description
add a new contextChecker for ccmp table in tokenizeText()
add a new substitutionFormat for 'lookup sub_5'

## Motivation and Context
This PR resoved multi-character emoji #511.
noto-emoji uses the 'ccmp' table to solve the 'multi-character for one emoji' problem, which is a 'Language forms' feature that should be executed before all other features. When parsing the noto-emoji font, I found that 'lookup sub_5' could not be parsed correctly, so I added a substitutionFormat to handle this issue.

## How Has This Been Tested?
The unit tests passed and I added a test case for this feature. 

## Types of changes

## Checklist:


Connum:
This looks like a great addition! Please check why the build is failing: https://github.com/opentypejs/opentype.js/actions/runs/8388243929/job/22972963564?pr=688
This PR also implements some additional substitution formats/types, which might conflict with some other outstanding PRs, but we'll have to check.

src/features/applySubstitution.js
```diff
@@ -1,3 +1,11 @@
+/*
```
Connum, line 1: Can you please remove these file headers?

src/features/ccmp/ccmpReplacementLigatures.js
```diff
@@ -0,0 +1,56 @@
+/*
```
Connum, line 1: Can you please remove these file headers?

src/features/ccmp/contextCheck/ccmpReplacement.js
```diff
@@ -0,0 +1,22 @@
+/*
```
Connum, line 1: Can you please remove these file headers?

test/emoji.js
```diff
@@ -0,0 +1,47 @@
+/*
```
Connum, line 1: Can you please remove these file headers?

src/features/applySubstitution.js
TonyJR: oh, of course

TonyJR:
okay, i removed the file header and fixed a code error.

Connum:
There are still some lining issues, can you please resolve them? Then you can run `npm run lint` locally before committing in order to check if everything is OK.

src/features/featureQuery.js
```diff
@@ -169,6 +169,13 @@ function chainingSubstitutionFormat3(contextParams, subtable) {
                         const substitution = lookup(glyphIndex);
                         if (substitution) substitutions.push(substitution);
                     }
+                } else if (substitutionType === '21') {
+                    for (let n = 0; n < inputLookups.length; n++) {
+                        const glyphIndex = contextParams.get(n);
+                        const substitution = lookup(glyphIndex);
+                        if (substitution)
+                            substitutions.concat(substitution);
```
Connum, line 177: `.concat()` returns a new array, which is not stored anywhere here? So this should cause problems with specific substitutions.
We also want to avoid `.concat()` for performance reasons (see #690).

Connum:
Please also add tests for the ligature substitution format 3 and the substitution type 21 (which should catch the mistake I mentioned in the comment I just added to the associated code)

src/features/featureQuery.js
TonyJR: Thank you for your reminder. I readed #690, you are right. I'll fix this.

TonyJR:
> Please also add tests for the ligature substitution format 3 and the substitution type 21 (which should catch the mistake I mentioned in the comment I just added to the associated code)

I'm trying to add tests for ligature with sub_5, it doesn't work. Maybe I need a few days to fix it. This may modify the code related to "liga" and "rlig".
Any suggestions?

TonyJR:
> Please also add tests for the ligature substitution format 3 and the substitution type 21 (which should catch the mistake I mentioned in the comment I just added to the associated code)

I removed substitution type 21 and ligatureSubstitutionFormat3. As an alternative, use chainSubstitutionFormat3 to process the sub5 table. Sub5 may exist in liga and ccmp, both of which are processed using ligatureSubstitutionFormat3. I added test code to ensure they work properly.

src/features/featureQuery.js
Connum: let me resolve this conversation once I did a full review

Connum:
Thanks for your work! Two more things, then we'll hopefully be ready for review:
1. please add the information for the newly added test fonts to the `test/fonts/LICENSE` file (alphabetic order)
2. Instead of adding a new `liga.js` test file, please add the tests for the two new lookup formats to `test/featureQuery.js` in analog form as the existing tests.

Big thanks again!

TonyJR:
image
@Connum Do you have a todo list of the LookupTypes? I want to implement this part of the code, but I am concerned that it may overlap with your work. So, do you have any suggestions?

Connum:
Great to see someone being interested in contributing! There is an issue with an overview of TODOs: https://github.com/opentypejs/opentype.js/issues/374
It might not be up to date, so some things might have been implemented meanwhile. You could also search the code base for instances of "not supported" or "not yet supported".

TonyJR:
> Great to see someone being interested in contributing! There is an issue with an overview of TODOs: #374 It might not be up to date, so some things might have been implemented meanwhile. You could also search the code base for instances of "not supported" or "not yet supported".

Thanks for your suggestion. I'll finsh the format 5.1,5.2,5.3 in this PR, please merge them later.

TonyJR:
> Great to see someone being interested in contributing! There is an issue with an overview of TODOs: #374 It might not be up to date, so some things might have been implemented meanwhile. You could also search the code base for instances of "not supported" or "not yet supported".

Thanks for your suggestion. I'll finsh the format 5.1,5.2,5.3 in this PR, please merge them later.

src/features/ccmp/ccmpReplacementLigatures.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {any} tokens a list of tokens
+ * @param {number} index current item index
+ */
+function getContextParams(tokens, index) {
+    const context = tokens.map(token => token.activeState.value);
+    return new ContextParams(context, index || 0);
+}
+
+/**
+ * Apply Arabic required ligatures to a context range
```
Connum, line 18: Should be "Apply ccmp replacement ligatures to a context range"

src/features/applySubstitution.js
```diff
@@ -30,6 +30,15 @@ function chainingSubstitutionFormat3(action, tokens, index) {
     for(let i = 0; i < action.substitution.length; i++) {
         const subst = action.substitution[i];
         const token = tokens[index + i];
+        if (Array.isArray(subst)) {
+            if (subst.length){
+                // TODO: replace one glyph with multiple glyphs
```
Connum, line 35: Is that something that should be in this PR?

src/features/applySubstitution.js
```diff
@@ -50,14 +59,17 @@ function ligatureSubstitutionFormat1(action, tokens, index) {
     }
 }
 
+
```
Connum, line 62: we don't need this additional line break

src/features/ccmp/ccmpReplacementLigatures.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {ContextRange} range a range of tokens
+ */
+function ccmpReplacementLigatures(range) {
+    const script = 'delf';
```
Connum, line 22: I guess this should be 'DFLT'?

src/features/featureQuery.js
```diff
@@ -435,6 +532,26 @@ FeatureQuery.prototype.lookupFeature = function (query) {
                         }));
                     }
                     break;
+                case '51':
+                    substitution = lookup(contextParams);
+                    if (Array.isArray(substitution) && substitution.length) {
```
Connum, line 537: Are there cases where substitution is not an array or might be empty? Or should that theoretically never happen in a well-formed font and this is just to catch any malformed ones?

src/features/featureQuery.js
```diff
@@ -435,6 +532,26 @@ FeatureQuery.prototype.lookupFeature = function (query) {
...
+                        substitutions.splice(currentIndex, 1, new SubstitutionAction({
+                            id: 51,
+                            tag: query.tag,
+                            substitution
+                        }));
+                    }
+                    break;
+                case '53':
+                    substitution = lookup(contextParams);
+                    if (Array.isArray(substitution) && substitution.length) {
```
Connum, line 547: Same as the previous comment above

src/features/featureQuery.js
```diff
@@ -435,6 +532,26 @@ FeatureQuery.prototype.lookupFeature = function (query) {
                         }));
                     }
                     break;
+                case '51':
```
Connum, line 535: can we get rid of the code duplication between '51' and '53'?

```
                case '51':
                case '53':
                    substitution = lookup(contextParams);
                    if (Array.isArray(substitution) && substitution.length) {
                        substitutions.splice(currentIndex, 1, new SubstitutionAction({
                            id: parseInt(substType),
                            tag: query.tag,
                            substitution,
                        }));
                    }
                    break;
```

test/emoji.js
```diff
@@ -0,0 +1,39 @@
+import assert from 'assert';
```
Connum, line 1: Instead of adding a new `emoji.js` test file, please add these tests to `test/bidi.js`.

Connum (changes requested):
I added a few comments for things that need clarification or fixing.

Connum:
Another thing I noticed: The 'ccmp' tag was already present in the `requiredThaiFeatures`:
https://github.com/opentypejs/opentype.js/blob/6408975ed2c6e91ac8376d8be6290b2f705e1724/test/bidi.js#L114

Should this be present in `requiredArabicFeatures` and `latinFeatures` as well?
https://github.com/opentypejs/opentype.js/blob/6408975ed2c6e91ac8376d8be6290b2f705e1724/test/bidi.js#L25
https://github.com/opentypejs/opentype.js/blob/6408975ed2c6e91ac8376d8be6290b2f705e1724/test/bidi.js#L40

Or is it maybe not needed at all, because the `'ccmp'` is always applied? And if so, does the code make sure of that?

This might also relate to your comment https://github.com/opentypejs/opentype.js/pull/688#issuecomment-2031453250 ? I think you forgot to describe the issue with the language tag for us. Could it be related to the wrong `'delf'` script tag I hinted at in the review comment?

Connum (approved):

PR #692, opened 2024-03-30
Output module file with .mjs extension

DavidJFelix:
Change the output file for ESM to have an `.mjs` file extension. NodeJS says that it recognizes `.mjs` extensions as modules. This will help since `type` is not set in `package.json`. 

## Motivation and Context

I ran into an issue where the only fix was renaming the output module file.
I'm using typescript, vite, vitest and pandacss. While vitest and typescript understand that `.module.js` is ESM, it seems like pandacss is relying on node and it thinks it's `commonjs`. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I did `npm run test` and all tests passed green (including code styling checks).
- [ ] I have added tests to cover my changes.
- [x] My change requires a change to the documentation.
- [x] I have updated the **README** accordingly.
- [ ] I have read the **CONTRIBUTING** document.

Issue #675, opened 2024-02-11
Why did we depricate the load function?


ILOVEPIE:
It's a perfectly good function that's useful.

vinuel:
Came across this too lately and was wondering if there is any explanation or documentation about it...

Connum:
The README was updated to state that the method is deprecated with this commit: https://github.com/opentypejs/opentype.js/commit/88c89dcdb2cc8952da9b270f39506afe2b715efa

The [deprecated] tag was added to the load() test cases in this commit: https://github.com/opentypejs/opentype.js/commit/a9acb24d90737c9432d77df38ede1c32e270c508
Maybe @yne can shed some light on it?

yne:
Hi, the simple reason is that
- loading from Buffer is platform agnostic
- loading from path/URL makes us platform adherent (node/browser/...)

Also, as demonstrated on the README, doing the path/fs to buffer only take 1 line anyway **on the user-side** (using require/import/fs/node:fs/fetch/XHR/net or whatever you need for your env/version/context)

the goal is to keep the lib out of those noises (look at the pain we have just to migrate to a module package.json)

I'd like to avoid that

Connum:
@ILOVEPIE any thoughts in @yne's response?

axkibe:
in my opinion opentype should concentrate itself on bein a typing library and avoid when reasonable doing the fluff around it. doing the fetch() is an example of this, since this also inclused error handling, improving startup speeds by adding preloads to html header, caching etc. let the user of the library do it*.  The same applies on the other side, canvas stroking and filling, there is an argument for opentype making the path in canvas, otherwise we'd redo the same thing over and over translating the opentype.js datastructure into pathing commands.. but the fill and stroke? Leave it to the user, then they can decide on colors, translations etc. without bloating the library.

* If you want we could do "opentype-fluffy.js" or something as another module that does this fluffy things ontop, so people can decide they want to use that or the core only. For example since I'm rendering with opentype.js I have to do caching of the rasterized glyphs as well, which I also would avoid in opentype.js itself (it doesnt yet do it anyway).

TLDR: it was a good/useful function, but there is an argument to be made it isn't in the focus of the lib. It's certainly a one-liner to fetch it yourself (without considering preloading, then its 2 lines).

PR #693, opened 2024-03-31
Switch from Reify to SWC for ensuring compatability with older browsers.

### LINKED ISSUE #619: Should we create a compatibility focused build in addition to the non-esm build?
There is a lot of IOT devices (like smart TVs) that use old versions of WebKit to run web apps, having a build compatible with ES5 would be helpful for people targeting use cases like this. A secondary build pass with closure compiler on the non-esmodule version would be one way of doing this. The reason why I'm bringing this up is that my current project has support as far back as Safari 5.1 if you exclude opentype.js (due to using closure compiler) but is being limited from being compatible with a lot of embedded/smart devices because of opentype.js's compatibility with older browsers. There are other tools that could do the job other than closure compiler but I'm not too familiar with them. What do you guys think?

ILOVEPIE:
## Description
Created a utility that runs esbuild and SWC to build the library.
This supports a compat profile that works as far back as safari 8, maybe earlier.

This closes #619

## Motivation and Context
The library is limited in compatibility by the number of new features we are using, this PR allows us to keep using those features while dropping reify and also supporting even older browsers.

## How Has This Been Tested?
I tested the compat outputs using Browser Stack on multiple old browsers including safari 8 and some ancient versions of chrome. I also tested the other builds using more modern browsers.

## Types of changes

## Checklist:


ILOVEPIE:
@yne does this look OK or do we need to make changes?

yne:
to be honest
- I think it's a step in the wrong direction because I expect the opentype v2 to move to a modern stack with less complexity
- If you want to use older browser, use older opentype.js
- WebKit/apple are well known for being anti-web so I doubt they users (who exactly ?) expect anything
- This commit add even more build script to maintain which I'm fighting against ( eslint-local-rules.js I'm also looking at you)
- This commit remove mocha tests ?!

So no, I won't accept it.

ILOVEPIE:
> to be honest
> - I think it's a step in the wrong direction because I expect the opentype v2 to move to a modern stack with less complexity
> - If you want to use older browser, use older opentype.js
> - WebKit/apple are well known for being anti-web so I doubt they users (who exactly ?) expect anything
> - This commit add even more build script to maintain which I'm fighting against ( eslint-local-rules.js I'm also looking at you)
> - esbuild support transpiling to older ES version https://esbuild.github.io/api/#supported and I don't see any explanation why it is not used 
> - This commit remove mocha tests ?!
> 
> So no, I won't accept it.

The problem is older versions of OpenType.js don't work properly. Also, we already discussed all this in #619.

Also, esbuild does not support transpiling. It only errors when you use a feature that isn't supported by the target. I did a lot of research into this. Also, old versions of Safari are not the only thing targeted by the compat profile. It also supports old versions of Firefox and Chrome and old browsers like Internet Explorer. While you may think there is no use in supporting older browsers, I think there is a large amount of developers out there who work with legacy code who would appreciate being able to use a modern library. Also, do you know how many modern smart TVs are still running old versions of WebKit? Nearly all of them. So if somebody wanted to use the library in a smart TV app, then they're SOL without this. I'm not suggesting we make this the default version of opentype. I'm just saying it's okay to support older browsers for those who need it. And besides, this allows us to drop reify for a dependency that's actually up to date: SWC.

Hell, I'll even take over complete responsibility for maintaining the compat profile if you're that against including the profile.

It's not like I'm asking for Netscape Navigator support.

@Jolg42 & @Connum should weigh in with their opinions before a decision is made.

ILOVEPIE:
> - This commit remove mocha tests ?!

Yeah, that was a mistake. I just saw the word "reify" and removed it. I was pretty tired at the time.

yne:
> Also, esbuild does not support transpiling. It only errors when you use a feature that isn't supported by the target. I did a lot of research into this.

I'm confused, the esbuild documentation state that

> [--target] tells esbuild to transform JavaScript syntax that is too new for these environments into older JavaScript syntax that will work in these environments. For example, the ?? operator was introduced in Chrome 80 so esbuild will convert it into an equivalent (but more verbose) conditional expression when targeting Chrome 79 or earlier.

So I would have expected some kind of rational/justification explaining why esbuild was not the chosen solution.

I'm not against allowing users to build a "compat" version of opentype.js. I'm against complexifing our codebase/buildflow for that.
if it would only have required a `--target=es5` to the esbuild I wouldn't have minded.

yne:
also, the increasing number of variability that we have to test may require some re-focus

- esm/cjs
- low-memory
- min
- legacy

sum up to 16 possible build profile combination to test

maybe low memory and legacy are for the same purpose ?

I hope that even if we disagree on this subject you'll still ping me some other times 😀

ILOVEPIE:
> 
> > Also, esbuild does not support transpiling. It only errors when you use a feature that isn't supported by the target. I did a lot of research into this.
> 
> I'm confused, the esbuild documentation state that
> 
> > [--target] tells esbuild to transform JavaScript syntax that is too new for these environments into older JavaScript syntax that will work in these environments. For example, the ?? operator was introduced in Chrome 80 so esbuild will convert it into an equivalent (but more verbose) conditional expression when targeting Chrome 79 or earlier.
> 
> So I would have expected some kind of rational/justification explaining why esbuild was not the chosen solution.
> 
> I'm not against allowing users to build a "compat" version of opentype.js. I'm against complexifing our codebase/buildflow for that.
> if it would only have required a `--target=es5` to the esbuild I wouldn't have minded.

esbuild does transform some syntax but it's a very limited transform and it's not as thorough as SWC. Also SWC does a code optimization pass to make the code smaller and more performant by rearranging elements of the code.

Also:
> Note that this is only concerned with syntax features, not APIs. It does not automatically add polyfills for new APIs that are not used by these environments. You will have to explicitly import polyfills for the APIs you need (e.g. by importing core-js). Automatic polyfill injection is outside of esbuild's scope.

SWC does automatic ponyfill injection. So for example we have usages of map that are not supported in older browsers because map was introduced later on. SWC fixes this automatically by injecting a ponyfill of Array.prototype.map that gets used if the function is not available in the browser. It also only injects the ponyfills needed to target a specific browser version at a minimum. So what it does is it only injects the minimum number of ponyfills needed for that.

ILOVEPIE:
The reason I was talking about the erroring was because you originally only linked the supported flag, which is not the same as the target flag.

Connum:
I actually like this PR, and as @ILOVEPIE stated, we already had a discussion about it. Getting rid of reify will finally allow us to use modern syntax and language features, such as private class fields and methods. iirc it was also holding us back from switching upgrading to the most recent version of mocha. So it actually *is* a step to more modern development. SWC is also a modern build tool, and many active projects provide legacy versions (may using a "hail Mary" approach and not even running the tests on them again) so I frankly don't get that argument at all. 😄

I don't really care about increased build times, and the complexity is no really increasing that much. There are still so many projects relying on webpack, babel and a great number of plugins for them... So what we'd have now is already way better than that.

Regarding the local eslint rules you mentioned: these are for specific performance improvements that we found made an impact on the library, and there were no existing packages for them. You write them once, nothing to maintain except extending them maybe, they're only used for lining, so I don't really see an issue with that either. I think it's even better to have control over them instead of having to rely on yet another third party package.

So I'm very happy to see this happening and I'm all for it!

esbuild-runner.mjs
```diff
@@ -0,0 +1,217 @@
...
+        "root": Path.dirname(params.options.input??params.options.args[0]),
+        "sourceMaps": params.options.sourcemap??false,
+        //"inlineSourcesContent": true,
+        "isModule": true,
+        "minify": params.options.minify??false
+    };
+    let esbuildTarget;
+    {
+        //These are the targets for a compatibility build that should work with all browsers in use today, even in embedded devices.
+        const compatTargets = {
```
Connum, line 152: Maybe we could put these into package.json and retrieve them from there, so it will be easier to maintain, as we'll probably forget about this file if we don't have to touch it very often.

Connum:
Furthermore, I strongly believe that built complexity or maintenance never was and never will be the reason why opentype.js development is slow. It's our work and private lifes and the lack of sponsorship. 😉

yne:
If this allows removing the barrier on mocha then I'll be for the best. :heavy_check_mark: 

Is the Closure Compiler externs file still relevant once transpiling will work ?

Echoing to https://github.com/opentypejs/opentype.js/issues/675 and the last line of https://github.com/opentypejs/opentype.js/pull/693#issuecomment-2031000706 I think one day we shall have a clear view of what opentype.js vision is: "sleek & modern" or "battery-included works-everywhere " (kind of a mac vs windows)

package.json
```diff
@@ -24,23 +24,26 @@
...
-    "b:esm": "esbuild --bundle src/opentype.js --outdir=dist --external:fs --external:http --external:https --target=es2018 --format=esm  --out-extension:.js=.module.js",
-    "d:esm": "esbuild --bundle src/opentype.js --outdir=dist --external:fs --external:http --external:https --target=es2018 --format=esm  --out-extension:.js=.module.min.js --minify --sourcemap"
+    "start": "node esbuild-runner.mjs --input src/opentype.js --output dist/opentype.module.js --externals \"['fs','http','https']\" --target es2015 --module --watch --servedir . --global-name opentype --footer \"(function (root, factory) { if (typeof define === 'function' && define.amd)define(factory); else if (typeof module === 'object' && module.exports)module.exports = factory(); else root.opentype = factory(); })(typeof self !== 'undefined' ? self : this, function(){return ((function(a,b){var keys=Object.keys(b);for(var i=0;i<keys.length;i++)a[keys[i]]=b[keys[i]];return a})({'default':opentype},opentype));});\"",
```
yne, line 32: why are fs/http/https back in town ?

ILOVEPIE:
> If this allows removing the barrier on mocha then I'll be for the best. :heavy_check_mark: 
> 
> Is the Closure Compiler externs file still relevant once transpiling will work ?
> 
> Echoing to https://github.com/opentypejs/opentype.js/issues/675 and the last line of https://github.com/opentypejs/opentype.js/pull/693#issuecomment-2031000706 I think one day we shall have a clear view of what opentype.js vision is: "sleek & modern" or "battery-included works-everywhere " (kind of a mac vs windows)
> 
> 

the two approaches aren't mutually exclusive. That closure compiler externs file is outdated and needs fixes, it basically tells google closure compiler what our API is so it doesn't get mangled or optimized away when people using closure compiler use opentype.js. Closure compiler is like SWC's less hip but way stronger older brother. I opted for SWC because it's newer but that doesn't mean our users will be using it over closure compiler.

yne:
> > * This commit remove mocha tests ?!
> 
> Yeah, that was a mistake. I just saw the word "reify" and removed it. I was pretty tired at the time.

@ILOVEPIE : I've cloned your branch and re-enabled unit-tests (so without --reify) and it failed.

Did you created the PR without running the unit-test ? Or where you using a solution that you didn't commited ?

If the later (which I really hope), could you share it because the only solution we had for now was to simply declare `module:true` in package.json but you [refused it](https://github.com/opentypejs/opentype.js/pull/579#issuecomment-1649105104) as too modern (because it require NodeJS v15.3 from 2020?)

Also, I setup SWC from scratch by
- adding `"@swc/cli": "^0.3.12"` in `"devDependencies"`
- adding `"compat": "swc -C jsc.target=es3 ./dist/opentype.js -o ./dist/opentype.compat.js",` in `script`

and it worked.

So what was the motivation for creating all those swc wrapper and why was this wrapper even used on non compat builds ?

ILOVEPIE:
There was a few issues with the PR. I'm going to fix them later today. I was pretty tired when I finalized the PR and didn't fully take everything into account. The reason why I have the wrapper is because it gives us finer control over the targeting of the builds and also allows us to specifically enable or disable features of SWC that are causing issues in a specific build etc. The solution for fixing the tests I'm probably going to go with is renaming the source files to .mjs because that should let Mocha run them properly. Unless you want me to make a wrapper that runs mocha (just kidding). Also SWC assumes that it's output will be inside a IIFE or module, which means it's better to apply it before bundling.

ILOVEPIE:
Also the reason why it was used on non-compat builds was because that way the code would be ES6 compatible even if new features are added and we use them. For example, the spread operator is not ES6 compatible, but we can use it with SWC and still be ES6 compatible.

ILOVEPIE:
This also closes #692 as it required reworking the same stuff.

ILOVEPIE:
I fixed the issues, @yne if you want to take a look, this PR is now using latest mocha.

package.json
```diff
@@ -61,7 +68,7 @@
     ],
     "extends": "eslint:recommended",
     "parserOptions": {
-      "ecmaVersion": 2018,
+      "ecmaVersion": 2023,
```
yne, line 71: at this point we may as well use "latest" or just nothing (default=latest)
so we won't have to update it again

yne:
glad we finaly moved to module format, but I think that because of all those renaming, all other PR will now be broken ?

`loadFromFile`/`loadFromUrl` are getting more an more uggly everyday...
Can't we just remove them someday ? (for example: today)

I would really prefere using the official `@swc/cli` wrapper that having to read/understand thoses 500+ line of build scripts (buildLib/commandLineHelper.mjs buildLib/esbuild-plugin-swc.mjs esbuild-runner.mjs) which I did not btw.

Connum:
> but I think that because of all those renaming, all other PR will now be broken ?

I think (or maybe it's more kind of a hope) that git will be clever enough so that they can be easily rebased, however, any added files in a PR and their imports will have to be renamed

yne:
> > but I think that because of all those renaming, all other PR will now be broken ?
> 
> I think (or maybe it's more kind of a hope) that git will be clever enough so that they can be easily rebased, however, any added files in a PR and their imports will have to be renamed

Yes let's hope :crossed_fingers: 

Since https://github.com/opentypejs/opentype.js/issues/619 did not mention `tsc`, I'll exploring it use as JS transpiler (default to es5, which shall work on @ILOVEPIE smart toaster :smirk:)

yne:
Just realized by looking at https://github.com/opentypejs/opentype.js/pull/695 that all HTML page importing `opentype.module.js` shall also be updated

ILOVEPIE:
> glad we finaly moved to module format, but I think that because of all those renaming, all other PR will now be broken ?
> 
> `loadFromFile`/`loadFromUrl` are getting more an more uggly everyday...
> Can't we just remove them someday ? (for example: today)
> 

I wouldn't remove them in 2.0.0. I would remove them in 2.0.1 or 2.1.0. Also, we didn't move to the module format. All I did was rename the existing module file names to .mjs. The thing is though I kept the .module.js option for people who are still depending on that.

ILOVEPIE:
> Just realized by looking at https://github.com/opentypejs/opentype.js/pull/695 that all HTML page importing `opentype.module.js` shall also be updated

This PR doesn't require that, as opentype.module.js still exists. It's just deprecated now.

ILOVEPIE:
> > > but I think that because of all those renaming, all other PR will now be broken ?
> > 
> > I think (or maybe it's more kind of a hope) that git will be clever enough so that they can be easily rebased, however, any added files in a PR and their imports will have to be renamed
> 
> Yes let's hope :crossed_fingers: 
> 
> Since https://github.com/opentypejs/opentype.js/issues/619 did not mention `tsc`, I'll exploring it use as JS transpiler (default to es5, which shall work on @ILOVEPIE smart toaster :smirk:)

The problem with using TSC is that it doesn't work with JavaScript files. We'd have to translate everything to TypeScript. Also, another thing is that the output, I think, only goes as far back as ES5.

ILOVEPIE:
> > but I think that because of all those renaming, all other PR will now be broken ?
> 
> I think (or maybe it's more kind of a hope) that git will be clever enough so that they can be easily rebased, however, any added files in a PR and their imports will have to be renamed

That's specifically why I used git mv instead of the regular move command.

yne:
> The problem with using TSC is that it doesn't work with JavaScript files. 

it does. i've used it. But you are correct on ES5 part. It wont transpile to older standards.
but it seems the new "compat" opentype targets are also ES5 so...

https://github.com/opentypejs/opentype.js/pull/693/files#diff-84f0a237b3664b42983eec0d5925a9529eb52beee9aab70776b7c21cfdb0cd65R169

yne:
> I wouldn't remove them in 2.0.0. I would remove them in 2.0.1 or 2.1.0.

I would replace them with a deprecation/migration notice/console.error for 2.0.0 , then remove them in 2.1.0

that's what majors are for,
you dont break API in minor/micro

ILOVEPIE:
> 
> > I wouldn't remove them in 2.0.0. I would remove them in 2.0.1 or 2.1.0.
> 
> I would replace them with a deprecation/migration notice/console.error for 2.0.0 , then remove them in 2.1.0
> 
> that's what majors are for,
> you dont break API in minor/micro
> 

Sorry, I think there was a miscommunication. That's exactly what I was saying.

ILOVEPIE:
> 
> > The problem with using TSC is that it doesn't work with JavaScript files. 
> 
> it does. i've used it. But you are correct on ES5 part. It wont transpile to older standards.
> but it seems the new "compat" opentype targets are also ES5 so...
> 
> https://github.com/opentypejs/opentype.js/pull/693/files#diff-84f0a237b3664b42983eec0d5925a9529eb52beee9aab70776b7c21cfdb0cd65R169
> 
> 

No, there are some compatibility targets listed there that are not ES5 but rather have partial support for ES5.

Connum:
I'd suggest we wait till the SVG support PR is merged, then bring this PR up-to-date and take care of the file renaming of files that have meanwhile been added via several PRs. Then we should get this merged quite soon, so we don't have to do that over and over again. Your thoughts, @yne @ILOVEPIE?

ILOVEPIE:
> I'd suggest we wait till the SVG support PR is merged, then bring this PR up-to-date and take care of the file renaming of files that have meanwhile been added via several PRs. Then we should get this merged quite soon, so we don't have to do that over and over again. Your thoughts, @yne @ILOVEPIE?

Sounds like a plan to me.

yne:
sure, the js->mjs renames may even be done in it specific PR that could be merged instantaneously,
so we don't mixup rename / reify-drop / browser-support in the same PR

yne:
I succeed (more or less) to make it run on Firefox 4 (2011) using the following build script

```json
"scripts": {
  "comp": "tsc --allowJs dist/opentype.js --outFile dist/opentype.compat.js"
}
```
However It lack the following classes `Proxy`, `WeakMap`, `DataView`, `Array.includes`, `Array.from` so I had to polyfill/mock them but in the end it works (at least it extract the name table):

image


```html
<script src="dist/opentype.compat.js"></script>
<script>
    var xhr = new XMLHttpRequest(); // member when no fetch() ? I member.
    xhr.open('GET', 'test/fonts/TestHVAROne.otf', true); // python -m http.server it
    xhr.responseType = 'arraybuffer';
    xhr.onload = function(e) {TEST=opentype.parse(this.mozResponseArrayBuffer)};
    xhr.send();
</script>
```

ILOVEPIE:
> I succeed (more or less) to make it run on Firefox 4 (2011) using the following build script
> 
> ```json
> "scripts": {
>   "comp": "tsc --allowJs dist/opentype.js --outFile dist/opentype.compat.js"
> }
> ```
> 
> However It lack the following classes `Proxy`, `WeakMap`, `DataView`, `Array.includes`, `Array.from` so I had to polyfill/mock them but in the end it works (at least it extract the name table):
> 
> image
> 
> ```
> <script src="dist/opentype.compat.js"></script>
> <script>
>     var xhr = new XMLHttpRequest(); // member when no fetch() ? I member.
>     xhr.open('GET', 'test/fonts/TestHVAROne.otf', true); // python -m http.server it
>     xhr.responseType = 'arraybuffer';
>     xhr.onload = function(e) {TEST=opentype.parse(this.mozResponseArrayBuffer)};
>     xhr.send();
> </script>
> ```
> 
> see [9fc86d4](https://github.com/opentypejs/opentype.js/commit/9fc86d4b34c15d0eb9e73b9f762c1d591b096369)

- tsc is not as good at optimization as swc is
- tsc is not designed for this like swc is
- tsc doesn't do ponyfill injection like swc
- tsc is objectively slower at this than swc

Connum:
@ILOVEPIE Now that #703 has been merged, do you want to rework this PR or abandon it alltogether?

ILOVEPIE:
I'll rework it.

ILOVEPIE:
@Connum @yne This PR has been reworked.

ILOVEPIE:
@Connum @yne When either of you get the chance can you do a review of this PR?

yne:
when this PR will be rebased :)

ILOVEPIE:
> when this PR will be rebased :)

It's already been re-based. The only thing that hasn't been updated is the package JSON.

yne:
I'm asking because there are 7300+ line added in 55 files... which is not just the kind of "I've added 5 line in package.json to build with reify" PR I was expecting...

All thoses changes are mandatory for reify to work ?

ILOVEPIE:
> I'm asking because there are 7300+ line added in 55 files... which is not just the kind of "I've added 5 line in package.json to build with reify" PR I was expecting...
> 
> All thoses changes are mandatory for reify to work ?

I'll go back through it and try and simplify, but we're not adding reify. We're removing reify and adding support for SWC for version compatibility.

Connum:
I don't understand why the mjs files are listed as completely new and there are no deletions of their counterparts listed (which should be evaluated as renaming anyway). I'm on my phone though, so maybe I missed something.

Also the build fails, which needs to be fixed before we can start the review.

yne:
> > I'm asking because there are 7300+ line added in 55 files... which is not just the kind of "I've added 5 line in package.json to build with reify" PR I was expecting...
> > All thoses changes are mandatory for reify to work ?
> 
> I'll go back through it and try and simplify, but we're not adding reify. We're removing reify and adding support for SWC for version compatibility.

Sorry I was tired, I meant SWC.
What a week huh?

ILOVEPIE:
Should be fixed now.

yne:
There is still a new `test/tables/gasp.mjs` which shall be `test/tables/gasp.spec.mjs`

package.json
```diff
@@ -24,21 +27,37 @@
...
+    "b:esm": "node esbuild-runner.mjs --input src/opentype.mjs --output dist/opentype.mjs --externals \"['fs']\" --target es2015 --module",
+    "d:esm": "node esbuild-runner.mjs --input src/opentype.mjs --output dist/opentype.min.mjs --externals \"['fs']\" --target es2015 --module --minify --sourcemap",
+    "b:compat": "node esbuild-runner.mjs --input src/opentype.mjs --output dist/opentype.compat.js  --externals \"['fs']\" --target compat --global-name opentype --footer \"(function (root, factory) { if (typeof define === 'function' && define.amd)define(factory); else if (typeof module === 'object' && module.exports)module.exports = factory(); else root.opentype = factory(); })(typeof self !== 'undefined' ? self : this, function(){return ((function(a,b){var keys=Object.keys(b);for(var i=0;i<keys.length;i++)a[keys[i]]=b[keys[i]];return a})({'default':opentype},opentype));});\"",
```
yne, line 40: so `bun`, `deno` et al. cannot be used anymore to build opentype.js ?

yne: If I understand correctly, `b:compat` is the only target that require a post processing via SWC. If that's the case, maybe we could keep the other target as they were (ie. not limited to node) and only use the `node esbuild-runner.mjs` for the b:compat target ?

esbuild-runner.mjs
```diff
@@ -0,0 +1,327 @@
...
+        shortKey: false,
+        propertySyntax: false,
+        default: false
+    }
+});
+const params = parseArgs();
+
+async function build(params) {
+    const swcConfig = {
+        "jsc": {
```
yne, line 172: could this be in a /.swcrc file ?

buildLib/commandLineHelper.mjs
```diff
@@ -0,0 +1,234 @@
+import process from 'node:process';
+
+/**
+ * @file This file handles parsing command line arguments for any custom build tools used in building opentype.js
+ */
+
+export default function createCommandLineArgumentsHandler(validArguments) {
```
yne, line 7: if I understand correctly, SWC use JSON as argument format,
and since there is no official CLI wrapper, we do the --arg => JSON conversion with this mjs file

If that's the case, my question is : could we store those arguments (as JSON) in our package.json (not far from the the builds steps) so we could have a more direct control on SWC (without having to be limited by this intermediate CLI values representation)

package.json
ILOVEPIE: > If I understand correctly, `b:compat` is the only target that require a post processing via SWC. If that's the case, maybe we could keep the other target as they were (ie. not limited to node) and only use the `node esbuild-runner.mjs` for the b:compat target ?

The other ones do use processing via SWC to keep them compatable with es2015 even if we include more advanced features from later ECMAScript versions.

esbuild-runner.mjs
ILOVEPIE: > could this be in a /.swcrc file ?

No, the problem with doing that is the config needs to be changed for certain files, ran into that issue just now which is why I had to push an amended commit again.

buildLib/commandLineHelper.mjs
ILOVEPIE: > if I understand correctly, SWC use JSON as argument format, and since there is no official CLI wrapper, we do the --arg => JSON conversion with this mjs file
> 
> If that's the case, my question is : could we store those arguments (as JSON) in our package.json (not far from the the builds steps) so we could have a more direct control on SWC (without having to be limited by this intermediate CLI values representation)

No, this is not an arguments to json converter for SWC. This validates arguments passed to esbuild-runner.mjs to confirm they're valid (using a config passed to it from esbuild-runner.mjs) and generates a json object representing them that is more easily used from the code in esbuild-runner.mjs.

The purpose of esbuild-runner is to generate the correct SWC and ESBuild configs for the target specified.

I did move the configuration for the compat target to the package.json

package.json
ILOVEPIE: > so now that we explicit `node` in our command, `bun`, `deno` et al. cannot be used anymore to build opentype.js ?

can't they just do this?:
```bash
alias node deno
npm run build
```

ILOVEPIE:
Finally i got the missing dependency, polyfill issues, tests and eslint errors fixed and working.

esbuild-runner.mjs
ILOVEPIE: > Maybe we could put these into package.json and retrieve them from there, so it will be easier to maintain, as we'll probably forget about this file if we don't have to touch it very often.

done.

package.json
ILOVEPIE: > why are fs/http/https back in town ?

This has been fixed now I think.

esbuild-runner.mjs
yne: i was thinking more of something like
```js
const swcConfig={...readFromPackage(), otherDynamicArg}
```
but nevermind.

package.json
yne: understood, so all the target goes through SWC to be compatible with old browser.
in this case, what is the point of b:compat ?

yne (dismissed):
I'm still unsure about some part especially the coreJs part and the argument parsing, but that's could be improved later (if any)

but the overall gain (being able to use any ES feature without being scared of breaking some old browser/node) worth it.

test/tables/gasp.mjs
```diff
@@ -0,0 +1,32 @@
+import assert from 'assert';
```
yne, line 1: bad rebase: this file is gasp.spec.mjs now

ILOVEPIE:
There are a few code cleanup changes that I'd like to make before we actually go through with this.

test/tables/gasp.mjs
ILOVEPIE: OK, will fix.

ILOVEPIE:
I will probably get some time to make those code cleanup changes on Saturday. I"ve been a bit busy between work and preparing to give a workshop on taking over control of other running processes in both Windows and Linux at DEFCON 32 in Las Vegas.

ILOVEPIE:
@Connum @yne we should jump on this again.

yne:
maybe the state of the art has changed since then (2 years is an eternity in the JS world 😏)

and what is an "older browser" two years later might have become obsolete browser two years later

ILOVEPIE:
> maybe the state of the art has changed since then (2 years is an eternity in the JS world 😏)
> 
> and what is an "older browser" two years later might have become obsolete browser two years later

This is irrelivant, it will let us use the latest tech, and be able to target any point in the past, regardless of when. An "obsolete browser" can be changed for a "older browser" at any time.

PR #695, opened 2024-04-06
COLR/CPAL color glyph rendering

Connum:
## Description

1. Implements calculation and drawing of color glyph layers in fonts using the COLRv0/CPALv0 format. 
* Collects and calculates the path data for all the layers in a color glyph when calculating the path for drawing
* Provides higher-level methods instead of having to modify the table data manually, by providing a `LayerManager` object in `Font.layers` and a `PaletteManager` in `Font.palettes`
2. Includes some performance optimizations needed due to the complexity added by supporting layers
* Providing methods for binary search and binary insertion in `util.js`
* Not calling `path.close()` if a path is rendered without a stroke, as filling a path will automatically close the path anyway, resulting in an unnecessary calculation. `path.close()` was a big factor for long computation times when analyzing performance using the browser dev tools.
**Note:** The more complex the supported features become (color glyphs, in the future hopefully font variation support), the worse the drawing performance gets. We will definitely have to implement some optimizations like caching in the future, or maybe even going full-async, in order to reduce computations and therefore increasing performance.
3. Adds some testing functionality by providing utility functions in `testutil.js`
* `createMockObject()`: creates a mock object for any browser element or API that we couldn't test in node otherwise (e.g. canvas contexts). The logs array includes all the calls and property accesses on the object, then it can be tested if the methods and parameters match expected operations.
* `enableMockCanvas()` enables testing canvas functionality that would normally only work in browser context (creating a mock canvas element, CSS color name conversion)
4. Extends the Glyph Inspector to test and showcase the new functionality:
* Font data will be kept in session storage, so that reloading the page (e.g. automatically via HMR) will preserve the last loaded font. This made development and testing way faste for me.
* Made several CSS & JS bugfixes
* Show the palettes available in the font
* Allow changing the used palette with immediate effect
* Very basic palette modification: adding/deleting palettes, adding new color fields, changing colors
* Display layer information and their glyphs
5. Made some updates to the README while I was adding the new APIs

## Motivation and Context

#490 added parsing and writing support for both tables, but actually getting the paths and colors for a color glyph and drawing it was not implemented. Manipulating the layers and palettes is quite complex, and developers using opentype.js would all have had to implement similar functionality in order to make use of the data. With this PR it works out-of-the-box.

## How Has This Been Tested?

I added fairly extensive tests that should cover everything.

image
image

## Types of changes

## Checklist:


docs/font-inspector.html
```diff
@@ -231,6 +243,26 @@ <h1>Free Software</h1>
...
+            const palettes = font.palettes.getAll('hexa');
+            if ( palettes.length ) {
+                palettes.forEach((palette, idx) => {
+                    markup += `<dt><strong>↪ Palette ${idx}</strong></dt><dd><div class="color-swatches">`;
+                        palette.forEach(color => {
+                            markup += `<span class="color-swatch" style="background:${color}" title="${color}"></span>`;
+                        });
+                    markup += `</div></dd>`;
+                });
+                cpal.innerHTML += markup;
```
yne, line 261: `inneHMTL` could be the result of a double .map()

```js
cpal.innerHTML += palettes.map((palette, idx) =>
    `<dt><strong>↪ Palette ${idx}</strong></dt><dd><div class="color-swatches">`+
    palette.map(color => `<span class="color-swatch" style="background:${color}" title="${color}"></span>`).join('')+
    `</div></dd>`);
});
```

docs/font-inspector.html
```diff
@@ -231,6 +243,26 @@ <h1>Free Software</h1>
             element.innerHTML = '<dt>' + Object.keys(font.kerningPairs).length + ' Pairs</dt><dd>' + JSON.stringify(font.kerningPairs) + '</dd>';
         }
     }
+
+    if(font.tables.cpal) {
+        let markup = ' ';
+        const cpal = document.getElementById("cpal-table");
+        const dt = cpal.querySelector('dt');
```
yne, line 250: Why check for existing dt ? Can't we directly update the parent <dl> in any case ?

docs/font-inspector.html
Connum: Nice simplification indeed!

docs/glyph-inspector.html
```diff
@@ -339,8 +384,117 @@ <h1>Free Software</h1>
...
+
+function updateColrOptions() {
+    const palettes = window.font.palettes.getAll('hexa');
+    const palettesDiv = document.getElementById('palettes');
+    form.drawLayers.checked = window.fontOptions.drawLayers;
+    palettesDiv.innerHTML = '';
+    let markup = '';
+    if ( palettes.length ) {
+        palettes.forEach((palette, idx) => {
+            markup += `<dt><label><input type="radio" name="usePalette[]" oninput="window.fontOptions.usePalette = parseInt(this.value); window.redraw();" value="${idx}" ${idx === window.fontOptions.usePalette ? ' checked' : ''}><strong>↪ Palette ${idx}</strong></label> <button type="button" class="color-swatch" onclick="window.font.palettes.delete(${idx}); redraw();">❌</button></dt><dd><div class="color-swatches">`;
```
yne, line 449: At this point we shall move to a `Element` builder like
```js
const el = (tag, props={}, ch=[]) => ch.reduce((e,c) => (e.appendChild(c),e),Object.assign(document.createElement(tag),props))
```

so this long HTML string would become this (also long) declaration
```js
el("dt",{},[
    el("label",{},[
        el("input",{ type:"radio", name:"usePalette[]", value:idx,
            checked:idx === fontOptions.usePalette,
            oninput(){
                fontOptions.usePalette = parseInt(this.value);
                redraw();
            }
        }),
        el("strong",{innerText:`↪ Palette ${idx}`})
    ]),
    el("button", {innerText:"❌",type:"button",classList:"color-swatch", onclick(){
        font.palettes.delete(idx);
        redraw();
    }})
])
```

This prevent any HTML syntax error (which is not obvious to see in string form) Also, this prevent any HTML injection

docs/glyph-inspector.html
```diff
@@ -410,8 +564,19 @@ <h1>Free Software</h1>
...
     try {
         const data = await file.arrayBuffer();
-        onFontLoaded(opentype.parse(isWoff2 ? Module.decompress(data) : data));
-        showErrorMessage('');
+        let error;
+        try {
+            const base64String = arrayBufferToBase64(data);
+            sessionStorage.setItem('fontData', base64String);
+        } catch (err) {
+            if ( err.message === 'Maximum call stack size exceeded' ) {
```
yne, line 572: Is this Exception `message` the same on all browser/JSengine ?

src/layers.js
```diff
@@ -0,0 +1,223 @@
+import { binarySearch, binarySearchIndex, binarySearchInsert } from './util.js';
+
+export class LayerManager {
+    // private properties don't work with reify
+    // @TODO: refactor once we migrated to ES6 modules, see https://github.com/opentypejs/opentype.js/pull/579
```
yne, line 5: even after migrating to another transpiler we will never get real "private" attribute since our code has to be dumbed-down to ES5 (or lower?) :neutral_face:

src/path.js
```diff
@@ -623,6 +633,13 @@ Path.prototype.toPathData = function(options) {
  * @return {string}
  */
 Path.prototype.toSVG = function(options, pathData) {
+    if (this._layers && this._layers.length) {
```
yne, line 636: Maybe some inspiration: https://yne.fr/picon/list?flags
Code: https://github.com/yne/picon/blob/master/edit.html

src/tables/cpal.js
```diff
@@ -44,5 +50,263 @@ function makeCpalTable({ version = 0, numPaletteEntries = 0, colorRecords = [],
...
+    ) {
+        return color;
+    } else if (typeof color === 'object') {
+        if (targetFormat == 'bgra') {
+            return color;
+        }
+        if (returnRaw) {
+            return bgraToRaw(color);
+        }
+    } else if(!isRaw && /^#([a-f0-9]{3}|[a-f0-9]{4}|[a-f0-9]{6}|[a-f0-9]{8})$/.test(color.trim())) {
```
yne, line 183: Uppercase are forbidden ? `[a-fA-F0-9]`

test/fonts/OpenMojiCOLRv0.ttf
yne, line 1: Can't we find a smaller color font ?
Sure this one is funny, but it is also larger than all 17 other test font combined

docs/glyph-inspector.html
Connum: Sure, let's do that in the future. We should also get rid of the code duplication across the different html files' script blocks, by moving the code (or at least the duplicated code) to a common file, just like the css. But I don't want to bloat this PR more than it already is.

docs/glyph-inspector.html
Connum: Good point, probably not. Thinking about it, I think it was caused by an error in the code and didn't occur since, so we might not need this at all.

src/layers.js
Connum: Well, we might distribute different flavours in the future, and one of them could be a pure ES6 build.

src/tables/cpal.js
Connum: Good catch, forgot the i modifier. I'll add a test for this as well.

test/fonts/OpenMojiCOLRv0.ttf
Connum: I'll see if I can make a subset with just a couple of glyphs.

src/path.js
Connum: Yeah, maybe we should just implement it in a way we see fit, and then adapt the format via the bin script, once the guys over at the unicode test suite decided on a format they'd like to have.

src/tables/cpal.js
Connum: fixed

docs/font-inspector.html
Connum: Implemented. It actually needed a second `.join('')` at the and for multiple palettes to look right.

docs/font-inspector.html
Connum: I just checked: The headlines for alle tables we output are always there, but the content of the cell may be `"Undefined"`, so the `<dt>` might not always be there.

docs/glyph-inspector.html
Connum: Implemented as a general error message now.

test/fonts/OpenMojiCOLRv0.ttf
Connum: Replaced the full font with a subset of glyphs specifically for the tests.

yne (approved):

PR #699, opened 2024-04-11
Implement variable font rendering (gvar, avar, hvar, cvar)

Connum:
This PR implements rendering of variable fonts.

## Description

* Docs (index and Glyph Inspector): Selection of predefined instances from the fvar table, all axes available as range sliders, render variations instantly. Some other improvements. 
* supports gvar point transformation
* supports avar point transformation
* supports hvar advanceWidth and leftSideBearing transformation
* support for variation coordinates in the rendering options
* able to set default variation for the whole font
* works with hinting and cvar table
* adapted test render script to have less tests fail due to minor path differences

**TODOs before this is ready for review:**
- [x] look into the [bug](https://rawgit.com/unicode-org/text-rendering-tests/master/reports/fontkit.html#GVAR-9) that the fontkit implementation has as well
- [X] support advanceWidth transform
- [x] see if we can easily support CFF2 blend transforms
- [X] see if we can easily support hvar transform
- [x] see if it makes sense to support cvar as well
- [x] Check toSVG results
- [x] run unicode test suite
- [x] try to fix more unicode tests
- [x] Add gvar tests, including those from the unicode test suite
- [x] Add avar tests
- [X] Add hvar tests
- [X] Add cvar glyph tests
- [x] Add cvar tests
- [x] Update API documentation in README

**TODOs for future PRs:**
- add cvar writing capability
- add gvar writing capability
- add HVAR writing support
- add higher level methods for working more comfortably with variation data (like implemented color glyph layers and palettes)

## Motivation and Context

Finally, after supporting parsing of the required tables, we can also render glyph variations.

## How Has This Been Tested?

Visual testing, comparing coordinates with other implementations, added tests.

## Screenshots (if appropriate):
image
image

The tests in the unicode test suite regarding the supported tables look fine, only a few are failing due to minor rounding differences. Several unrelated tests have been fixed along the way.
`passing: 212, failing: 539)` as opposed to `passing: 37, failing: 708` with 1.3.4.
image
image
image
image
image
image
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I did `npm run test` and all tests passed green (including code styling checks).
- [X] I have added tests to cover my changes.
- [X] My change requires a change to the documentation.
- [X] I have updated the **README** accordingly.
- [X] I have read the **CONTRIBUTING** document.

PR #701, opened 2024-04-20
Variable font (= CFF2) writing

Connum:
**This PR is still a draft & work in progress!**

## Description

This PR is based on #699, which implements reading of different *var tables and rendering of variable fonts. This has to be merged first before merging this PR. It will implement the writing side, and as we always write in CFF format, it made it necessary to implement writing of CFF2 format as well, which is needed for variation data.

- [X] Store `blend` operation deltas
- [ ] Use deltas instead of re-parsing the CharString on transformation
- [ ] write CFF2 format
- [ ] support writing (global) sub routines
- [ ] support sub routines for glyphs that only consist of them
- [ ] refactor cff2 table structure (right now we store indices e.g. in vstore and the actual data in _vstore, which we do for none of the other tables; also sometimes dicts are objects with a single property instead of directly containing the data, which just adds overhead)
- [ ] make CFF2 the default, with a setting to force version 1 (for non-variable fonts or saving a non-variable version of a font)
- [ ] implement a method to create a non-variable version from any variation instance or coords

**TODOs for the future**
* Implement an algorithm that finds candidates for (global) sub routines and adds them accordingly

## Motivation and Context

Once we can render variable fonts via #699, with this PR we'll be able to create and save them as well. We can also make use of the reduced data size of CFF2 vs CFF.

## How Has This Been Tested?

Made sure that existing tests still pass (or adapted tests to account for the newly to-be-expected data), added new tests.

## Types of changes

## Checklist:


hybridherbst:
I've been spending some time (and thought, and tokens) on variable font writing support and got pretty far – it works for creating axes, masters, instances that are functioning correctly across browsers and softwares. It's currently here: https://github.com/hybridherbst/opentype.js/tree/font-variation-writing

Issue #62, opened 2014-08-14
Loading the font with @fontface?


Aetherpoint:
Curious – is there anyway to access the font object with a CSS @fontface? Or would that require recompiling the font somehow?

interactionist:
@fpirsch the 33% more additional base64 bytes should be mitigated by gzipping when you send the data ==> http://davidbcalhoun.com/2011/when-to-base64-encode-images-and-when-not-to/

Pomax:
Note that if opentype.js is already running in-browser, then bear in mind that the data:uri will basically be immediately unpacked again in-memory, so you only save a bit of space in the `data:` string. The opentype.js font representation, and the font as loaded from `data:` source will both stay exactly the same sizes. For transmission, using gzipped tables inside the WOFF is a good idea. If it's all in the same client session, there's virtually no benefit.

interactionist:
Ah yes, if it is all in the client that would make no sense. 
I thought they want to persist it in a CSS/HTML and transfer the font embedded over network.

cyberwombat:
Has any of this been implemented?

fdb:
OpenType.js doesn't have this yet, but Plumin.js (which is built on top of OpenType.js) uses [this technique](https://github.com/byte-foundry/plumin.js/blob/d0473591230fe7854e589aae14a199d8af86ea07/src/Font.js#L286).

cyberwombat:
I see thank you. Is it possible to extract a base64 rendition of the font from OT once it's loaded? I may be able to use it that way instead

cyberwombat:
Thanks. I was able to use the `toArrayBuffer()` and convert to base64 and load it that way in my css.

fdb:
Yeah, we should probably add that to OpenType.js directly....

Pomax:
As a small aside, given that this issue is grounded in the past (filed august 2014), when the webfont landscape was very different: these days you *do* want just WOFF (possibly with WOFF2), as `eot` is obsolete (MS no longer supports any browser that *needs* the `eot` format), `svg` has been disavowed (most browsers that still had support for it in 2014 have since removed support again), and `ttf`/`otf` are full system fonts, subject to potentially stricter validation than "explicitly for web" fonts.

Connum:
I agree there should be an easy integrated way to use the (possibly modified) fonts for text preview in browser context. Doesn't have to be in 2.0.0, but it shouldn't be too complicated to have this soon in 2.x.x at least.

[...]

PR #703, opened 2024-04-21
Sane buildflow

yne:
## Description

Based on existing work done by @ILOVEPIE in https://github.com/opentypejs/opentype.js/pull/693 this PR migrate to a sane build pipeline (which is urgent) while leaving any compatibility-related change to https://github.com/opentypejs/opentype.js/pull/693 (which is nice, but less urgent).

- Rename `src/*.js` to `src/*.mjs` and `test/*.js` to `test/*.spec.mjs`
- Remove unused reify dependency
- Use latest Mocha => no more critical vulnerabilities when `npm i`
- Use latest ESLint 9 => need to migrate to they new config format
- Deprecate load*/download functions (https://github.com/opentypejs/opentype.js/issues/675) and remove associated tests
- Remove dead code (private load*() function, isNode/isBrowser...)
- Fix some issues in recent tests addition
- Add support for many popular test runner:
```sh
npx mocha
npx bun test
npx jasmine "**/*.spec.mjs" # may fail on strict compare
```
*BONUS:* This PR also cleanup our ESLint rules:
- "ban forEach" is migrate to a "selector" rule
- prevent import loop => did nothing
- force import file extension => esbuild enforce it already

> [!TIP]
> This halves (243 => 144) our NPM dependencies

## Motivation and Context

Lower our attack surface by half, while keeping the backward-compatibility improvment in https://github.com/opentypejs/opentype.js/pull/693

## How Has This Been Tested?

```
npm i && npm run test
```

## Types of changes

## Checklist:


test/bidi.spec.mjs
```diff
@@ -98,12 +98,12 @@ describe('bidi.js', function() {
...
 
-    describe('thai scripts', () => {
+    describe('thai scripts', function() {
 
         let thaiFont;
         let bidiThai;
 
-        before(()=> {
+        beforeEach(()=> {
             thaiFont = loadSync('./test/fonts/NotoSansThai-Medium-Testing-v1.ttf');
```
Connum, line 107: Why do we need to load/parse the font before each test? If it's not being modified and therefore interfering with following tests, this just adds overhead and increases testing runtime.

test/bidi.spec.mjs
```diff
@@ -133,24 +133,24 @@ describe('bidi.js', function() {
...
                 assert.deepEqual(words, ['เป็นคนใจดีสำหรับทุกคน', 'ความรักคือทุกสิ่ง']);
             });
         });
     });
 
-    describe('noto emoji with ccmp', () => {
+    describe('noto emoji with ccmp', function () {
         let notoEmojiFont;
-        before(()=> {
+        beforeEach(()=> {
```
Connum, line 149: same as above: Why do we need to load/parse the font before each test? If it's not being modified and therefore interfering with following tests, this just adds overhead and increases testing runtime.

test/bidi.spec.mjs
```diff
@@ -133,24 +133,24 @@ describe('bidi.js', function() {
             });
         });
 
-        describe('thai contexts', () => {
+        describe('thai contexts', function () {
             it('should match thai words in a given text', () => {
                 const tokenizer = bidiThai.tokenizer;
-                tokenizer.tokenize('The king said: เป็นคนใจดีสำหรับทุกคน because ความรักคือทุกสิ่ง');
+                bidiThai.processText('The king said: เป็นคนใจดีสำหรับทุกคน because ความรักคือทุกสิ่ง');
```
Connum, line 139: I'm not sure what that does, can you explain please? Do we need to mix it into this PR?

test/glyph.spec.mjs
```diff
@@ -1,18 +1,18 @@
...
 
-describe('glyph.js', function() {
+describe('glyph.mjs', function() {
 
     describe('lazy loading', function() {
         let font;
         let glyph;
 
-        before(function() {
+        beforeEach(function() {
```
Connum, line 15: see above

test/glyph.spec.mjs
```diff
@@ -48,7 +48,7 @@ describe('glyph.js', function() {
         let trueTypeFont;
         let openTypeFont;
 
-        before(function() {
+        beforeEach(function() {
```
Connum, line 51: see above

test/glyph.spec.mjs
```diff
@@ -211,14 +211,14 @@ describe('glyph.js', function() {
...
-describe('glyph.js on low memory mode', function() {
+describe('glyph.mjs on low memory mode', function() {
     let opt = {lowMemory: true};
 
     describe('lazy loading', function() {
         let font;
         let glyph;
 
-        before(function() {
+        beforeEach(function() {
```
Connum, line 221: see above

test/glyph.spec.mjs
```diff
@@ -254,7 +254,7 @@ describe('glyph.js on low memory mode', function() {
         let trueTypeFont;
         let openTypeFont;
 
-        before(function() {
+        beforeEach(function() {
```
Connum, line 257: see above

src/util.mjs
```diff
@@ -31,16 +10,6 @@ function arraysEqual(ar1, ar2) {
     return true;
 }
 
-// Check if 2 objects are equal
-function objectsEqual(obj1, obj2) {
```
Connum, line 35: I was so sure that we were using this function, but I couldn't actually find it in the code base... strange! This could come in handy for caching transformed glyphs depending on the set variation coordinates, though, and as it's not adding much in file size (and should actually be tree-shaken if I'm not mistaken) and we already have the working test for it - do we really need to remove it?

test/util.js
```diff
@@ -1,18 +0,0 @@
-import assert from 'assert';
-import { objectsEqual } from '../src/util.js';
```
Connum, line 2: see above

Connum (changes requested):
See my comments - other than that, this looks good! Nice work getting rid of all those depedencies and moving towrds a more modern code base.

src/util.mjs
yne: I just took the opportunity while doing the cleanup of ther other utils.js dead code  (isBrowser/isNode)

test/util.js
yne: we could ad it back, but it feels weird to add dead code :wink:

test/bidi.spec.mjs
yne: `tokenizer.tokenize` did not registerContextChecker("thaiWord") so the test failed
`processText` however call tokenizeText() which register it here is the stacktrace

```
  1) bidi.mjs
       thai scripts
         thai contexts
           should match thai words in a given text:
     TypeError: ranges.map is not a function
      at Context.<anonymous> (file:///home/yne/opentype.js/test/bidi.spec.mjs:141:38)
      at process.processImmediate (node:internal/timers:478:21)

```

src/font.mjs
```diff
@@ -628,39 +627,10 @@ Font.prototype.toArrayBuffer = function() {
...
-    } else {
-        const fs = require('fs');
-        const buffer = Buffer.alloc(arrayBuffer.byteLength);
-        const view = new Uint8Array(arrayBuffer);
-        for (let i = 0; i < buffer.length; ++i) {
-            buffer[i] = view[i];
-        }
-        fs.writeFileSync(fileName, buffer);
-    }
+Font.prototype.download = function() {
```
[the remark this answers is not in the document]
yne: It's exact same reason:
all those platform-specific snippet have nothing to do with opentype.js itself and belongs to the documentation (if users can't ask "how to write a node ArrayBuffer to a file" to chatGPT)

I can put the replacement line in the deprecation notice if you want.

```js
fs.writeFileSync("out.otf", Buffer.from(font.toArrayBuffer()));
```
Or the 2 lines browser version:

```js
const href = window.URL.createObjectURL(new Blob([font.toArrayBuffer()]), {type: "font/opentype"});
Object.assign(document.createElement('a'), {download: "out.otf", href}).click();
```

src/tables/glyf.mjs
```diff
@@ -35,18 +35,18 @@ function parseGlyphCoordinate(p, flag, previousValue, shortVectorBitMask, sameBi
 // Parse a TrueType glyph.
 function parseGlyph(glyph, data, start) {
```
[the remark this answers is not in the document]
yne: Yes, 2 tests broke as they values are "undefined", here is the stacktrace

```js
  1) glyph.mjs
       lazy loading
         lazily loads numberOfContours:
     AssertionError [ERR_ASSERTION]: undefined == 2
      at Context.<anonymous> (file:///home/yne/opentype.js/test/glyph.spec.mjs:37:20)
      at process.processImmediate (node:internal/timers:478:21)

  2) glyph.mjs on low memory mode
       lazy loading
         lazily loads numberOfContours:
     AssertionError [ERR_ASSERTION]: undefined == 2
      at Context.<anonymous> (file:///home/yne/opentype.js/test/glyph.spec.mjs:243:20)
      at process.processImmediate (node:internal/timers:478:21)
```

Yet the values works for xMin/xMax/yMin/yMax so I simply migrated to the same format as them.

src/tables/cpal.mjs
```diff
@@ -220,7 +220,7 @@ function parseColor(color, targetFormat = 'hexa') {
         if(targetFormat == 'bgra') {
             return color;
         }
-    } else if(globalThis.window && globalThis.window.HTMLCanvasElement && /^[a-z]+$/i.test(color)) {
+    } else if(typeof document !== 'undefined' && /^[a-z]+$/i.test(color)) {
```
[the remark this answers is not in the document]
yne: `HTMLCanvasElement` arrived in firefox 1.5 (2005), yet our codebase use much more recent browser API (`Proxy`,`WeakMap`, ... which appeard in Firefox 18 in 2013)

So I think we are pretty much safe for <canvas> usage :)

See [Firefox4 backport](https://github.com/opentypejs/opentype.js/pull/693#issuecomment-2068088620)

src/glyphset.mjs
```diff
@@ -140,6 +140,7 @@ function ttfGlyphLoader(font, index, parseGlyph, data, position, buildPath) {
             return path;
         };
 
+        defineDependentProperty(glyph, 'numberOfContours', '_numberOfContours');
```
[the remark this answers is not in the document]
yne: see below. No Idea if this related to the loss of reify or the bump of Mocha, but it it's related to the PR

test/bidi.spec.mjs
yne: it allows to use the same codebase on the 3 most popular test runner
yet it has no significant impact on the overall test execution time 

https://mochajs.org/#hooks use `.before()`/`.after()` :cry: 
https://bun.sh/docs/test/lifecycle use `.beforeAll()`/`.afterAll()`
https://jasmine.github.io/api/edge/global use `.beforeAll()`/`.afterAll()`

I could also switch from mocha to `bun test` (and save even more NPM dependencies)
```
npx mocha
npx bun test
npx jasmine "**/*.spec.mjs"
```

Or just create a wrapper in `test/testutil.mjs` with `before = before||beforeAll`

test/bidi.spec.mjs
yne: There was a bug here
https://github.com/opentypejs/opentype.js/blob/master/test/path.js#L158

where `global.document` were created at init, but deleted after each test (which broke some of them)

yne:
> Nice work getting rid of all those depedencies and moving towrds a more modern code base.

Props to @ILOVEPIE who did the initial migration. I just reused it PR but stopped before doing legacy-compatibility stuff so he could rebase it PR on it

ILOVEPIE:
So, the point of this is to split my pr into two?

eslint-local-rules.js
ILOVEPIE, line 1: Why was this deleted?

package.json
```diff
@@ -22,76 +22,29 @@
   },
   "main": "./dist/opentype.js",
   "browser": "./dist/opentype.js",
-  "module": "./dist/opentype.module.js",
+  "module": "./dist/opentype.mjs",
   "scripts": {
     "build": "npm run b:umd && npm run b:esm",
     "dist": " npm run d:umd && npm run d:esm",
-    "test": "npm run build && npm run dist && mocha --require reify --recursive && npm run lint",
```
ILOVEPIE, line 29: I think we still need the ``--recursive``

src/opentype.mjs
```diff
@@ -3,135 +3,46 @@
...
+import meta from './tables/meta.mjs';
+import gasp from './tables/gasp.mjs';
+import svg from './tables/svg.mjs';
+import { PaletteManager } from './palettes.mjs';
 /**
  * The opentype library.
  * @namespace opentype
  */
 
-// File loaders /////////////////////////////////////////////////////////
```
ILOVEPIE, line 47: I thought we agreed that this would still function in 2.0.0 with a warning, and be removed in a subsequent release.

src/opentype.mjs
```diff
@@ -533,51 +444,20 @@ function parseBuffer(buffer, opt={}) {
  * We use the node.js callback convention so that
  * opentype.js can integrate with frameworks like async.js.
  * @alias opentype.load
- * @param  {string} url - The URL of the font to load.
- * @param  {Function} callback - The callback.
+ * @deprecated
  */
-function load(url, callback, opt = {}) {
```
ILOVEPIE, line 539: I thought we agreed that this would still function in 2.0.0 with a warning, and be removed in a subsequent release.

src/opentype.mjs
```diff
@@ -533,51 +444,20 @@ function parseBuffer(buffer, opt={}) {
...
 /**
  * Synchronously load the font from a URL or file.
  * When done, returns the font object or throws an error.
  * @alias opentype.loadSync
- * @param  {string} url - The URL of the font to load.
- * @param  {Object} opt - opt.lowMemory
- * @return {opentype.Font}
+ * @deprecated
  */
-function loadSync(url, opt) {
```
ILOVEPIE, line 579: I thought we agreed that this would still function in 2.0.0 with a warning, and be removed in a subsequent release.

test/opentype.spec.mjs
```diff
@@ -15,33 +15,6 @@ describe('opentype.js', function() {
         assert.equal(aGlyph.path.commands.length, 15);
     });
 
-    it('[deprecated] .load() promise resolve uppon success', function(done) {
```
ILOVEPIE, line 18: I thought we agreed that this would still function in 2.0.0 with a warning, and be removed in a subsequent release.

test/opentype.spec.mjs
```diff
@@ -84,23 +57,6 @@ describe('opentype.js', function() {
         assert.equal(aGlyph.path.commands.length, 14);
     });
 
-    it('[deprecated] .load() handles a parseBuffer error', function(done) {
```
ILOVEPIE, line 87: I thought we agreed that this would still function in 2.0.0 with a warning, and be removed in a subsequent release.

test/opentype.spec.mjs
```diff
@@ -119,7 +75,7 @@ describe('opentype.js', function() {
     });
 });
 
-describe('opentype.js on low memory mode', function() {
+describe('opentype.mjs on low memory mode', function() {
```
ILOVEPIE, line 78: Error here, this should still read opentype.js

ILOVEPIE (changes requested):
I can't approve of this in it's current state.

Connum:
As my brain needed a break from trying to implement CFF2 writing for variable glyphs, I'm currently working on a slim (=really just a couple of lines) plugin API to offer support for standalone CFF1 and PS1/Adobe Type 1 files without bloating the main library.

I think I would be on board with removing the load and download functions directly in 2.0.0 without prior depreciation notice if we were to offer a "browserhelper" and "nodehelper" plugin at the same time, references in a console message when trying to call the functions. The browserhelper could also provide the functionality for adding a preview of the font via CSS @font-face as discussed in #62.

That way, all that users had to do would be to add or import another small file delivered together with our lib in order to keep the functionality.

What do you guys think?

eslint-local-rules.js
yne: the usefull rule were put in the package.json eslint rule

src/opentype.mjs
yne: I thought the function would be *replaced* with a deprecation message

test/opentype.spec.mjs
yne: the 2.0.0 is a major release
major release does break things
so if we are brewing the 2.0.0 it make more sense to put the warning now to ease this major transition

yne:
> What do you guys think?

for the load/download function, a simple line in the "usage" chapter of the readme would suffice as it is just demonstration of how user can integrate opentype.js in they environment.

but I agree for a separation of concern

This is what jasmine is doing (jasmine/jasmine-core)

package.json
yne: you are right (226 vs 317)

src/font.mjs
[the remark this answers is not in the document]
yne: I've added a corresponding chapter in the README to explain how to get the Font back as file

yne:
> So, the point of this is to split my pr into two?

- Yes this PR remove reify so we can bump Mocha
- Your PR integrate SWC to add a `.legacy.js` build

test/util.js
Connum: ok, let's get rid of it for now...

ILOVEPIE:
Hey, @yne can we get this rebased on the current head so that it can be reviewed and merged if it's ok?

yne:
> Hey, @yne can we get this rebased on the current head so that it can be reviewed and merged if it's ok?

sure, I'll ping you once I get back home (~ end of week)

yne:
@ILOVEPIE It is in fact, already rebased on master.

So there was not much to do

ILOVEPIE (approved):
Looks good to me.

yne:
We can merge if @Connum is satisfied with the current justifications

PR #704, opened 2024-04-24
Implement plugins to support (subset) fonts in standalone CFF1 and Type1 format

### LINKED ISSUE #337: Corrupted font file handling

### LINKED ISSUE #370: TTF not contains name and cmap table parse error

### LINKED ISSUE #426: Support of subset-fonts

### LINKED ISSUE #643: Error parsing some cut font files using 'opentype. parse(...)'

Connum:
## Description

This PR implements a simple yet flexible plugin API and demonstrates its use with two plugins for parsing additional font formats, as well as some fixes.

The plugin API has not been documented in the README yet. It may still change drastically before we finally release 2.0.0, as we should figure out in which other places plugin hooks make sense and what data to supply to each entry point.

* `build` and `dist` separate plugin files
* gracefully parse partial font data from extracted subset fonts without throwing too early
* parse CFF1 standalone font files
* parse PostScript/PS1/T1/Adobe Type 1 format (.ps, .psa, .psb)
* fix some issues in the Glyph Inspector regarding contours of non-variable fonts
* keep comment blocks containing license information (required for meeting legal requirements and also stating our license in the minified dist files)
* fix nGlyphs calculation: count was treated differently for CFF1 and CFF2, but is actually the same (one more offset value than glyphs)

## Motivation and Context

* fix #643
* fix #370
* fix #337
* fix #426 

My PR #648 tried to achieve the graceful parsing by implementing a custom logger functionality, but that idea didn't seem to get any traction. This PR now achieves the same by logging warnings and errors to the console instead of throwing excepions in several places.

## How Has This Been Tested?

Added new test fonts and tests for the plugins.

image

## Types of changes

## Checklist:


Connum:
Pinging @yne @ILOVEPIE

src/encoding.js
```diff
@@ -317,7 +317,7 @@ function addGlyphNamesAll(font) {
         const c = charCodes[i];
         const glyphIndex = glyphIndexMap[c];
         glyph = font.glyphs.get(glyphIndex);
-        glyph.addUnicode(parseInt(c));
+        glyph && glyph.addUnicode(parseInt(c));
```
yne, line 320: you already used the optional chaining `.?` syntax early on,
so keep it unified (your choice :) )

src/plugins.mjs
```diff
@@ -0,0 +1,33 @@
+export const plugins = [];
```
yne, line 1: would it be cleaner to abstract this plugins array with a registerPlugins
so we could do all the registration check (typeof == function etc...) at registration time

(just a suggestion)

src/plugins/opentypejs.plugin.cff1file.js
```diff
@@ -0,0 +1,75 @@
...
+ * opentype.js
+ * https://github.com/opentypejs/opentype.js
+ * (c) 2015-present Frederik De Bleser and contributors
+ * opentype.js may be freely distributed under the MIT license.
+ */
+
+let isResponsible = new WeakMap();
+
+const plugin_cff1file = {
+    parseBuffer_signature: function(returnData, params) {
```
yne, line 15: I think there are two ways of declaring a method:
```js
parseBuffer_signature: function(returnData, params) {
```

```js
parseBuffer_signature(returnData, params) {
```

one of them bind the `this` at creation , the other at runtime
(maybe this would help avoid using those .call() usage ?)

src/tiny-inflate@1.0.3.esm.js
```diff
@@ -1,4 +1,6 @@
-/*
+/*!
```
yne, line 1: does `/*!` semantically mean something ? (just asking for ma curiosity)

yne (approved):
Love this new plugin approach !

We can now start considering offering a woff2 plugin !

src/tiny-inflate@1.0.3.esm.js
Connum: Yes, comment blocks starting with this will be considered license blocks and will not be removed from the minified output, so we are in accordance with the license terms of some third party code used (and also include our own license info in the minified dist files).

src/plugins/opentypejs.plugin.cff1file.js
Connum: Good point, I'll take a look at it

src/plugins.mjs
Connum: I thought about that as well, but then we'd also need methods to unregister a plugin and change the execution order in case two plugins modify the same entry point data. I think everyone should be familiar with modifying an array, so we should be fine leaving this without abstraction.

src/encoding.js
Connum: Oh, did I use optional chaining? I thought that would throw an error with the current reify setup... I'll take a look at it!

src/tiny-inflate@1.0.3.esm.js
yne: good to know !

src/encoding.js
yne: Yup here:

https://github.com/opentypejs/opentype.js/pull/704/files#diff-cd810153d07820a50b3950a985c42117e1dd7ea75207520e60b7306ce17b8901R396

ILOVEPIE:
@Connum this needs updating.

Issue #738, opened 2024-07-08
Gasp write support broken, opentype.js cannot read the fonts it creates


Balearica:
Support for the `gasp` table was added in #595, which ostensibly included both read and write support.  I cannot speak to the quality of the read function, as I am not familiar with the `gasp` table, however the write support is definitely broken.  It appears that any font file generated by opentype.js with a `gasp` table will be broken, and cannot be re-imported.

## Code At Issue
The write function attempts to iterate over a constant (`gasp.numRanges`) as if it were an array.  As a result, any code inside the loop will never run, and the `gasp` table will be broken.
https://github.com/opentypejs/opentype.js/blob/811a9c23b6806d6ffc169d04f092dddd68658b89/src/tables/gasp.mjs#L37-L40

While a broken `gasp` table is not a big deal per se, opentype.js currently throws an uncaught error if the `gasp` table fails to parse.  Therefore, opentype.js cannot read the fonts it writes that include a `gasp` table.

## Reproducible Example
This can be reproduced with any font with a `gasp` table, including many of the fonts in the `test` directory.  For example, the [Changa-Regular.ttf](https://github.com/opentypejs/opentype.js/blob/master/test/fonts/Changa-Regular.ttf) file works.  Simply load the font, convert it to an array buffer with `.toArrayBuffer()`, and then attempt to parse that ArrayBuffer using `opentype.parse`. 

## Suggested Fixes
I think there are several different aspects to this issue.  I will try and write a small patch that prevents the crashing tonight.  However, I have no knowledge of or interest in the `gasp` table, so don't plan on working on `gasp` support past that point. 

### 1. Patch the bug in the `gasp` write function
It looks like the author was trying to iterate over `gaspRanges` (an array) rather than `numRanges` (an integer).  I will confirm this change stops the error from being thrown and submit as a PR.

### 2. Add automated tests for the `gasp` write functions
While the PR that included the `gasp` write function did include some tests for reading `gasp` tables, it included no unit tests for writing `gasp` tables.  I think anybody adding a new feature of this nature should, at the very least, add a test to confirm that their read functions can parse the tables produced by their own write functions.

### 3. Prevent issues with optional tables from throwing an uncaught error
[...]

PR #739, opened 2024-07-08
Fixed gasp support bugs per #738

Balearica:
## Context
See #738.  In short, write support for the `gasp` table was implemented recently in #595, however is broken.  Any font written by opentype.js that includes a `gasp` table cannot be re-imported by opentype.js.  

## Changes
This PR makes two changes to resolve.  First, it fixes the typo in the `gasp` write function, which was trying to iterate over the integer `gasp.numRanges` rather than the array `gasp.gaspRanges`.  Second, it wraps the entire `gasp` read function in a `try` block.  This should prevent any future issues where failing to parse this optional table prevents a font from being read.  

## How Has This Been Tested?
I confirmed that the [Changa-Regular.ttf](https://github.com/opentypejs/opentype.js/blob/master/test/fonts/Changa-Regular.ttf) font could be saved/read using `toArrayBuffer`/`opentype.parse`, which (as described in #738) previously resulted in an error.  

## Types of changes

## Checklist:

### DIFF
--- a/src/opentype.mjs
+++ b/src/opentype.mjs
@@ -310,8 +310,13 @@ function parseBuffer(buffer, opt={}) {
                 metaTableEntry = tableEntry;
                 break;
             case 'gasp':
-                table = uncompressTable(data, tableEntry);
-                font.tables.gasp = gasp.parse(table.data, table.offset);
+                try {
+                    table = uncompressTable(data, tableEntry);
+                    font.tables.gasp = gasp.parse(table.data, table.offset);
+                } catch (e) {
+                    console.warn('Failed to parse gasp table, skipping.');
+                    console.warn(e);
+                }
                 break;
             case 'SVG ':
                 table = uncompressTable(data, tableEntry);
--- a/test/tables/gasp.spec.mjs
+++ b/test/tables/gasp.spec.mjs
@@ -31,7 +31,7 @@ describe('tables/gasp.mjs', function () {
 
     it('can write tables that are read as identical to the original', function() {
         const font2 = parse(font.toArrayBuffer());
-        assert.strictEqual(JSON.stringify(font.tables.gasp), JSON.stringify(font2.tables.gasp));
+        assert.deepStrictEqual(font.tables.gasp, font2.tables.gasp);
     });
 
 });
--- a/test/tables/gasp.spec.mjs
+++ b/test/tables/gasp.spec.mjs
@@ -29,4 +29,9 @@ describe('tables/gasp.mjs', function () {
         assert.equal(font.tables.gasp.gaspRanges[1].rangeGaspBehavior, 0x0001 + 0x0002 + 0x0004 + 0x0008); // all flags set = 15
     });
 
+    it('can write tables that are read as identical to the original', function() {
+        const font2 = parse(font.toArrayBuffer());
+        assert.strictEqual(JSON.stringify(font.tables.gasp), JSON.stringify(font2.tables.gasp));
+    });
+
 });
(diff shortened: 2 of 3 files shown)


Connum:
Please add a test that will prevent this issue from being reintroduced in the future.

Balearica:
I added a test that confirms fonts written with a `gasp` table can be read, and that the `gasp` table is identical to the original.

Connum:
> I added a test that confirms fonts written with a `gasp` table can be read, and that the `gasp` table is identical to the original.

Very nice, thanks! We shall have a review on this fix soon.

test/tables/gasp.spec.mjs
```diff
@@ -29,4 +29,9 @@ describe('tables/gasp.mjs', function () {
         assert.equal(font.tables.gasp.gaspRanges[1].rangeGaspBehavior, 0x0001 + 0x0002 + 0x0004 + 0x0008); // all flags set = 15
     });
 
+    it('can write tables that are read as identical to the original', function() {
+        const font2 = parse(font.toArrayBuffer());
+        assert.strictEqual(JSON.stringify(font.tables.gasp), JSON.stringify(font2.tables.gasp));
```
yne, line 34: we are using JSON because there are no assert functions that can do a deep compare ?

src/opentype.mjs
```diff
@@ -310,8 +310,13 @@ function parseBuffer(buffer, opt={}) {
                 metaTableEntry = tableEntry;
                 break;
             case 'gasp':
-                table = uncompressTable(data, tableEntry);
-                font.tables.gasp = gasp.parse(table.data, table.offset);
+                try {
```
yne, line 313: not sure if the try/catch is still needed here
and if it is needed, maybe we shall have it for every cases to keep a unified behavior

yne (approved):
thanks for the fix :)
just 2 simple questions/suggestion but no issue overall

src/opentype.mjs
Balearica: I believe that increasing the scope of OpenType.js to include a bunch of optional tables necessitates making them fail gracefully if they cannot be parsed.  The primary issue in my case is not that `gasp` could not be parsed correctly (I don't care about this table at all), but rather that my program crashed simply because this optional table could not be parsed.  Even after the fixes in this PR, exceptions during parsing this table may occur due to (1) undiscovered bugs in the parsing/writing functions or (2) flaws in the `gasp` table of otherwise valid user-provided fonts.  FontForge handles failures when parsing optional tables by skipping and printing a warning message, which I think makes sense. 

If you want, I can update such that all optional tables are handled in a similar way.  I only added for `gasp` as I see that table as higher-risk (as it was added the most recently without being fully tested), but I don't think crashing when an optional table cannot be parsed is ever a desirable behavior.

test/tables/gasp.spec.mjs
Balearica: I modified to use `assert.deepStrictEqual`.

src/opentype.mjs
yne: I would have expected the inner parse function to do the console.war/error + return null
But I just saw that `gasp.parse` had a `check.argument` that could `throw`
so my comment is irrelevant and the try/catch is well used here.
And I agree with you that we shall be error-resilient.

Connum: > I believe that increasing the scope of OpenType.js to include a bunch of optional tables necessitates making them fail gracefully if they cannot be parsed.

Graceful table handling will be implemented via #704

Connum:
Hi, (why) are we waiting for another review? It looks to me as if this is ready to be merged.

yne:
> Hi, (why) are we waiting for another review? It looks to me as if this is ready to be merged.

I've approved but the repo owner seems to want 2 approval, and @ILOVEPIE don't seems available to approve.

can't you bypass the process and force the merge ?

Issue #757, opened 2024-08-27
Fonts crafted from scratch always result in "not a valid font file" on Windows


mattlag:
I usually use v1.3.4, and assemble outlines into paths / glyphs / fonts and use opentype.js to export them to `otf` files. I started playing with the latest commit (which I will call the "July 30th" commit, because it is not a full release) to start seeing how SVG Table and Variable features are coming along. At some point I realized that no font assembled programmatically with the July 30th code results in a functional `otf` file.

## Expected Behavior

Fonts / Glyphs / Paths assembled programmatically should result in usable `otf` files.

## Current Behavior

I have a simple example in a [repo](https://github.com/mattlag/OTFJS-Export-Test/) - but it is basically the "Craft a font" example used in docs. If you try to do this with the July 30th code, it never (as far as I can tell) produces a functioning font file.

## Possible Solution

I tried using Fonttools TTX to look into both functional and non-functional `otf` files generated with v1.3.4 and the July 30th code. The main differences I saw in the new code is that it added an `ltag` table, and also added many `namerecord` entries in the `name` table with `platformID="0"`. I tried removing one and both of these new tables/entries and re-generating a font file with TTX, but the error seems to remain. So, I'm not sure what is causing Windows to not recognize the generated font files.

## Steps to Reproduce (for bugs)

The [repo src folder](https://github.com/mattlag/OTFJS-Export-Test/tree/main/src) is a basic bit of code that creates a good and bad otf file using two versions of opentype.js, you'll have to run it with something like `npx http-server` so the module imports work.

But a set of result files are already in the [repo result folder](https://github.com/mattlag/OTFJS-Export-Test/tree/main/result) in both otf and ttx format.

## Context

Overall, this is an investigation for Glyphr Studio - we leverage opentype.js to read and write font files. Opentype.js v1.3.4 is working great so far, but it would be awesome to integrate the new SVG Font and Variable Font features. Generating font objects from scratch then being able to export them to `otf` files is crucial.

## Your Environment

* Version used: Good export example, v1.3.4 / Bad export example, Latest commit from July 30th
* Font used: Generated from code
* Browser Name and version: Edge latest and Firefox latest
* Operating System and version (desktop or mobile): Windows 11
* Link to your project: https://github.com/mattlag/OTFJS-Export-Test/

Connum:
Thanks for bringing this to our attention! Could you please try to find the last working commit by gradually stepping back?

mattlag:
[...]

PR #758, opened 2024-08-29
Bug fix for #757 - write all name table entries for fonts crafted from scratch

mattlag:
## Description
Due to a change in 858a5d2, fonts that are created from scratch are not generating two default name table entries, which is causing fonts to be un-openable in Windows.

Lots of details captured in #757 - but here is a quick overview of the bug. The bug that was causing issues for two reasons:

1) the `for... in` loop is looping over the keys of the array, `0, 1, 2` as opposed to the values of the array (which looks like what was intended).
2) If an empty value is found, the `unicode` value is updated as opposed to whatever platform is currently being looped over.

## Motivation and Context
Any font object generated programmatically would successfully export, but the font file itself would not be able to be opened. In Windows the "This is not a valid font file" dialog would be shown.

## How Has This Been Tested?
Lots of tests explained in #757 

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/sfnt.mjs
+++ b/src/tables/sfnt.mjs
@@ -292,16 +292,16 @@ function fontToSfntTable(font) {
     const fontNamesWindows = font.names.windows || {};
 
     // do this as a loop to reduce redundant code
-    for (const platform in ['unicode', 'macintosh', 'windows']) {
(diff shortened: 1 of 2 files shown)


Connum (approved):
This looks good!

PR #761, opened 2024-09-10
os2: Fill Panose when writing font

Finii:
Parsing a font from a buffer and writing it back to a buffer changes all Panose values to zero.
This PR fixes this.

## Description

**[why]**
When an existing font file is read from a buffer and then written back into a buffer all Panose values will be zero.

The problem has been most likly been introduced with PR #630 (but I did not check that).

There are two disjunct data structures in the font.tables.os2 object that describe the panose values:
* An array with the Panose values `.panose = [ 1, 2, ...]`
* Dedicated properties for each Panose value e.g. `.bFamilyType`

The aforementioned PR seems to address only fonts created from scratch and not parsed from a buffer.

Writing out the font into a buffer will always use the dedicated Panose properties and ignore the .panose array.

Parsing a font does set the array but not the dedicated properties. They are not even existing then.

**[how]**
When an existing font is parsed the `.panose` array is filled (as before). But now the dedicated properties are also created and filled with the individual values.

**[note]**
The written font is always using the dedicated values. If a user changes the panose array that will have no effect. There are no checks if the data is consistent. Having the same data in two disjunct structures is not so nice to handle.

## How Has This Been Tested?

Manual tests with ttx ;)

## Types of changes

## Checklist:

### DIFF
--- a/src/tables/os2.mjs
+++ b/src/tables/os2.mjs
@@ -165,6 +165,16 @@ function parseOS2Table(data, start) {
     for (let i = 0; i < 10; i++) {
         os2.panose[i] = p.parseByte();
     }
+    os2.bFamilyType = os2.panose[0];
+    os2.bSerifStyle = os2.panose[1];
+    os2.bWeight = os2.panose[2];
+    os2.bProportion = os2.panose[3];
+    os2.bContrast = os2.panose[4];
+    os2.bStrokeVariation = os2.panose[5];
+    os2.bArmStyle = os2.panose[6];
+    os2.bLetterform = os2.panose[7];
+    os2.bMidline = os2.panose[8];
+    os2.bXHeight = os2.panose[9];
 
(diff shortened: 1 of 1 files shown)


Finii:
I believe how Panose is represented in other applications can be important, to see what users might expect.

### `showttf`

image

### `ttfdump`

image

### `fontforge`

image

Access via Python, via List (Array)

image


### Glyphs3

Screenshot 2024-09-10 at 22 12 22

Finii:
Added some test for os2 table parsing writing parsing (full round trip).

I struggled a bit with where to put the tests, as `test/tables/` in principle maybe should only test stuff inside that modules; but we need other modules to check this completely, so maybe test in `font`, but then the tests are to specific to one table.
Hope you get what I mean.

Anyhow, this is at least some starting point, feel free to move/remove the tests ;-D

Issue #765, opened 2024-09-25
Cannot read properties of undefined (reading 'getTransform')


ahaoboy:
## Expected Behavior
Show glyph info correctly

## Current Behavior
Throw error, and doesn't show glyph info

## Possible Solution
```
    if (glyph.numberOfContours > 0 && window.font.variation) {

```

## Steps to Reproduce (for bugs)
1. download font [JetBrainsMonoNerdFontMono-Regular.zip](https://github.com/user-attachments/files/17126038/JetBrainsMonoNerdFontMono-Regular.zip)
2. click any glyphs

PR #766, opened 2024-09-25
fix: Cannot read properties of undefined (reading 'getTransform')

ahaoboy:
## Description

close: https://github.com/opentypejs/opentype.js/issues/765

## Types of changes

## Checklist:

### DIFF
--- a/docs/glyph-inspector.html
+++ b/docs/glyph-inspector.html
@@ -161,7 +161,7 @@ function displayGlyphData(glyphIndex) {
         html += '<dt>leftSideBearing</dt><dd>'+glyph.leftSideBearing+'</dd>';
     }
     html += '</dl>';
-    if (glyph.numberOfContours > 0) {
+    if (glyph.numberOfContours > 0 && window.font.variation) {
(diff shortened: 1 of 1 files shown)


docs/glyph-inspector.html
```diff
@@ -161,7 +161,7 @@ <h1>Free Software</h1>
         html += '<dt>leftSideBearing</dt><dd>'+glyph.leftSideBearing+'</dd>';
     }
     html += '</dl>';
-    if (glyph.numberOfContours > 0) {
+    if (glyph.numberOfContours > 0 && window.font.variation) {
```
Connum, line 164: But couldn't there be glyphs with multiple contours but no variation axes? So I think the check for `window.font.variation` should happen inside, and depending on its availability, either the regular glyph or the transformed variant should be used.

Connum (changes requested):

Connum:
Good catch! Please see my review comment!

docs/glyph-inspector.html
ahaoboy: I’m not very familiar with this professional knowledge. I only encountered this issue during use, so there might be some oversights in the fix. Thank you for pointing it out.

PR #797, opened 2025-03-17
Rounding issues / bugfix

udatny:
i tried to use the 2.0 version / masterbranch for converting text into svg. 
i ran into issues with NaN's in the svg paths. 
this is my bugfix for it, including a test. sorry for the formatting issues. 
why is this repo / project so abandoned?

### DIFF
--- a/src/path.mjs
+++ b/src/path.mjs
@@ -32,8 +32,14 @@ function roundDecimal(float, places) {
         const roundedDecimalPart = decimalRoundingCache[places][decimalPart];
         return integerPart + roundedDecimalPart;
     }
-    
+
     const roundedDecimalPart = +(Math.round(decimalPart + 'e+' + places) + 'e-' + places);
+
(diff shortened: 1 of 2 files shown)


Connum:
> why is this repo / project so abandoned?

I have been the only one who still spends at least some time to look at new issues for quite some time, all other maintainers are too busy with their jobs and private lives. But maintaining a project of this size is too much for a single person to handle in their free time. I would love to add more features, but contributions by one maintainer require a review by at least another one, so some of my own PRs have taken very long or have still not been reviewed, which has taken away from my motivation to spend time on this project as well.

If there were more active maintainers, I would be willing to spend more time myself again.

Regarding your PR: I would prefer to find out why the result is `NaN` in the first place and handle the rounding correctly in those cases, instead of returning the unrounded value.

Connum:
@udatny this should have been fixed by #822 now, can you please check with the latest master?

PR #798, opened 2025-03-24
Add method to retrieve information about what characters in the original string were replaced by each glyph.

opengraphica:
Add a `stringToGlyphMapping` API to `font.mjs` to retrieve information about what characters in the original string were replaced by each glyph.

## Description
There exists `stringToGlyphIndexes` and `stringToGlyphs` APIs, which convert a string to an array of glyphs. Unfortunately, information about which character is transformed into each glyph is lost when using these methods.

This PR adds a `stringToGlyphMapping` method which returns a list of glyphs, as well as character indices in the original string that was provided to be able to track which characters were replaced by each glyph.

```stringToGlyphMapping(s: string): Array<{ glyph: Glyph, replaced: number[] }>;```

Example Input: `"fla"`
Example Output: `[ { glyph: Glyph, replaced: [0, 1] }, { glyph: Glyph, replaced: [2] } ]`

## Motivation and Context
A text editor needs to know which characters the user typed are converted to each glyph, otherwise it is impossible to implement text highlighting and cursor placement. 

For example, if a text document has ["f", "l"] and that is transformed into a single glyph for display ["fl"]. Pressing backspace should operate on the original array ["f", "l"] and modify the text document  into ["f"].  It can't operate on the "fl" glyph.

## How Has This Been Tested?
Beyond the unit tests, checked in editor code https://github.com/opengraphica/opengraphica. This adds a new API that does not touch existing APIs, so it should not negatively impact any other part of the codebase.

## Types of changes

## Checklist:

### CODE REVIEW

**Review by Connum — CHANGES_REQUESTED** [review]
As I'm not allowed to push to this PR, please update README.md after the `Font.stringToGlyphs(string)` section:
```
### `Font.stringToGlyphMapping()`
Convert the given text to an array of Glyphs and associated mapping of which characters in the original text were replaced by each Glyph index.
For example, if the font contains a ligature for "fl", `stringToGlyphMapping('fla')` would return ``[ { glyph: opentype.Glyph, replaced: [0, 1] }, { glyph: opentype.Glyph, replaced: [2] } ]`
```

Other than that, this looks good, and could come in handy for other use cases.

PR #818, opened 2026-03-03
Fix small bugs, typos, and inconsistencies

fdb:
## Summary
- **Fix actual bugs**: `.command` → `.type` in path optimization (dead code), `pathTramsform` typo (callback silently ignored), `getULong` → `parseULong` (would crash), `numberOfContours` → `_numberOfContours` (wrong property access), null crash in `Glyph.draw`
- **Fix throw hygiene**: throw `Error` objects instead of strings in Font constructor, add `new` keyword to `throw Error()` calls
- **Code quality**: `==` → `===`, `var` → `const`, remove unnecessary variable, fix two malformed JSDoc blocks

Each fix is in its own commit. 3 fixes include red/green TDD tests (321 total, up from 318).

## Test plan
- [x] All 321 tests pass
- [x] ESLint passes
- [x] Each commit individually tested before committing

### DIFF
--- a/src/font.mjs
+++ b/src/font.mjs
@@ -77,11 +77,11 @@ function Font(options) {
 
     if (!options.empty) {
         // Check that we've provided the minimum set of names.
-        if (!options.familyName) throw 'When creating a new Font object, familyName is required.';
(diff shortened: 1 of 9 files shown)


ILOVEPIE:

PR #819, opened 2026-03-03
Optimize cmap table: merge contiguous segments

fdb:
## Summary

- **Merge contiguous cmap segments** in `makeCmapTable()` — previously one segment was created per Unicode codepoint, now adjacent segments with contiguous codepoints and matching deltas (contiguous glyph indices) are merged into a single segment. For a font with A–Z, this reduces 26 segments to 1.
- **Remove leftover debug `console.log`** that printed "Adding CMAP format 12 (needed!)" to the console whenever a non-BMP glyph was encountered.
- **Remove outdated comment** that described the old one-segment-per-codepoint behavior.

The `mergeSegments()` function is a single-pass O(n) algorithm on sorted segments. It guards against merging into `0xFFFF` (reserved for the Format 4 terminator) while correctly allowing merging of non-BMP segments for Format 12.

## Test plan

- [x] Contiguous codepoints with same delta merge into one segment
- [x] Non-contiguous codepoints do NOT merge
- [x] Different deltas (non-contiguous glyph indices) do NOT merge
- [x] BMP boundary (0xFFFD + 0xFFFE) merges safely without hitting 0xFFFF
- [x] Multiple unicodes on a single glyph handled correctly
- [x] Non-BMP segments merge for Format 12
- [x] Round-trip test: create Font with A–Z → `toArrayBuffer()` → `parse()` → verify `charToGlyphIndex()` for all 26 letters
- [x] All 318 existing tests continue to pass

### DIFF
--- a/src/tables/cmap.mjs
+++ b/src/tables/cmap.mjs
@@ -250,6 +250,25 @@ function addTerminatorSegment(t) {
     });
 }
 
+function mergeSegments(segments) {
+    if (segments.length === 0) return segments;
+    const merged = [segments[0]];
+    for (let i = 1; i < segments.length; i++) {
+        const prev = merged[merged.length - 1];
+        const curr = segments[i];
+        if (
+            prev.end + 1 === curr.start &&
+            prev.delta === curr.delta &&
+            curr.end !== 0xFFFF
+        ) {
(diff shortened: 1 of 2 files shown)


ILOVEPIE:

PR #820, opened 2026-03-03
Add built-in TypeScript type definitions

### LINKED ISSUE #809: Built-in TypeScript types
Having TypeScript types would be super useful for TypeScript users. Currently the types as https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/opentype.js are outdated (missing at least font variations). For us types are _the_ starting point to figure out what a library can do. If they are included in the library, that saves a second dependency and ensures they are always synchronized with the latest version.

fdb:
## Summary

Closes #809. Ships `.d.ts` type definitions inside the npm package so TypeScript users no longer need the separate `@types/opentype.js` package.

- Enhances JSDoc annotations in core source files (`Font`, `Glyph`, `Path`, `BoundingBox`, `parse()`) and uses `tsc --allowJs --declaration --emitDeclarationOnly` to generate `.d.mts` files at build time
- Adds `PathCommand` discriminated union (`M`/`L`/`C`/`Q`/`Z`) for type-safe command handling, plus typed `FontOptions`, `GlyphRenderOptions`, and `GlyphOptions` interfaces
- Adds `"types"` field and `"exports"` map to `package.json` for proper TypeScript/bundler resolution
- Renames `tiny-inflate@1.0.3.esm.mjs` → `tiny-inflate-1.0.3.esm.mjs` (the `@` broke TypeScript module resolution)
- Adds type verification test (`test/types/test-types.ts`) that exercises the full public API under `--strict` mode

## Test plan

- [x] `npm run build` succeeds (UMD + ESM + types generation)
- [x] `npm run test:types` passes (strict-mode type verification test)
- [x] `npx mocha --recursive` — all 328 tests pass
- [x] `npm run lint` passes
- [x] No runtime behavior changes — only JSDoc annotations and build infrastructure


Connum:
@Jolg42 @yne @ILOVEPIE do any of you use TypeScript and could do the review for this?

ILOVEPIE:
I do use strictly typed languages so I'll review the types. We also need to update the closure compiler types too. Definitively typed was waiting on 2.0.0 to update their types, but it may be better for us to take over. @Connum

ILOVEPIE:
@fdb, by closure compiler types I mean [this](https://github.com/opentypejs/opentype.js/blob/master/externs/opentype.js), which got completely removed from the packaging output due to your recent packaging changes as far as I can tell, along with our non-ESM build and a whole bunch of other stuff that was also included.

EDIT: Nvmd I looked in the wrong folder regarding the ESM stuff, but the externs aren't there.

ILOVEPIE:
Sorry for making such a fuss, I'm just a little flustered trying to make sense of the changes that have happened without me knowing what was going on.

src/font.mjs
```diff
@@ -597,7 +617,7 @@ Font.prototype.validate = function() {
 /**
  * Convert the font object to a SFNT data structure.
  * This structure contains all the necessary tables and metadata to create a binary OTF file.
- * @return {opentype.Table}
+ * @return {any}
```
yne, line 620: why the loss of typing ?

src/glyph.mjs
```diff
@@ -253,8 +266,8 @@ Glyph.prototype.getLayers = function(font) {
 };
 
 /**
- * @param {opentype.Font} font
- * @returns {import('./svgimages.mjs').SVGImage | undefined}
+ * @param {import('./font.mjs').default} font
```
yne, line 269: might as well factorise those import() at the start once for all

tsconfig.json
yne, line 1: jsconfig.json would have been more meaningfull for me

src/tables/avar.mjs
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
yne, line 1: no rational given.

is there any roadmap that plan on finalizing this typing migration ?

I'm not a big fan of disabling checks unless thereis a reason (batched for later / unparseable file...)

src/tables/cmap.mjs
```diff
@@ -424,6 +426,7 @@ function makeCmapTable(glyphs) {
     return t;
 }
 
+/** @type {any} */
```
yne, line 429: why remove the implicit typing ?

yne (changes requested):

yne:
not sure if my comment will be answered since Claude was involved.

but overall I'm all in for better/stricter typing (auto generating the d.ts or having self-typed code via jsdoc) so we won't have to rely on 3rd party for it.

I'm still puzzled by some of typing regressions

ILOVEPIE:
@fdb @yne @Connum should we update the closure compiler types as a separate PR? Also we need to fix the closure compiler types to be distributed with the NPM package.

yne:
> @fdb @yne @Connum should we update the closure compiler types as a separate PR? Also we need to fix the closure compiler types to be distributed with the NPM package.

no issue for me if it's part of this PR but can't the closure types be autogenerated from the existing .d.ts types ? so we wont have to maintain both system at each change.

ILOVEPIE:
> can't the closure types be autogenerated from the existing .d.ts types ?

They can if I write some code to do that, but can't both be generated from JSDoc already? Closure types are literally just stubs of the functions and objects, but labeled with JSDoc types in a specific style. If we're going to go down that route we should probably decide what way we want to approach the generation from.

yne:
> can't both be generated from JSDoc already?

that would be ideal indeed. and `tsc` seems to be up to the task (at least to generate a .d.ts from a JSDoc project).

ILOVEPIE:
We would need to revamp the JSDoc comments we have to be closure compiler compatable and add more detail to them and write a script to generate the closure compiler types but it's totally possible for closure compiler too.

ILOVEPIE:
@yne Seeing as how @fdb hasn't updated this, I'm thinking of creating a new PR that does this properly and also adds Closure Compiler type generation.

fdb:
For me it's okay to nuke this PR and start over, I think the discussion is more interesting than the code itself. Is  the consensus that we go full in on JSDoc?

yne:
> full in on JSDoc?

yes

keeping the codebase "ready to use" while taking credits of the typing engine is the best of two worlds (until all JS engine adopt the type annotation)

PR #824, opened 2026-04-01
Fix OpenType Context Substitution Format 3 and decomposition type handling for Arabic/emoji shaping

fjobeir:
## Description
Replace the contextSubstitutionFormat3 implementation in src/features/featureQuery.mjs with one that uses lookupCoverageList and the standard lookup infrastructure (getLookupByIndex, getLookupMethod, getSubstitutionType) for proper GSUB lookup type 5, substFormat 3 handling. Additionally, fix both contextSubstitutionFormat3 and chainingSubstitutionFormat3 to use lookupRecord.sequenceIndex instead of iterating all input lookups, and add support for decomposition substitution type 21 in both functions.

Also adds a version field (1.4.0-beta.1) to package.json to support downstream dependency resolution.

## Motivation and Context

Some Arabic fonts (e.g. IBM Plex Sans Arabic) use OpenType Context Substitution Format 3 (GSUB lookup type 5, substFormat 3) for contextual letter forms. The previous implementation bypassed the standard lookup infrastructure and could fail to correctly resolve substitutions. Additionally, both contextSubstitutionFormat3 and chainingSubstitutionFormat3 iterated all input lookups for each lookup record instead of targeting the specific glyph at sequenceIndex, causing duplicate glyph output. Fonts like noto-emoji that use nested decomposition substitution (type 21) within context/chaining lookups also failed because only type 12 was handled.

This is part of a broader effort to fix RTL (Arabic/Hebrew) text rendering in Next.js OG image generation, which uses Satori -> opentype.js for font shaping.

## How Has This Been Tested?
- All 328 existing tests pass (npm run test), including:
  - featureQuery.mjs - "should parse multiple glyphs - ligature substitution format 3 (53)" — validates correct substitution output [54, 54]
  - bidi.mjs - "shape emoji with sub_5" — validates flag emoji shaping produces [1850] (not [1850, 1850])
  - featureQuery.mjs - "should find a substitute - chaining context substitution format 3 (63)" — validates chaining substitution still produces [1348]
- Downstream Satori test suite (432 tests) passes with this change, including the Arabic language rendering test

N/A — changes affect font shaping logic, validated via automated tests.

## Types of changes

## Checklist:


fjobeir:
Hi @fdb would you please take a look whenever possible

fdb:
Hi @fjobeir, I looked at this. Some notes:


1. please don't commit the version number in package.json or a package-lock.json; we handle this separately. Once this is merged, I can make a new version you can use upstream.

2. We're missing tests for this  PR. specifically it would be nice to have a font with OpenType Context Substitution Format 3 (GSUB lookup type 5, substFormat 3). I've tried validating with  IBM Plex Sans Arabic (all weights, all formats, downloaded from Google Fonts as well as from https://github.com/IBM/plex) but the font doesn't actually use the structures this PR fixes:
- it has no GSUB lookup type 5 (context substituion) lookups
- all its type 6 (chaining context) subtables have `inputCoverage.length === 1` with `sequenceIndex: 0` so the old (buggy) loop produces the same results as the fixes.
In other words, the PR changes are a no-op for that font. Can you point me to a specific font where this fix changes the output or crashes opentype.js? 

Ideally we have two things:
- A minimal, synthetic test font that uses our raw font generation helpers (see test/generate-circular-ref-font.mjs for an example). This would then create a font with GSUB type 5, format 3 lookup with coverages.length >= 2 and lookupRecords with different sequenceIndex values, so we can assert the fix works: 

```
// Pseudo-structure for the test font:
// - 5 glyphs: .notdef, A, B, A', B'
// - GSUB lookup type 5, format 3:
//     coverages: [covers A, covers B]        ← match A followed by B
//     lookupRecords:
//       { sequenceIndex: 0, lookupListIndex: 0 }  ← substitute A → A'
//       { sequenceIndex: 1, lookupListIndex: 1 }  ← substitute B → B'
// - Nested lookup 0: single sub A → A'
// - Nested lookup 1: single sub B → B'
//
// Input [A, B] should produce [A', B']
// The old bug would produce [A', B', A', B'] (iterating all inputs per record)
```

- In addition, a real-world font that exercises this path, so we can validate against HarfBuzz and compare. If you have the font that triggered the crash in Satori/next.js OG, that would be great. 

3. There are two ways you handle subtype 21: in chainingSubstitutionFormat3 you use a spread (`substitutions.push(...substitution)`) and in contextSubstitutionFormat3 you use a regular push (`substitutions.push(substitution)`). I think the non-spread version is the correct one, however now they accidentally yield the same result anyway.

```js
// In chainingSubstitutionFormat3, change:
if (substitution) substitutions.push(...substitution);
// to:
if (substitution) substitutions.push(substitution);
```

4. Extension subtable handling in `contextSubstitutionFormat3`

The existing chainingSubstitutionFormat3 handles type '71' (extension subtables). These are wrappers that exist purely to allow 32-bit offsets in tables that would otherwise be limited to 16-bit, they don't change the substitution logic, they just redirect to the real subtable data. The existing code handles this by unwrapping `subtable.extension` before dispatch (lines 157-161 in the current code). Your rewritten contextSubstitutionFormat3 doesn't include this unwrapping, so fonts that use extension wrappers inside a context substitution format 3 lookup will silently fail.

Sorry for the long review, but GSUB is a tricky beast 😁 If you need any further help, lmk

fjobeir:
Thanks @fdb I have updated the PR

fdb:
Thanks will look at it tomorrow. Can you also tell me which production font triggered the bug?

fjobeir:
@fdb The bug was discovered while fixing RTL (Arabic/Hebrew) text rendering in Next.js OG image generation (@vercel/og → Satori → opentype.js). During investigation of the shaping pipeline, the Context Substitution Format 3 issue surfaced with Noto Emoji (test/fonts/noto-emoji.ttf, already in the repo) — emoji flag sequences like 🇺🇺 produced duplicate glyphs [1850, 1850] instead of [1850]. The fix also benefits sub5.ttf (also already in the repo) which exercises the same lookup type.

fjobeir:
Hi @fdb a kind reminder to take a look here 😄

fdb (approved):

fdb:
Thanks for the reminder and the changes! Looks good now! 

Thanks again for your help, I'll make a new release so you can use that version directly.

fdb:
@fjobeir version 1.3.5 is now published: https://www.npmjs.com/package/opentype.js

PR #826, opened 2026-04-03
Prevent stack overflow from circular composite glyph references

fdb:
## Description
This change adds protection against stack overflow errors when loading TrueType fonts that contain circular references in composite glyphs. The fix tracks which glyphs are currently being resolved and skips any circular references encountered during path building.

### Changes Made:
1. **src/tables/glyf.mjs**: Added a module-scoped `_resolving` Set to track glyph indices currently being processed. When building composite glyph paths, the code now checks if a component glyph is already being resolved and skips it if a circular reference is detected. The tracking is properly cleaned up using try/finally to ensure the Set is maintained correctly across recursive calls.

2. **test/generate-circular-ref-font.mjs**: Added a utility script that generates a minimal test font (`circular-composite.ttf`) containing two composite glyphs that reference each other circularly. This font is used to verify the fix works correctly.

3. **test/glyph.spec.mjs**: Added a test case that loads the circular composite font and calls `getPath()` on both glyphs to verify they return valid Path objects without causing a stack overflow.

## Motivation and Context
Fonts with circular composite glyph references (where glyph A references glyph B and glyph B references glyph A) would previously cause infinite recursion and stack overflow errors when calling `getPath()`. While such fonts are malformed, the library should handle them gracefully rather than crashing.

## How Has This Been Tested?
- Added a new test case `circular composite glyph references` that loads a specially crafted font with circular references and verifies that `getPath()` returns valid Path objects without throwing errors
- The test font is generated by the new `generate-circular-ref-font.mjs` script
- All existing tests continue to pass with the circular reference detection in place

## Types of changes

## Checklist:
- [x] I have added tests to cover my changes
- [x] The fix handles the edge case gracefully without breaking existing functionality

https://claude.ai/code/session_014E56osqBcVfPwdgHHtrMBJ


ILOVEPIE:

PR #830, opened 2026-04-20
fix(sfnt): mask bytes to 8-bit in computeCheckSum for non-ASCII data

mtsubs:
### Context

I'm Meow P (From @MontageSubs), working on [ASS Subsetter](https://github.com/MontageSubs/ass-subset), a browser-based tool for subsetting fonts embedded in ASS subtitle files. It's built entirely on opentype.js, thank you for this excellent library that made our tool possible. I'm excited to contribute this fix back upstream.

While testing with large CJK CFF fonts (Source Han Sans, Source Han Serif), I found that fonts generated by `toArrayBuffer()` are rejected by Windows Font Viewer and the Windows DWRITE stack, even though they load fine in every media player and browser I tested. That discrepancy made me suspect a structural validation issue rather than a rendering one.

### What's wrong

`computeCheckSum` in `src/tables/sfnt.mjs` has **two separate root causes** that combine to produce incorrect per-table checksums:

#### Bug 1: Signed Integer Overflow in Bit Shifting

JavaScript's `<<` operator always returns a signed 32-bit integer. When the high byte of a 4-byte word is >= 0x80, shifting left by 24 bits produces a negative number:

```javascript
0xFC << 24  // -67108864, not 4227858432
```

The sum accumulates these negative contributions and drifts below zero. The `sum %= Math.pow(2, 32)` does not fix this because JavaScript's remainder operator preserves the sign of the dividend:

```javascript
-67108864 % 4294967296  // still -67108864
```

#### Bug 2: M.CHARARRAY Writes Values > 255 (Root Cause)

M.CHARARRAY encodes strings using `charCodeAt()`, which returns full Unicode codepoints for non-ASCII characters. For example:
- '思' returns `0x601D` (> 255)
- '源' returns `0x6E90` (> 255)

This causes a critical mismatch:

1.`computeCheckSum()` calculates using these raw values (> 255)
2. `Uint8Array` (used for file output) truncates each element to 8 bits when writing:
   - `0x601D` becomes `0x1D`
   - `0x6E90` becomes `0x90`

Since the checksum is computed on untruncated values but the file contains truncated bytes, **the checksums never match**.

Example: For a CFF table with Chinese family name, computeCheckSum produces `0xf1332993`, but the actual file checksum is `0xe290b993`.

#### Additional Issues

- The function mutates the input array directly via `bytes.push(0)`. If the caller's array is reused, subsequent reads include the padding bytes from the first checksum call.
- The checkSumAdjustment calculation can produce a negative result when checkSum > 0xB1B0AFBA, which should be coerced to unsigned 32-bit.

## Why It Shows Up on Large CJK Fonts Specifically

Small subsets tend to pass because the sign errors in individual words partially cancel each other out over short sequences. Once a CFF table is large enough (roughly 800+ glyphs of CJK data), there are enough high-byte words that the accumulated error becomes significant. Windows validates every table checksum strictly; most rendering engines skip the check entirely.

## Fix

Replace computeCheckSum with a version that:
1. Masks each byte to 8-bit (`& 0xFF`) to match Uint8Array truncation
2. Uses `>>> 0` throughout to stay in unsigned 32-bit arithmetic
3. Avoids mutating the input array

```javascript
function computeCheckSum(bytes) {
    let sum = 0;
    const padded = (bytes.length + 3) & ~3;
    for (let i = 0; i < padded; i += 4) {
        sum = (sum + (
            ((bytes[i]     || 0) & 0xFF) << 24 |
            ((bytes[i + 1] || 0) & 0xFF) << 16 |
            ((bytes[i + 2] || 0) & 0xFF) << 8  |
            ((bytes[i + 3] || 0) & 0xFF)
        ) >>> 0) >>> 0;
    }
    return sum;
}
```

And guard checkSumAdjustment against underflow:

```javascript
tableFields[i].value.checkSumAdjustment = (0xB1B0AFBA - checkSum) >>> 0;
```

Verified: patched build produces byte-for-byte matching checksums against fonttools on Source Han Sans subsets of 1200+ glyphs where the original was broken.

## How Has This Been Tested?

Apply the fix and test it in my tools:

**Stable deployment** (without this fix, using internal workaround):
https://subs.js.org/ass-subset/
https://github.com/MontageSubs/ass-subset/

**Test deployment** (with this fix applied):
https://mtsubs.github.io/ass-subset/
https://github.com/mtsubs/ass-subset/

Test fonts:
- https://github.com/adobe-fonts/source-han-serif/releases
- https://github.com/adobe-fonts/source-han-sans/releases

**Note:** Our tools tested against opentype.js 1.3.4. The GitHub version has API changes (e.g., fontFamily naming structure) that we have not yet adapted to.

## Types of changes

## Checklist:

- [X] I have read the **Contribute** README section.

### DIFF
--- a/src/tables/cmap.mjs
+++ b/src/tables/cmap.mjs
@@ -269,6 +269,12 @@ function mergeSegments(segments) {
     return merged;
 }
 
+// The Format 4 length field is a USHORT, capping the subtable at 65535 bytes.
+// Each segment contributes 8 bytes (one entry in each of the four parallel arrays),
+// plus the 14-byte header and the 2-byte reservedPad, leaving room for at most
+// floor((65535 - 16) / 8) = 8189 segments before overflow.
+const CMAP4_MAX_SEGMENTS = Math.floor((0xFFFF - 16) / 8);
+
 // Make cmap table, format 4 by default, 12 if needed only
 function makeCmapTable(glyphs) {
     // Plan 0 is the base Unicode Plan but emojis, for example are on another plan, and needs cmap 12 format (with 32bit)
@@ -282,6 +288,29 @@ function makeCmapTable(glyphs) {
             isPlan0Only = false;
             break;
         }
+    }
+
+    // Build and merge segments up front so we can measure the true segment count
+    // before committing to a table layout.
+    const allSegments = [];
+    for (i = 0; i < glyphs.length; i += 1) {
+        const glyph = glyphs.get(i);
+        for (let j = 0; j < glyph.unicodes.length; j += 1) {
+            addSegment({ segments: allSegments }, glyph.unicodes[j], i);
+        }
+    }
+    allSegments.sort(function(a, b) { return a.start - b.start; });
+    const mergedSegments = mergeSegments(allSegments);
+
+    // Count BMP segments (those that fit in Format 4).  If after merging the
+    // count still exceeds what a USHORT length field can encode, Format 4 cannot
+    // represent the full mapping without silent truncation.  In that case we fall
+    // back to emitting Format 12 for all glyphs, keeping a minimal Format 4
+    // subtable (terminator only) for parsers that require its presence.
+    const bmpSegmentCount = mergedSegments.filter(s => s.start <= 0xFFFF).length;
+    const cmap4Overflows = bmpSegmentCount > CMAP4_MAX_SEGMENTS;
+    if (cmap4Overflows) {
+        isPlan0Only = false;
     }
 
     let cmapTable = [
@@ -315,18 +344,7 @@ function makeCmapTable(glyphs) {
 
     const t = new table.Table('cmap', cmapTable);
 
-    t.segments = [];
-    for (i = 0; i < glyphs.length; i += 1) {
-        const glyph = glyphs.get(i);
-        for (let j = 0; j < glyph.unicodes.length; j += 1) {
-            addSegment(t, glyph.unicodes[j], i);
-        }
-    }
-    t.segments.sort(function (a, b) {
-        return a.start - b.start;
-    });
-
-    t.segments = mergeSegments(t.segments);
+    t.segments = mergedSegments;
 
     addTerminatorSegment(t);
 
@@ -347,8 +365,12 @@ function makeCmapTable(glyphs) {
     for (i = 0; i < segCount; i += 1) {
         const segment = t.segments[i];
 
-        // CMAP 4
-        if (segment.end <= 65535 && segment.start <= 65535) {
+        // CMAP 4: when the table would overflow, emit only the terminator segment
+        // so that Format 4 remains structurally valid.  All actual mappings are
+        // carried by the Format 12 subtable in that case.
+        const includeInCmap4 = segment.end <= 65535 && segment.start <= 65535 && !cmap4Overflows ||
+            (cmap4Overflows && segment.end === 0xFFFF && segment.start === 0xFFFF);
+        if (includeInCmap4) {
             endCounts.push({name: 'end_' + i, type: 'USHORT', value: segment.end});
             startCounts.push({name: 'start_' + i, type: 'USHORT', value: segment.start});
             idDeltas.push({name: 'idDelta_' + i, type: 'SHORT', value: segment.delta});
--- a/src/tables/cmap.mjs
+++ b/src/tables/cmap.mjs
@@ -290,6 +290,17 @@ function makeCmapTable(glyphs) {
         }
     }
 
+    // Create the CMAP table with some values that still need calculation missing.
+    let cmapTable = [
+        {name: 'version', type: 'USHORT', value: 0},
+        {name: 'numTables', type: 'USHORT', value: null},
+
+        // CMAP 4 header
+        {name: 'platformID', type: 'USHORT', value: 3},
+        {name: 'encodingID', type: 'USHORT', value: 1},
+        {name: 'offset', type: 'ULONG', value: null}
+    ];
+
     // Build and merge segments up front so we can measure the true segment count
     // before committing to a table layout.
     const allSegments = [];
@@ -307,19 +318,16 @@ function makeCmapTable(glyphs) {
     // represent the full mapping without silent truncation.  In that case we fall
     // back to emitting Format 12 for all glyphs, keeping a minimal Format 4
     // subtable (terminator only) for parsers that require its presence.
-    const bmpSegmentCount = mergedSegments.filter(s => s.start <= 0xFFFF).length;
+    let bmpSegmentCount = 0;
+    for (let i = 0; i < mergedSegments.length; i++)
+        if(mergedSegments[i].start <= 0xFFFF)
+            bmpSegmentCount++;
     const cmap4Overflows = bmpSegmentCount > CMAP4_MAX_SEGMENTS;
     isPlan0Only &&= !cmap4Overflows;
 
-    let cmapTable = [
-        {name: 'version', type: 'USHORT', value: 0},
-        {name: 'numTables', type: 'USHORT', value: isPlan0Only ? 1 : 2},
-
-        // CMAP 4 header
-        {name: 'platformID', type: 'USHORT', value: 3},
-        {name: 'encodingID', type: 'USHORT', value: 1},
-        {name: 'offset', type: 'ULONG', value: isPlan0Only ? 12 : (12 + 8)}
-    ];
+    // Fill in missing values in CMAP table
+    cmapTable[1].value = isPlan0Only ? 1 : 2;
+    cmapTable[4].value = isPlan0Only ? 12 : 20;
 
     if (!isPlan0Only)
         cmapTable.push(...[
--- a/src/tables/cmap.mjs
+++ b/src/tables/cmap.mjs
@@ -231,8 +231,8 @@ function parseCmapTable(data, start) {
     return cmap;
 }
 
-function addSegment(t, code, glyphIndex) {
-    t.segments.push({
+function addSegment(segments, code, glyphIndex) {
+    segments.push({
         end: code,
         start: code,
         delta: -(code - glyphIndex),
@@ -273,7 +273,7 @@ function mergeSegments(segments) {
 // Each segment contributes 8 bytes (one entry in each of the four parallel arrays),
 // plus the 14-byte header and the 2-byte reservedPad, leaving room for at most
 // floor((65535 - 16) / 8) = 8189 segments before overflow.
-const CMAP4_MAX_SEGMENTS = Math.floor((0xFFFF - 16) / 8);
+const CMAP4_MAX_SEGMENTS = 8189;
 
 // Make cmap table, format 4 by default, 12 if needed only
 function makeCmapTable(glyphs) {
@@ -296,7 +296,7 @@ function makeCmapTable(glyphs) {
     for (i = 0; i < glyphs.length; i += 1) {
         const glyph = glyphs.get(i);
         for (let j = 0; j < glyph.unicodes.length; j += 1) {
-            addSegment({ segments: allSegments }, glyph.unicodes[j], i);
+            addSegment(allSegments, glyph.unicodes[j], i);
         }
     }
     allSegments.sort(function(a, b) { return a.start - b.start; });
@@ -309,9 +309,7 @@ function makeCmapTable(glyphs) {
     // subtable (terminator only) for parsers that require its presence.
     const bmpSegmentCount = mergedSegments.filter(s => s.start <= 0xFFFF).length;
     const cmap4Overflows = bmpSegmentCount > CMAP4_MAX_SEGMENTS;
-    if (cmap4Overflows) {
-        isPlan0Only = false;
-    }
+    isPlan0Only &&= !cmap4Overflows;
 
     let cmapTable = [
         {name: 'version', type: 'USHORT', value: 0},
@@ -366,11 +364,10 @@ function makeCmapTable(glyphs) {
         const segment = t.segments[i];
 
         // CMAP 4: when the table would overflow, emit only the terminator segment
-        // so that Format 4 remains structurally valid.  All actual mappings are
+        // so that Format 4 remains structurally valid. All actual mappings are
         // carried by the Format 12 subtable in that case.
-        const includeInCmap4 = segment.end <= 65535 && segment.start <= 65535 && !cmap4Overflows ||
-            (cmap4Overflows && segment.end === 0xFFFF && segment.start === 0xFFFF);
-        if (includeInCmap4) {
+        if (segment.end <= 65535 && segment.start <= 65535 && !cmap4Overflows ||
+            (cmap4Overflows && segment.end === 0xFFFF && segment.start === 0xFFFF)) {
             endCounts.push({name: 'end_' + i, type: 'USHORT', value: segment.end});
             startCounts.push({name: 'start_' + i, type: 'USHORT', value: segment.start});
             idDeltas.push({name: 'idDelta_' + i, type: 'SHORT', value: segment.delta});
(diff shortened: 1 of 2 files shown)


mtsubs:
I think my use cases are really pushing opentypejs to its limits, haha, and I just found [another bug](https://github.com/MontageSubs/ass-subset/commit/55d20cf6d7da694abb1ef37680df7c571c2da15b). This doesn't usually happen so I'll double check it first and figure out how to fix it in this project. Then I'll submit PR. In the meantime @fdb could you review and test this PR for me? If it actually fixes the issue, I'm looking forward to seeing you merge it. But I really hope you can do some thorough testing though, because my test cases are pretty limited and can't really cover all the complex scenarios the project might run into, thank you.

ILOVEPIE:
> I think my use cases are really pushing opentypejs to its limits, haha, and I just found [another bug](https://github.com/MontageSubs/ass-subset/commit/55d20cf6d7da694abb1ef37680df7c571c2da15b). This doesn't usually happen so I'll double check it first and figure out how to fix it in this project. Then I'll submit PR. In the meantime @fdb could you review and test this PR for me? If it actually fixes the issue, I'm looking forward to seeing you merge it. But I really hope you can do some thorough testing though, because my test cases are pretty limited and can't really cover all the complex scenarios the project might run into, thank you.

Don't worry I became a maintainer of opentype.js to support my work on [SABRE.js](https://github.com/SABRE-JS/SABRE.js) a pure JS ASS subtitle renderer, so I know your pain. I'll take a look tomorrow.

ILOVEPIE:
Please note that 2.0.0 is already out.

ILOVEPIE:
2.0.0 has better CJK support so this may already be fixed, but I will check tomorrow.

ILOVEPIE:
Thank you for pointing out the windows DWRITE issue. I assume that causes issues with vsfilter and XY-Vsfilter. I'll go through your fix more thoroughly this coming evening.

src/tables/cmap.mjs
```diff
@@ -269,6 +269,12 @@ function mergeSegments(segments) {
     return merged;
 }
 
+// The Format 4 length field is a USHORT, capping the subtable at 65535 bytes.
+// Each segment contributes 8 bytes (one entry in each of the four parallel arrays),
+// plus the 14-byte header and the 2-byte reservedPad, leaving room for at most
+// floor((65535 - 16) / 8) = 8189 segments before overflow.
+const CMAP4_MAX_SEGMENTS = Math.floor((0xFFFF - 16) / 8);
```
ILOVEPIE: Why is this not a constant? Couldn't this just be:
```js
const CMAP4_MAX_SEGMENTS = 8189;
```
?

src/tables/cmap.mjs
```diff
@@ -284,6 +290,29 @@ function makeCmapTable(glyphs) {
         }
     }
 
+    // Build and merge segments up front so we can measure the true segment count
```
ILOVEPIE, line 304: More of a stylistic thing, but wouldn't it make more sense to create the segments after the table we put them in, i.e. moving this code below line 345?

src/tables/cmap.mjs
```diff
@@ -284,6 +290,29 @@ function makeCmapTable(glyphs) {
...
+    // before committing to a table layout.
+    const allSegments = [];
+    for (i = 0; i < glyphs.length; i += 1) {
+        const glyph = glyphs.get(i);
+        for (let j = 0; j < glyph.unicodes.length; j += 1) {
+            addSegment({ segments: allSegments }, glyph.unicodes[j], i);
+        }
+    }
+    allSegments.sort(function(a, b) { return a.start - b.start; });
```
ILOVEPIE, line 313: Do we need to sort the segments?

src/tables/cmap.mjs
```diff
@@ -284,6 +290,29 @@ function makeCmapTable(glyphs) {
         }
     }
 
+    // Build and merge segments up front so we can measure the true segment count
+    // before committing to a table layout.
+    const allSegments = [];
+    for (i = 0; i < glyphs.length; i += 1) {
+        const glyph = glyphs.get(i);
+        for (let j = 0; j < glyph.unicodes.length; j += 1) {
+            addSegment({ segments: allSegments }, glyph.unicodes[j], i);
```
ILOVEPIE: I'd rather rework the addSegment function that generate all these unnecessary objects in memory. we should just make the addSegment function take allSegments here instead of a table, or something.

src/tables/cmap.mjs
```diff
@@ -284,6 +290,29 @@ function makeCmapTable(glyphs) {
...
+
+    // Count BMP segments (those that fit in Format 4).  If after merging the
+    // count still exceeds what a USHORT length field can encode, Format 4 cannot
+    // represent the full mapping without silent truncation.  In that case we fall
+    // back to emitting Format 12 for all glyphs, keeping a minimal Format 4
+    // subtable (terminator only) for parsers that require its presence.
+    const bmpSegmentCount = mergedSegments.filter(s => s.start <= 0xFFFF).length;
+    const cmap4Overflows = bmpSegmentCount > CMAP4_MAX_SEGMENTS;
+    if (cmap4Overflows) {
+        isPlan0Only = false;
```
ILOVEPIE: this is pointless, just use cmap4 overflows then, or assign a new value to isPlan0Only, there's no need for this if.

ILOVEPIE: what i mean is:
```js 
isPlan0Only =  isPlan0Only && !cmap4Overflows;
```
can replace this whole if statement.

ILOVEPIE: or even more succinctly:
```js
isPlan0Only &= !cmap4Overflows;
```

src/tables/cmap.mjs
```diff
@@ -347,8 +365,12 @@ function makeCmapTable(glyphs) {
     for (i = 0; i < segCount; i += 1) {
         const segment = t.segments[i];
 
-        // CMAP 4
-        if (segment.end <= 65535 && segment.start <= 65535) {
+        // CMAP 4: when the table would overflow, emit only the terminator segment
+        // so that Format 4 remains structurally valid.  All actual mappings are
+        // carried by the Format 12 subtable in that case.
+        const includeInCmap4 = segment.end <= 65535 && segment.start <= 65535 && !cmap4Overflows ||
+            (cmap4Overflows && segment.end === 0xFFFF && segment.start === 0xFFFF);
```
ILOVEPIE: Why are you defining an unnecessary constant here, just put it in the condition along with a comment explaining the condition.

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
-            (bytes[i + 2] << 8) +
-            (bytes[i + 3]);
+    const padded = (bytes.length + 3) & ~3;
+    for (let i = 0; i < padded; i += 4) {
+        // Mask each byte to 0xFF: the byte array may contain values > 255 due to
+        // M.CHARARRAY encoding non-ASCII strings with full Unicode codepoints.
+        // Uint8Array (used for file output) truncates to 8 bits, so we must do the
+        // same here to produce a checksum that matches the actual file bytes.
+        sum = (sum + (
+            ((bytes[i]     || 0) & 0xFF) << 24 |
```
ILOVEPIE, line 46: Should this use ?? rather than ||?

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
-        bytes.push(0);
-    }
-
     let sum = 0;
-    for (let i = 0; i < bytes.length; i += 4) {
-        sum += (bytes[i] << 24) +
-            (bytes[i + 1] << 16) +
-            (bytes[i + 2] << 8) +
-            (bytes[i + 3]);
+    const padded = (bytes.length + 3) & ~3;
```
ILOVEPIE, line 39: The bitwise math here doesn't seem to serve a purpose... Maybe i'm missing something.

ILOVEPIE: This would make 
(0111 + 0011) & ~0011 = 1100

I'm not understanding this line.

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
+        // Mask each byte to 0xFF: the byte array may contain values > 255 due to
+        // M.CHARARRAY encoding non-ASCII strings with full Unicode codepoints.
+        // Uint8Array (used for file output) truncates to 8 bits, so we must do the
+        // same here to produce a checksum that matches the actual file bytes.
+        sum = (sum + (
+            ((bytes[i]     || 0) & 0xFF) << 24 |
+            ((bytes[i + 1] || 0) & 0xFF) << 16 |
+            ((bytes[i + 2] || 0) & 0xFF) << 8  |
+            ((bytes[i + 3] || 0) & 0xFF)
+        ) >>> 0) >>> 0;
```
ILOVEPIE: Why don't we just forgo this double ``>>> 0`` and (see next comment)

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
     }
-
-    sum %= Math.pow(2, 32);
     return sum;
```
ILOVEPIE: Do this:

```js
return sum >>> 0;
```

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
+        // same here to produce a checksum that matches the actual file bytes.
+        sum = (sum + (
+            ((bytes[i]     || 0) & 0xFF) << 24 |
+            ((bytes[i + 1] || 0) & 0xFF) << 16 |
+            ((bytes[i + 2] || 0) & 0xFF) << 8  |
+            ((bytes[i + 3] || 0) & 0xFF)
+        ) >>> 0) >>> 0;
     }
-
-    sum %= Math.pow(2, 32);
```
ILOVEPIE, line 49: You're right this part is stupid.

src/tables/sfnt.mjs
```diff
@@ -397,7 +398,7 @@ function fontToSfntTable(font) {
     let checkSumAdjusted = false;
     for (let i = 0; i < tableFields.length; i += 1) {
         if (tableFields[i].name === 'head table') {
-            tableFields[i].value.checkSumAdjustment = 0xB1B0AFBA - checkSum;
+            tableFields[i].value.checkSumAdjustment = (0xB1B0AFBA - checkSum) >>> 0;
```
ILOVEPIE, line 402: UINT32 - UINT32 = UINT32 the ```>>> 0``` isn't needed I don't think.

src/tables/sfnt.mjs
```diff
@@ -34,19 +34,20 @@ function log2(v) {
...
-            (bytes[i + 1] << 16) +
-            (bytes[i + 2] << 8) +
-            (bytes[i + 3]);
+    const padded = (bytes.length + 3) & ~3;
+    for (let i = 0; i < padded; i += 4) {
+        // Mask each byte to 0xFF: the byte array may contain values > 255 due to
+        // M.CHARARRAY encoding non-ASCII strings with full Unicode codepoints.
+        // Uint8Array (used for file output) truncates to 8 bits, so we must do the
+        // same here to produce a checksum that matches the actual file bytes.
+        sum = (sum + (
```
ILOVEPIE: Why are we not using ``+=``.

src/tables/sfnt.mjs
ILOVEPIE: also, ``| 0`` is 1 character shorter than ``>>> 0`` and has the same effect.

src/tables/cmap.mjs
ILOVEPIE: especially since it's explained in the comment.

ILOVEPIE (dismissed):
It requires some revision, there's some redundant operations and the code is written in a manner that adds unnecessary lexical complexity (it's unnecessarily harder to read). Please also explain the reasoning behind the padding constant, maybe I'm just misunderstanding what's going on there.

fdb:
@ILOVEPIE thanks for the review 🙏

ILOVEPIE:
> @ILOVEPIE thanks for the review 🙏

I felt I should take this one, as I am familiar with the use case of the PR author.

ILOVEPIE:
@fdb if you have any other comments, feel free to jump in on my review or run your own too.

mtsubs:
Glad to see @ILOVEPIE working on subtitle-related frontend practices as well.

I really appreciate @ILOVEPIE's code review. The code does have some inelegant patterns because I relied on AI assistance to help fix bugs, and thoroughly reviewing the entire repository isn't straightforward, haha.

I've added supplemental comments after carefully evaluating your suggestions. Some recommendations are valid, while others are better left as is for now.

This PR may not be ready for merge yet and needs further evaluation and testing, which is why I haven't adopted the direct fix in my production deployment and instead used a workaround approach.

I'm also glad to hear about the 2.0 release, though I won't adopt it immediately since it adds more granular CJK character table handling that I've already implemented by reading the font binary directly. I may prioritize that adaptation in the future. Still, thank you for the update, and thank you to @fdb and the other contributors for the excellent work.

Looking forward to seeing @fdb and @ILOVEPIE continue improving this PR. Your understanding of the project runs much deeper than mine. Thanks for everything.

mtsubs:
> > @ILOVEPIE thanks for the review 🙏
> 
> I felt I should take this one, as I am familiar with the use case of the PR author.

That would be great. I've committed a new fix based on yours.

It hasn't been synced to this PR yet. But you can see the changes here:

https://github.com/mtsubs/opentype.js/commit/5d23d21e337c762b130cee2acb67efe7edc90316

ILOVEPIE:
Do we need the double unsigned right shift? Can't we do that at the end of the function, also why aren't we using ``+=``?

ILOVEPIE:
Also if you are changing addSegment, make sure no public API stuff is changed, either by making a wrapper or verifying that all addSegment calls are internal and updated to the new signature.

mtsubs:
I've reconsidered and verified this. I apologize for my earlier reply,I was over-relying on AI to check and evaluate the code.

You're right. Once the inner `>>> 0` converts the assembled word to a proper non-negative uint32, `sum` stays positive throughout the accumulation and `+=` is safe. The final `return sum >>> 0` is sufficient to clamp the total back into uint32 range. The double `>>> 0` per iteration was unnecessary.

Regarding `addSegment`: I rely on AI to review the code, so my understanding of the entire codebase may not be as comprehensive as yours, and further verification on your end is welcome. That said, based on the AI review: `addSegment` is not exported. `cmap.mjs` only exports `parse`, `make`, `parseCmapTableFormat0`, `parseCmapTableFormat14`, and `makeCmapTable`. The function has exactly one call site inside the file, which is the loop already updated to the new signature. No wrapper should be needed, but please do verify.

ILOVEPIE:
Correct me if I'm wrong but isn't there a ``<<<`` operator for unsigned left shift meaning we don't need the shifts with zero. Maybe I'm misremembering but I could have sworn there was.

EDIT: nevermind, that's C.

ILOVEPIE:
Please update the PR, i don't know how, but your branch has become decoupled from the PR.

mtsubs:
> Please update the PR, i don't know how, but your branch has become decoupled from the PR.

Sure, the PR has been updated.

src/tables/sfnt.mjs
ILOVEPIE: Nevermind, this is actually clever, but we should add a comment explaining this rounds up to multiples of 4.

ILOVEPIE:
Please apply this patch to your PR, I cannot push to the PR branch for some reason.
[remaining_edits.patch](https://github.com/user-attachments/files/27839233/remaining_edits.patch)

mtsubs:
> Please apply this patch to your PR, I cannot push to the PR branch for some reason. [remaining_edits.patch](https://github.com/user-attachments/files/27839233/remaining_edits.patch)

Hello,
Thanks, the import is complete, and you are invited to join the mtsubs/opentype.js repository, so all your edits will be included in this PR.

ILOVEPIE (commented):
Looks good to me now, but seeing as I contributed code, I'm going to need another person to sign off on the changes.

ILOVEPIE:
It would appear the style checks don't like ``??`` we can use ``||`` for strict ES6 compliance.

It's less accurate to what we want in terms of what it does however.

yne (dismissed):

yne (approved):

yne:
> I'm going to need another person to sign off on the changes.

Looks good to me :)

PR #832, opened 2026-04-29
Harden release workflow

fdb:
## Description

Currently, the workflow to create a release is to use a specific GH action that automates the process and tries to avoid issues. This PR fixes this and adds some extra checks:

- Modernize actions (use newer versions)
- Use `npm ci` instead of `npm install` (makes sure we use the exact packages as specified in `package-lock.json`

When running:

- Check that there is no release running concurrently (very niche, but good to check)
- Check that the version does not exist yet on npm or as a git tag
- Also create a git tag + GitHub release

## Motivation and Context

I've added this to make sure releases are foolproof (the fool being me)

## How Has This Been Tested?

By looking over the code carefully; we can't test this unless by running a release, or doing a dry-run (but that means we first need to merge this PR)

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-node@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
       with:
-        node-version: 18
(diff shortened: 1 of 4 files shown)


ILOVEPIE (commented):
This should not have been merged. The versioning was completely incorrect.

ILOVEPIE:
See #837

ILOVEPIE:

PR #833, opened 2026-04-29
Use trusted npm publishing for releases

fdb:
## Summary

To make a new release of OpenType.js, we want to use npmjs' [trusted publishing](https://docs.npmjs.com/trusted-publishers). That means we don't rely on secrets in our repo anymore, but we make a direct link between the two that does the auth. 

- Switch the npm release workflow to trusted publishing by removing the long-lived npm token and explicit provenance config.
- Upgrade npm before publishing and run releases on Node 24.

## Test plan
- [x] Checked workflow diff.
- [x] Checked workflow file diagnostics.
- [ ] Run the release workflow dry run in GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


ILOVEPIE:

PR #839, opened 2026-05-05
Fix toPathData returns NaN

### LINKED ISSUE #673: An unexpected situation occurred when using path. toSVG(), where NaN caused text rendering errors
Recently, while testing the file again, I accidentally encountered several rendering failures after using toSVG() for text. The path data is shown in the figure:
Snipaste_2024-02-04_15-14-05

Then I briefly analyzed the problem and its corresponding source code, and found out where the problem lies:
Snipaste_2024-02-04_15-18-11
Because JS uses Scientific notation by default for floating point numbers with large numbers of digits, according to the previous code writing method, a NaN will be returned here, resulting in an error in the path. I think that although the probability of this situation is relatively low, it should also be considered.
Snipaste_2024-02-04_15-19-12

My personal suggestion is that if I want to solve this problem without refactoring the original code, I try to use tofix () to prevent js from using Scientific notation, and the accuracy parameters are controlled by the user to ensure correctness. After my simple test, I found that this problem can be effectively solved.
Snipaste_2024-02-04_15-19-28

So, does anyone have better suggestions or methods?

### LINKED ISSUE #838: toPathData returns NaN
Here is the minimal repro:
```js
import opentype from "opentype.js";

const path = new opentype.Path();
const x = -172.78799999998984;
const y = 1.4210854715202004e-14;

path.moveTo(x, y);

console.log("Raw command:", path.commands[0]);
console.log("Number.isFinite(y):", Number.isFinite(y));
console.log("Expected rounded y:", y.toFixed(10));
console.log("Actual toPathData(10):", path.toPathData(10));
console.log(
  "Actual toPathData({ decimalPlaces: 10, flipY: false, optimize: false }):",
  path.toPathData({ decimalPlaces: 10, flipY: false, optimize: false }),
);

```

Output is:
```txt
Raw command: { type: 'M', x: -172.78799999998984, y: 1.4210854715202004e-14 }
Number.isFinite(y): true
Expected rounded y: 0.0000000000
Actual toPathData(10): M-172.7880000000 NaN
Actual toPathData({ decimalPlaces: 10, flipY: false, optimize: false }): M-172.7880000000 NaN
```

It should not produce `NaN` while rounding.

miroljub1995:
Fixes https://github.com/opentypejs/opentype.js/issues/838

## Types of changes

## Checklist:

### DIFF
--- a/src/path.mjs
+++ b/src/path.mjs
@@ -20,6 +20,12 @@ function Path() {
 
 const decimalRoundingCache = {};
 
+function decimalShift(num, exp) {
+    const parts = String(num).split(/e/i);
+    const exponent = parts[1] ? +parts[1] : 0;
+    return +(parts[0] + 'e' + (exponent + exp));
+}
+
 function roundDecimal(float, places) {
     const integerPart = Math.floor(float);
     const decimalPart = float - integerPart;
@@ -33,15 +39,10 @@ function roundDecimal(float, places) {
         return integerPart + roundedDecimalPart;
     }
     
-    const decimalExponentParts = String(decimalPart).split('e');
-    const decimalExponent = decimalExponentParts[1] ? +decimalExponentParts[1] : 0;
-    const scaledDecimalValue = Math.round(+(decimalExponentParts[0] + 'e' + (decimalExponent + places)));
-    const roundedExponentParts = String(scaledDecimalValue).split('e');
-    const roundedExponent = roundedExponentParts[1] ? +roundedExponentParts[1] : 0;
-    const unscaledRoundedValue = +(roundedExponentParts[0] + 'e' + (roundedExponent - places));
-    decimalRoundingCache[places][decimalPart] = unscaledRoundedValue;
+    const roundedDecimalPart = decimalShift(Math.round(decimalShift(decimalPart, places)), -places);
+    decimalRoundingCache[places][decimalPart] = roundedDecimalPart;
 
-    return integerPart + unscaledRoundedValue;
+    return integerPart + roundedDecimalPart;
 }
 
 function optimizeCommands(commands) {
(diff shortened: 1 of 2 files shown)


ILOVEPIE:
I'm extra cautious with AI generated code, have you tested that this does not cause other issues? Also why is the extra function nessisary? I still need to verify the issue wasn't hallucinated too.

miroljub1995:
> I'm extra cautious with AI generated code, have you tested that this does not cause other issues?

I am too, you should have seen the first AI proposal haha. Yes, there are, also, [tests](https://github.com/opentypejs/opentype.js/blob/e2eaedebfa6187c1b435a87e6b7b02e6f1ba1b48/test/path.spec.mjs#L83) that would fail if something is really bad

> I still need to verify the issue wasn't hallucinated too.

I'm not sure why would it be a problem as it is used on 2 places, not just one.

ILOVEPIE:
I see what this does now, I'm unsure if this fixes a real world problem however as I don't think don't think font files have that many places of precision. I'll leave this up to the other maintainers.

Connum (approved):
Looking good, this also fixes at least two other issues: 
fixes #673 
fixes #814

lucascaton:
Hi there. Thanks for fixing this issue 🙂
When are you planning to release **v2.0.1**? Thanks!

ILOVEPIE:
Sorry about not being as active, I've been pretty busy unexpectedly.

lucascaton:
Hi @ILOVEPIE, I completely understand how hard it can be to find time for an open-source project (I have a few myself 😅), and I really appreciate all the work you put into opentype.js! ❤️

Since the fix was already merged, would you happen to have some time to publish a new release? It'd make a big difference for everyone affected by this issue 🙂

And if there's anything I can do to help with the release, I'd be more than happy to!

ILOVEPIE:
Unfortunately I don't think I have control of releases but I'll check.

Issue #801, opened 2025-04-11
Not supporting for some feature tags-"calt" (Contextual Alternates)


trangluongdoi2:
I have a font that enables the "calt" feature. For example, the text "**abcd**" is rendered like that (Just a span text on browser with CSS (font-feature-settings: "calt" on;)
![Image](https://github.com/user-attachments/assets/7bc21c41-4734-4dc3-8e43-673cdf073ce6)

But when I use the function **_forEachGlyphs_** for the text with **GlyphRenderOptions** like this:
`const renderOptions = {
		script: 'latn',
		kerning: true,
		features: {
      calt: true,
      liga: true,
      rlig: true,
		},
	}` 
I received the results and used them to render on the browser by rendering them in an SVG file with paths, It is not working well like I expected.

![Image](https://github.com/user-attachments/assets/79caf85b-764c-4b6b-9fab-88d4b31c9cbc)

Don't worry about the position when I use paths to render because it is for illustration purposes only. 
How can I fix it?

wiltse439:
Note: This may be fixed by PR https://github.com/opentypejs/opentype.js/pull/848

PR #848, opened 2026-05-13
Add support for Contextual Alternatives

wiltse439:
## Description
Add support for Contextual Alternative (calt) substitutions. This makes some latin cursive fonts work correctly, like Google Playwright.

## Motivation and Context
Some fonts use substitutions based on context of a glyph. Context includes beginning or end of a word, middle of a word, and so forth. Not only may a glyph be substituted, it can be expanded to multiple glyphs, sort of an inverse of a ligature. Fonts like the Playwright cursive font use multi-step substitutions where interior letters in a word are replaced with medial forms and then an additional contextual step adds glyphs in between for connecting strokes.

## How Has This Been Tested?
Primary testing has been done using the Google Playwright variable font.

### Before

Font is Google Playwright.

image

### After

image

## Types of changes

## Checklist:

## Bugs fixed

https://github.com/opentypejs/opentype.js/issues/801

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ Options is an optional _{GlyphRenderOptions}_ object containing:
 * `language`: language system used to determine which features to apply (default: `"dflt"`)
 * `kerning`: if true takes kerning information into account (default: `true`)
 * `features`: an object with [OpenType feature tags](https://docs.microsoft.com/en-us/typography/opentype/spec/featuretags) as keys, and a boolean value to enable each feature.
-Currently only ligature features `"liga"` and `"rlig"` are supported (default: `true`).
+Currently only ligature features `"liga"`, `"rlig"` and `"calt"`  are supported (default: `true`).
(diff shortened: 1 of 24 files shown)


ravenwits:
Thanks for this work `calt` support has been a meaningful gap for anyone generating or rendering fonts that rely on contextual alternates. I write the GSUB type 6 table directly via a custom patcher, but I need opentype.js to correctly apply those rules in my in-browser font preview. This PR would close that loop for me. Would it be possible to prioritize a review of #816 first so this can follow? Both PRs together would unblock a real and underserved set of font use cases.

wiltse439:
I didn't focus on font generation, but I am happy to work on the write paths if I had a way to test it. As for reviews, I am not really clear if there is any maintainer that can review these PRs or not. I do see some activity on the repo, but my PRs have been waiting for several months now. Maybe I need to do something different to get attention to them?

PR #849, opened 2026-05-14
Fix (gvar): fix wrong interpolation when tuple has private points flag with empty list.

fabiocaccamo:
Fix (gvar): fix wrong interpolation when tuple has private points flag with empty list.

## Description
When a gvar tuple has the `PRIVATE_POINT_NUMBERS` flag set but an empty point list (packed count = 0), it means "all glyph points". The code was treating this the same as "no flag", falling back to `sharedPoints` and reading the wrong number of deltas. Added `hasPrivatePoints` to track the flag separately from the list length.

## Motivation and Context
Variable fonts where some tuples use this encoding (e.g. **Inter Variable** `wght` axis) were producing wrong glyph outlines because the Y deltas were read from the wrong offset in the binary data.

## How Has This Been Tested?
Verified visually with "Inter Variable" (`wght` + `opsz` axes) - glyph outlines (specifically "DJ") now match the expected coordinates from `fontTools`. Ran the full test suite (all passing).

### Before
Screenshot 2026-05-14 at 12 58 07

### After
Screenshot 2026-05-14 at 12 58 55

## Types of changes

## Checklist:


ILOVEPIE:
@Connum requesting a review.

wiltse439:
Funny, I just created a draft PR to fix this same problem I think. I'll compare my version of the change with this to see if we did it the same way.

src/parse.mjs
```diff
@@ -897,6 +897,10 @@ Parser.prototype.parseTupleVariationStore = function(tableOffset, axisCount, fla
         if(header.flags.privatePointNumbers) {
             header.privatePoints = this.parsePackedPointNumbers();
         }
+        // remember if this tuple had the private points flag set.
+        // an empty list (length 0) means "all glyph points", which is
+        // different from no flag at all (= use shared points).
+        header.hasPrivatePoints = !!header.flags.privatePointNumbers;
```
wiltse439: Alternatively, you can just remove the default value set for header.privatePoints at line 890 above, and then in the test on line 918 it can be `pointsCount = header.privatePoints ? header.privatePoints.length : sharedPoints.length;`, and the same change in variationprocessor.mjs. That avoids the need to have an extra boolean in the header.

fabiocaccamo: @wiltse439 good point, thanks! I pushed the suggested changes.

fabiocaccamo:
@wiltse439 any update?

fabiocaccamo:
@Connum @wiltse439 this PR also fix the same all-points encoding for composite glyphs: the delta count was computed from the resolved outline points instead of one delta per component, reading past the tuple data, and `transformComponents` never applied the offset deltas because `indexOf` on an empty point list always returns `-1`. 

Verified against `fontTools` with a subset of "Geist Italic" (OFL), where the dot of the composite "j" must shift +39 units at wght 900.

wiltse439:
I am not a maintainer so I can't approve this PR. I created a discussion item to ask about forking this project to a new repo where we can move at a faster pace on new development.

Issue #847, opened 2026-05-12
Closure Compiler types need an update and inclusion in package.


ILOVEPIE:
We need to update our closure compiler extern types for recent changes and include them in the published package.

PR #853, opened 2026-05-31
Automatically generate Closure compiler types.

ILOVEPIE:
Added a utility to automatically generate types for closure compiler.

## Description
This adds `scripts/externs-generator.js` to generate externs for the closure compiler.

## Motivation and Context
Our closure compiler externs are outdated and it is difficult to keep them updated, so let's automate it.

## How Has This Been Tested?
I verified the output matches the source code.

**This PR was mostly AI generated, I plan to rework and revise the code manually before merging to improve it.** I didn't use AI to change code in src. 

## Checklist:

### DIFF
--- a/src/parse.mjs
+++ b/src/parse.mjs
@@ -117,44 +117,156 @@ const masks = {
     MAP_ENTRY_SIZE_MASK: 0x30,
 };
 
-// A stateful parser that changes the offset whenever a value is retrieved.
-// The data is a DataView.
+/**
+ * @typedef {Object} LangSysTable
+ * @property {number} reserved
+ * @property {number} reqFeatureIndex
+ * @property {number[]} featureIndexes
+ */
+
+/**
+ * @typedef {Object} LangSysRecord
+ * @property {string} tag
+ * @property {LangSysTable} langSys
+ */
+
+/**
+ * @typedef {Object} ScriptRecord
+ * @property {string} tag
+ * @property {{defaultLangSys?: LangSysTable, langSysRecords: LangSysRecord[]}} script
+ */
+
+/**
+ * @typedef {Object} FeatureRecord
+ * @property {string} tag
+ * @property {{featureParams?: number, lookupListIndexes: number[]}} feature
+ */
+
+/**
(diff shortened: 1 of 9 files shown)


ILOVEPIE:
#847 is partially solved by this, we still need to add the externs into the NPM package.

src/parse.mjs
```diff
@@ -117,77 +117,214 @@ const masks = {
...
+/**
+ * @typedef {Object} ScriptRecord
+ * @property {string} tag
+ * @property {{defaultLangSys?: LangSysTable, langSysRecords: LangSysRecord[]}} script
+ */
+
+/**
+ * @typedef {Object} FeatureRecord
+ * @property {string} tag
+ * @property {{featureParams?: number, lookupListIndexes: number[]}} feature
```
yne, line 142: why not declare the whole type as single statement ?
```js
/**
@typedef {{
  tag: string
  feature: { 
    featureParams?: number,
    lookupListIndexes: number[]
  }
}
} FeatureRecord
*/
```
since it's what's being done for the @property "feature"

yne (dismissed):

src/parse.mjs
ILOVEPIE: Sorry I was really tired when I added the types. I'll fix them. Property does allow describing properties, which the other way doesn't however.

ILOVEPIE:
I made extern generation more elegant. It no longer includes private members, and properly handles typedefs.

yne:
if this file can be generated, can we generate it from a npm build target ?

I'm not confortable commiting build artefact
also, some types are too wide (object instead of the class)

but that's not the goal of this PR so let's do that later

yne (approved):

ILOVEPIE:
> if this file can be generated, can we generate it from a npm build target ?
> 
> I'm not confortable commiting build artefact
> also, some types are too wide (object instead of the class)

The generator script has to be present to generate the types. If you are referring to the output, that's something to remove when the extens get included in the package.

yne:
yes i was referring to the output.

thanks for the clarification

PR #861, opened 2026-07-24
Support GSUB type 6 format 2 with nested single substitutions

mh0pe:
## Description

Adds execution support for the GSUB chaining contextual substitution format that
the parser already exposes as lookup type 6, format 2.

The implementation:

- gates rules through coverage and the input class set;
- matches backtrack classes in reverse logical order, then input and lookahead
  classes through their separate class definitions;
- applies nested lookup records in design order, with later records consuming
  earlier substitutions;
- retains each `sequenceIndex`, so only the targeted token is changed;
- stops at the first matching rule and subtable, including deliberate no-op
  rules;
- supports nested single substitutions (formats 1 and 2), including type 7
  extension wrappers.

The nested-lookup scope is intentionally limited to single substitutions. Other
nested lookup types fail explicitly rather than being silently skipped. Lookup
flag filtering remains a broader shaping-engine follow-up; both real-world
reproducers below use lookup flag 0 and nested type 1.

## Motivation and Context

On current `master`, the presence of a type 6 / format 2 subtable causes
`getPath()` and `getAdvanceWidth()` to throw before coverage is even evaluated:

```text
substitutionType : 62 lookupType: 6 - substFormat: 2 is not yet supported
```

This reproduces with both SF Hebrew Rounded and the current Google Fonts Roboto
variable font.

This is a focused extraction of the GSUB 6/2 portion of #816, following the
review suggestion there to split the GSUB work from unrelated variation
changes. Thanks to @wiltse439 for the original implementation work; the commit
also credits them as a co-author. This PR leaves the existing type 6 / format 3
path untouched, preserving the `sequenceIndex` fix from #824.

## How Has This Been Tested?

- `npm test`: 348 passing, including build, minified distributions, and source
  lint.
- Six synthetic regressions cover reverse backtrack order, first-match no-op
  rules, ordered records, nonzero `sequenceIndex` through token application,
  first matching nested and top-level subtables, extension wrappers, and
  explicit unsupported-type errors.
- SF Hebrew Rounded:
  - an end-of-run mark changes from glyph 110 to final-form glyph 163;
  - the same mark followed by a protected lookahead remains glyph 110;
  - the previous unrelated Latin probe renders without throwing.
- Google Fonts Roboto variable font:
  - `i` + combining acute shapes to glyphs `[636, 169]`;
  - `affinity` renders without throwing.
- Both generated ESM and CommonJS distributions produce the same real-font
  results.

Not applicable.

## Types of changes

## Checklist:

### DIFF
--- a/src/features/applySubstitution.mjs
+++ b/src/features/applySubstitution.mjs
@@ -21,6 +21,22 @@ function singleSubstitutionFormat2(action, tokens, index) {
 }
 
 /**
+ * Apply chaining context substitution format 2.
+ * @param {Array} substitutions substitutions
+ * @param {any} tokens a list of tokens
+ * @param {number} index token index
+ */
+function chainingSubstitutionFormat2(action, tokens, index) {
+    for (let i = 0; i < action.substitution.length; i++) {
+        const substitution = action.substitution[i];
+        tokens[index + substitution.sequenceIndex].setState(
+            action.tag,
+            substitution.substitution
+        );
+    }
+}
+
+/**
  * Apply chaining context substitution format 3
  * @param {Array} substitutions substitutions
  * @param {any} tokens a list of tokens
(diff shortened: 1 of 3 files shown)


mh0pe:
Friendly maintainer ping @fdb @Connum — this is the focused GSUB type 6 format 2 extraction suggested during review of #816. It keeps the scope to nested single substitutions, fails explicitly for unsupported nested lookup types, and includes synthetic plus real-font coverage. I’d appreciate a review when time permits and am happy to adjust the scope.

PR #866, opened 2026-07-31
feat: parse font collections (.ttc/.otc), selecting a member by PostScript name

### LINKED ISSUE #379: Error: Unsupported OpenType signature ttcf
I am trying to open font file with multiple styles inside.

## Expected Behavior

file is handled, it shows multiple styles.

## Current Behavior
image

## Possible Solution
I have know idea to be honest :)

compile multiple ttf fonts into one file using this guide https://graphicdesign.stackexchange.com/a/104892/137343

## Context
I want to get supported styles of a font from font file

## Your Environment
* Version used: web preview
* Font used: roboto from google fonts
* Browser Name and version: latest stable firefox and chrome
* Operating System and version (desktop or mobile): mac, ubuntu 19.04
* Link to your project:

orochi235:
## Description

Adds support for parsing TrueType/OpenType **collections** (`.ttc` / `.otc`), with `opt.postscriptName` selecting which member to parse.

```js
const font = opentype.parse(buffer, { postscriptName: 'Helvetica-Bold' });
```

Omitting `postscriptName` parses the first member. Asking for a name the collection does not hold throws an error listing the names it does hold, which doubles as a way to discover a collection's contents:

```
Font collection has no font named "Nope". Available: Helvetica, Helvetica-Bold,
Helvetica-Oblique, Helvetica-BoldOblique, Helvetica-Light, Helvetica-LightOblique
```

The change is smaller than it might sound. A collection is *N* table directories over one shared pool of tables, and the only thing standing in the way was that `parseOpenTypeTableEntries` hardcoded the directory at byte 12. Everything downstream is already offset-agnostic — table records carry absolute file offsets, and tables are read from a view over the whole buffer — so this adds a `directoryOffset` parameter, a `ttcf` branch that locates the requested member's directory, and changes nothing else.

The selected member's own sfnt version sets `outlinesFormat`, so a collection mixing `glyf` and CFF members works.

## Motivation and Context

Closes #379.

Font collections are the format a large share of macOS system families ship as — Helvetica, Times, Courier, Menlo, Avenir, Apple Color Emoji. They are also what `queryLocalFonts()` hands back for those families, so any browser code reading installed fonts hits `Unsupported OpenType signature ttcf` immediately.

The API follows @Jolg42's suggestion in #379 (a `postscriptName` on the open method, iterating the collection to match), which also matches `fontkit`'s `create(buffer, postscriptName)`.

## How Has This Been Tested?

`npm test` green — 346 passing, including build, dist and lint.

**New tests** (`test/opentype.spec.mjs`, `describe('font collections')`): first-member default, selection by PostScript name, reading shared tables from a non-first member, and the not-found error.

**New fixture**, generated by `test/generate-collection-font.mjs`, following the existing `generate-*.mjs` pattern. Its two members deliberately **share every table except `name`** — that detail is the point of the test. A parser that assumed a member's tables follow its own directory would read member 0's tables and still look correct for member 1; only honouring the absolute offsets in each record gives the right answer. The tests assert both the `name` table and the shared outline.

**Against real files:** every font collection on a macOS install (`/System/Library/Fonts` and `Supplemental`, 128 files) — **110 parse end-to-end**, glyph outlines included. The 18 that fail do so *after* the collection layer has resolved the directory, on pre-existing limitations a standalone `.ttf` would hit identically (13 on `cmap` format 6, 5 on an unrelated undefined read). No file failed to have its member directories located.

Tested on Node 24, macOS.

## Types of changes

## Checklist:


orochi235:
Hi, actual human here! I ran into this and then found #379, where you basically wrote an LLM prompt implementing it seven years ago. I just pushed the ball down the hill a little :)

shbernal:
Corroboration, if it helps this along. This hit me in a .pptx generation library that measures text to decide autofit: on Windows the faces you need for East Asian text (MS Gothic, Yu Gothic, SimSun, YaHei, JhengHei, MingLiU) all ship as `.ttc`, and so does Cambria, so `parseBuffer` threw on exactly the fonts that mattered. I ended up writing the same unwrap without having seen this PR, and reached the same conclusion about the format: table records carry file-absolute offsets, so a member never has to be copied out, and locating its directory is the whole job.

Worth adding that I checked it against real files rather than only a generated one — a builder and a reader can agree on a layout real fonts don't use. Comparing advance widths to what WPF's `GlyphTypeface` reports, 38 faces across 15 Windows collections agree. Deliberately parsing member 0 regardless of the selector fails 8 of those, which is the failure mode I'd watch: the wrong member returns plausible numbers rather than an error (MS Gothic vs MS PGothic differ ~26% on Latin advances).

Implementation and the fixture oracle: https://github.com/shbernal/ts-pptx/blob/408b7c27/src/measure/font-collection.ts — happy to hand over the oracle JSON if a real-file test would be useful here.

PR #871, opened 2026-08-21
Add support for parsing PostScript Type 1 fonts (PFA/PFB)

KeyPDF:


KeyPDF:
Screenshot_22-8-2026_115238_localhost

demo with pfb font.

